PR #9, opened 2020-04-21
Vcpkg dependencies

nosferalatu:
This PR adds a new script `vcpkg\install_vcpkg_dependencies.bat` so that Tracy can build out of the box on Windows after you run the script.

The previous Tracy code required that you run `vcpkg integrate install` but that modifies the global Windows environment. Another way to use vcpkg is to have it build everything in a self-contained directory, rather than modify global state. This makes it possible to have different versions of vcpkg with different dependencies on the same machine. 

We're using both vcpkg and Tracy within Adobe in various ways on different projects, and we'd like to avoid `vcpkg integrate install`, so that projects are self-contained.

The .bat file does three things:
* clone vcpkg
* build vcpkg (using its bootstrap script)
* run vcpkg install ...the Tracy dependencies...

`vcpkg\install_vcpkg_dependencies.bat` puts everything in the `vcpkg\vcpkg` directory, which is ignored via .gitignore. You can safely erase `vcpkg\vcpkg` and re-run the script if you want to build the dependencies from scratch; it's all self-contained. However, you should only need to run the script once after you clone the repo, or when you need to add a new dependency.

This PR also adjusts the .vcxproj files so the include and lib paths point to the local vcpkg installation, and updates the manual with this information.

One small suggestion I have that I didn't implement in this PR: we can set up a Visual Studio .props file that the .vcxproj files refers to, and that single .props file can add the vcpkg include and lib paths, so that information is set in a single place. I didn't do that in this PR but I can make a second PR with that change if you'd like me to.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,7 @@ manual/t*.synctex.gz
 manual/t*.toc
 profiler/build/win32/packages
 profiler/build/win32/Tracy.aps
+
+# include the vcpkg install script but not the files it produces
+vcpkg/*
+!vcpkg/install_vcpkg_dependencies.bat
--- a/capture/build/win32/capture.vcxproj
+++ b/capture/build/win32/capture.vcxproj
@@ -89,10 +89,12 @@
       <PreprocessorDefinitions>TRACY_NO_STATISTICS;_CRT_SECURE_NO_DEPRECATE;_CRT_NONSTDC_NO_DEPRECATE;WIN32_LEAN_AND_MEAN;NOMINMAX;_USE_MATH_DEFINES;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <EnableEnhancedInstructionSet>AdvancedVectorExtensions2</EnableEnhancedInstructionSet>
       <LanguageStandard>stdcpplatest</LanguageStandard>
+      <AdditionalIncludeDirectories>..\..\..\vcpkg\vcpkg\installed\x64-windows-static\include</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
-      <AdditionalDependencies>ws2_32.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>ws2_32.lib;capstone.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <SubSystem>Console</SubSystem>
+      <AdditionalLibraryDirectories>..\..\..\vcpkg\vcpkg\installed\x64-windows-static\lib</AdditionalLibraryDirectories>
     </Link>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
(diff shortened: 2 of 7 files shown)


wolfpld:
Is it necessary to list all the libraries to link to?

nosferalatu:
> Is it necessary to list all the libraries to link to?

AFAIK there is no way to pass a wildcard to Visual Studio's linker. It would be nice to be able to say "link all .lib files in directory XXX" but I don't think that's possible. So, I added the libraries to link to explicitly in the linker settings. 

I don't know exactly how `vcpkg integrate install` does its magic to avoid needing to explicitly list the libraries to link to. But vcpkg itself isn't magic: after you install a dependency, it's just a .lib in the `vcpkg\installed\x64-windows-static\lib` directory.

wolfpld:
> AFAIK there is no way to pass a wildcard to Visual Studio's linker. It would be nice to be able to say "link all .lib files in directory XXX" but I don't think that's possible.

It seems that's the magic vcpkg is actually doing. From `scripts/buildsystems/msbuild/vcpkg.targets`:

```
<Link>
  <AdditionalDependencies Condition="'$(VcpkgNormalizedConfiguration)' == 'Debug' and '$(VcpkgAutoLink)' != 'false'">%(AdditionalDependencies);$(VcpkgRoot)debug\lib\*.lib</AdditionalDependencies>
  <AdditionalDependencies Condition="'$(VcpkgNormalizedConfiguration)' == 'Release' and '$(VcpkgAutoLink)' != 'false'">%(AdditionalDependencies);$(VcpkgRoot)lib\*.lib</AdditionalDependencies>
  <AdditionalLibraryDirectories Condition="'$(VcpkgNormalizedConfiguration)' == 'Release'">%(AdditionalLibraryDirectories);$(VcpkgRoot)lib;$(VcpkgRoot)lib\manual-link</AdditionalLibraryDirectories>
  <AdditionalLibraryDirectories Condition="'$(VcpkgNormalizedConfiguration)' == 'Debug'">%(AdditionalLibraryDirectories);$(VcpkgRoot)debug\lib;$(VcpkgRoot)debug\lib\manual-link</AdditionalLibraryDirectories>
</Link>
```

But there's also this: `Deactivate Autolinking if lld is used as a linker. (Until a better way to solve the problem is found!)`, which reminded me that I actually want clang builds to be possible, so explicitly listing the libraries might be a good thing.

There are two things to do before this can be merged:
1. I want the current vcpkg setup to also be supported. Your changes do not interfere with the global vcpkg config, but the manual should provide documentation for both ways of configuring vcpkg.
2. The debug build incorrectly uses release libraries. Vcpkg keeps debug libraries in a different directory (see the snippet above), some of them also have different names (e.g. `bz2d.lib`, not `bz2.lib`).

nosferalatu:
I updated the PR with the two things that you requested.

> the manual should provide documentation for both ways of configuring vcpkg

Good point. Okay, I updated the manual with both ways of using vcpkg.

> The debug build incorrectly uses release libraries

Right you are. This has been fixed.

PR #13, opened 2020-05-04
ImGui Multiviewport fixes

ikrima:
- set ImGuiConfigFlags_ViewportsEnable
- correct render loop logic with viewport api calls, `SetNextWindowViewport(), UpdatePlatformWindows(), RenderPlatformWindowsDefault()`
- Fix: coords in abs space now, `SetNextWindowPos()`

NOTE:
- I have viewports turned on by default so you can easy test (comment out `io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable`; and you get old behavior)
- Jankiness with multiviewports isn't bc perf hit; it's bc profiler reduces it's tick rate when it's not in focus. So, that bit of logic needs to be updated if you really care
- I haven't encountered any issues over past week but discount that by 50% since i'm new to tracy. No promises some UI wasn't regresssed
- Key things to watch out for is enabling viewports turns ImGui into using absolute monitor coords instead of window coords (ie SetPosition(0,0) => monitor top left, not window top/left

### DIFF
--- a/profiler/src/main.cpp
+++ b/profiler/src/main.cpp
@@ -281,6 +281,9 @@ int main( int argc, char** argv )
     std::string iniFileName = tracy::GetSavePath( "imgui.ini" );
     io.IniFilename = iniFileName.c_str();
     io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard | ImGuiConfigFlags_DockingEnable;
+    io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable;  // Enable Multi-Viewport / Platform Windows
+    //io.ConfigWindowsMoveFromTitleBarOnly = true;
+    //io.ConfigDockingWithShift = false;
 
     ImGui_ImplGlfw_InitForOpenGL( window, true );
     ImGui_ImplOpenGL3_Init( "#version 150" );
@@ -908,12 +911,21 @@ static void DrawContents()
 
     // Rendering
     ImGui::Render();
-    glfwMakeContextCurrent(s_glfwWindow);
     glViewport(0, 0, display_w, display_h);
     glClearColor(clear_color.x, clear_color.y, clear_color.z, clear_color.w);
     glClear(GL_COLOR_BUFFER_BIT);
     ImGui_ImplOpenGL3_RenderDrawData(ImGui::GetDrawData());
 
-    glfwMakeContextCurrent(s_glfwWindow);
+    // Update and Render additional Platform Windows
+    // (Platform functions may change the current OpenGL context, so we save/restore it to make it easier to paste this code elsewhere.
+    //  For this specific demo app we could also call glfwMakeContextCurrent(window) directly)
+    if (ImGui::GetIO().ConfigFlags & ImGuiConfigFlags_ViewportsEnable)
+    {
+        GLFWwindow* backup_current_context = glfwGetCurrentContext();
+        ImGui::UpdatePlatformWindows();
+        ImGui::RenderPlatformWindowsDefault();
+        glfwMakeContextCurrent(backup_current_context);
+    }
+
     glfwSwapBuffers(s_glfwWindow);
 }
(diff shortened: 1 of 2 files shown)


ikrima:
Also, maybe it'll be useful for future reference/debugging, here are my viewport upgrade notes when I integrated it into our anim tool (most from imgui sourcecode/commit comments)

# Steps to use multi-viewports in your application, when using a custom back-end:

- Important: THIS IS NOT EASY TO DO and comes with many subtleties not described here!
              It's also an experimental feature, so some of the requirements may evolve.
              Consider using default back-ends if you can. Either way, carefully follow and refer to examples/ back-ends for details.

- [x] Application:  Enable feature with 'io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable'.
- [x] Back-end:     Hook ImGuiPlatformIO's Platform_* and Renderer_* callbacks (see below).
  - [x] Set 'io.BackendFlags |= ImGuiBackendFlags_PlatformHasViewports' and 'io.BackendFlags |= ImGuiBackendFlags_PlatformHasViewports'.
  - [x] Update ImGuiPlatformIO's Monitors list every frame.
  - [x] Update MousePos every frame, in absolute coordinates.
- [x] Application:  In your main loop, call ImGui::UpdatePlatformWindows(), ImGui::RenderPlatformWindowsDefault() after EndFrame() or Render().
  - [x] You may skip calling RenderPlatformWindowsDefault() if its API is not convenient for your needs. Read comments below.
- [x] Application:  Fix absolute coordinates used in ImGui::SetWindowPos() or ImGui::SetNextWindowPos() calls.

About ImGui::RenderPlatformWindowsDefault():

- [x] This function is a mostly a _helper_ for the common-most cases, and to facilitate using default back-ends
- [x] You can check its simple source code to understand what it does.
  - It basically iterates secondary viewports and call 4 functions that are setup in ImGuiPlatformIO, if available:
  - Platform_RenderWindow(), Renderer_RenderWindow(), Platform_SwapBuffers(), Renderer_SwapBuffers()
  - Those functions pointers exists only for the benefit of RenderPlatformWindowsDefault().
    - If you have very specific rendering needs (e.g. flipping multiple swap-chain simultaneously, unusual sync/threading issues, etc.),
      you may be tempted to ignore RenderPlatformWindowsDefault() and write customized code to perform your renderingg.
    - You may decide to setup the platform_io's *RenderWindow and *SwapBuffers pointers and call your functions through those pointers,
      or you may decide to never setup those pointers and call your code directly. They are a convenience, not an obligatory interface.

- IMPORTANT: When multi-viewports are enabled (with io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable),
  all coordinates/positions will be in your natural OS coordinates space. It means that:
  - Reference to hard-coded positions such as in SetNextWindowPos(ImVec2(0,0)) are _probably_ not what you want anymore.
    Use GetMainViewport()->Pos to offset hard-coded positions, e.g. SetNextWindowPos(GetMainViewport()->Pos).
  - Likewise io.MousePos and GetMousePos() will use OS coordinates.
    If you query mouse positions to interact with non-imgui coordinates you will need to offset them.
    e.g. subtract GetWindowViewport()->Pos.
- Render function: the ImDrawData structure now contains 'DisplayPos' and 'DisplaySize' fields.
  To support multi-viewport, you need to use those values when creating your orthographic projection matrix.
  Use 'draw_data->DisplaySize' instead of 'io.DisplaySize', and 'draw_data->DisplayPos' instead of (0,0) as the upper-left point.
  You need to subtract 'draw_data->DisplayPos' from your scissor rectangles to convert them from global coordinates to frame-buffer coordinates.
- IO: Moved IME support functions from io.ImeSetInputScreenPosFn, io.ImeWindowHandle to the PlatformIO api.
- IO: Removed io.DisplayVisibleMin, io.DisplayVisibleMax settings (they were marked obsoleted, used to clip within the (0,0)..(DisplaySize) range).

wolfpld:
Viewports will be disabled by default, as they crash on Linux.

PR #19, opened 2020-05-11
Make TracySourceView.cpp build with somewhat old clang

ethercrow:
This was the only issue preventing build on macOS High Sierra with whatever version of clang it has.


wolfpld:
This was actually invalid code. Interesting that it worked. Small string optimization maybe?

Issue #20, opened 2020-05-11
import-chrome: "args" is not imported


ethercrow:
I'm importing trace data in Chrome format where most of the useful info is contained under `args` key like this:

``` 
 293   │   {
 294   │     "ts": 1589232324367717,
 295   │     "args": {
 296   │       "args": "[\"--user\",\"--no-user-package-db\",\"--package-db\",\"/Users/ethercrow/.stack/snapshots/x86_64-osx/87e154bcf63658f29f1e89e07af68ff63
       │ 963d64fc0e107f1e947001fc28f6600/8.6.5/pkgdb\",\"dump\",\"--expand-pkgroot\"]",
 297   │       "process": "/Users/ethercrow/.stack/programs/x86_64-osx/ghc-8.6.5/bin/ghc-pkg-8.6.5"
 298   │     },
 299   │     "tid": 4,
 300   │     "name": "sinkProcessStdout",
 301   │     "pid": 1,
 302   │     "ph": "B"
 303   │   },
 304   │   {
 305   │     "ts": 1589232324470187,
 306   │     "tid": 4,
 307   │     "name": "sinkProcessStdout",
 308   │     "pid": 1,
 309   │     "ph": "E"
 310   │   },
```

Unfortunately Tracy doesn't show this info:

<img width="1435" alt="Screen Shot 2020-05-11 at 11 27 02 PM" src="https://user-images.githubusercontent.com/222467/81615493-f96fcd80-93e1-11ea-8483-01f6dfb5ac20.png">

and as far as I can see in import-chrome.cpp, doesn't read that field. `tracy::Worker::ImportEventTimeline` doesn't appear to have a field where one could put this data.

wolfpld:
Chrome tracing format is not well documented. While the format description lists every parameter possible, more often than not there's lack of explanation of the intended usage.

I have added support for passing per-zone custom text (as `std::string`) in https://github.com/wolfpld/tracy/commit/8e11cd5ebb38b488ea19bde071c6c7bf4bc6c792 (not tested), you can use that to extend parsing input data in the importer.

ethercrow:
@wolfpld I'll try, thanks! Are there limitations on length and allowed characters like newlines there?

wolfpld:
Fundamentally there shouldn't be. Newlines are explicitly supported (see `Worker::ProcessZoneText()`) and the text _might_ be cut-off by ImGui, or not fit on the screen, if too long, but there are no storage limits.

PR #21, opened 2020-05-12
import-chrome: import zone metadata from "args" key

ethercrow:
Fixes https://github.com/wolfpld/tracy/issues/20

Screen Shot 2020-05-12 at 3 47 15 PM

### DIFF
--- a/import-chrome/src/import-chrome.cpp
+++ b/import-chrome/src/import-chrome.cpp
@@ -70,13 +70,23 @@ int main( int argc, char** argv )
     for( auto& v : j )
     {
         const auto type = v["ph"].get<std::string>();
+
+        std::string zoneText = "";
+        if ( v.contains( "args" ) )
+        {
+            for ( auto& kv : v["args"].items() )
+            {
+                zoneText += kv.key() + ": " + kv.value().dump() + "\n";
+            }
+        }
+
         if( type == "B" )
         {
             timeline.emplace_back( tracy::Worker::ImportEventTimeline {
                 v["tid"].get<uint64_t>(),
                 uint64_t( v["ts"].get<double>() * 1000. ),
                 v["name"].get<std::string>(),
-                "",
+                std::move(zoneText),
                 false
             } );
         }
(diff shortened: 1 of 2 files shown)


wolfpld:
`std::move` should be used for `zoneText`.

import-chrome/src/import-chrome.cpp
```diff
@@ -70,13 +70,23 @@ int main( int argc, char** argv )
     for( auto& v : j )
     {
         const auto type = v["ph"].get<std::string>();
+
+        std::string zoneText = "";
+        if ( v.contains( "args" ) )
+        {
+            for ( auto& kv : v["args"].items() )
+            {
+              zoneText += kv.key() + ": " + kv.value().dump() + "\n";
```
wolfpld, line 79: Please fix indentation.

ethercrow:
I'm rather casual C++ user, updated.

ethercrow:
Is it fine to use repeated `+=` for `std::string` concatenation?

import-chrome/src/import-chrome.cpp
```diff
@@ -96,7 +106,7 @@ int main( int argc, char** argv )
             const auto ts0 = uint64_t( v["ts"].get<double>() * 1000. );
             const auto ts1 = v["dur"].is_object() ? ts0 + uint64_t( v["dur"].get<double>() * 1000. ) : ts0;
             const auto name = v["name"].get<std::string>();
-            timeline.emplace_back( tracy::Worker::ImportEventTimeline { tid, ts0, name, "", false } );
+            timeline.emplace_back( tracy::Worker::ImportEventTimeline { tid, ts0, name, zoneText, false } );
```
wolfpld, line 109: Also here :)

ethercrow: 👍

wolfpld:
In general case it is not recommended, as it involves memory allocation, but in this case it won't matter that much. Chrome traces are tiny.

Issue #26, opened 2020-05-21
Problem with Tracy in a DLL and TRACY_NO_EXIT


mcleary:
Hi, I am doing more tests with Tracy and this time I am trying to use a DLL as described in the Section 2.1.4 of the manual. For this setup I have a single thread application that calls into a library (DLL) that is supposed to be profiled with Tracy but I am getting an access violation in the Profiler destructor. Here are more details of my setup

1. DLL Called `TracyClient.dll` containing just `TracyClient.cpp`
2. A DLL that I want to profile with Tracy, called `TracyBackend.dll`
3. An application that links against `TracyBackend.dll`
4. The application runs with `TRACY_NO_EXIT`

This only happens when `TRACY_NO_EXIT` is defined. Is there a limitation here?

The following image shows the place of the crash.

![image](https://user-images.githubusercontent.com/1537712/82558678-19cc2480-9b66-11ea-8459-e4c9cfabffd5.png)

wolfpld:
> This only happens when TRACY_NO_EXIT is defined. Is there a limitation here?

This is certainly possible. Linking with DLLs is unfortunately a bit of no man's land.

mcleary:
I know that Windows has some weird shenanigans when shutting down DLLs. My guess is that internal threads of tracy are being destroyed without notice and memory is getting corrupted by the time the 'Profiler' destructor gets executed. If there is no easy fix for that I'll keep investigating with this in mind.

mcleary:
Well, my hypothesis was somewhat correct. The `TRACY_NO_EXIT` mechanism operates in a thread. By the time the main function finishes, there are no threads anymore, so that's why it doesn't work. At least I have a patch to prevent the Profiler from crashing. I will submit is soon.

mcleary:
I will close this issue for now as DLLs are a bit complicated but I am glad the application is not crashing anymore due to my patch.

livinamuk:
How did the patch work? I'm having trouble with running Tracy inside my Maya plugin which is a dll.

wbkangtc:
Sorry to necro this, but I noticed that on MSVC with multi_DLL, this is still the case?
I've built `Tracy::TracyClient` as shared lib(dll), and linked it to my executable with:
```
#define TRACY_ENABLE 1
#define TRACY_NO_EXIT 1
#define TRACY_IMPORTS 1
```
And it doesn't wait for connection and just exits.
Building `Tracy::TracyClient` as static and linking it works fine.

[...]

PR #27, opened 2020-05-21
Fix crash when running Tracy from DLLs

mcleary:
Instantiating Tracy from within a DLL will tie its internal threads life-time to the DLL. Windows does not guarantee that threads will be alive after the main function. This has implications in the Profiler dtor since will try to perform some deallocations, however, _memory_deallocate_large will try to get the heap of the current thread which can be invalid at the point of shutdown causing a crash. Checking the pointer here will won't make TRACE_NO_EXIT work, but it will prevent the Profiler from crashing.


wolfpld:
Tagging @mjansson to see if he wants to upstream this change. Related issue: #26.

PR #31, opened 2020-05-27
Add OpenCL trace support

mcleary:
I have been working on adding OpenCL support to Tracy. It took me a while to figure out the proper way to sync the timers so events appear in the right place in the UI.

OpenCL profiling works by querying information from `cl_event` objects, so it is a little bit different than OpenGL and Vulkan, where you can create query ids. I tried to keep the API as close as possible to Vulkan and OpenGL.

There a few things that still need to be done, for instance, the UI will show OpenCL events under *Vulkan Context*, as can be seen in the attached screenshot, but before proceeding with more changes I would like to know if this is something you would be interested in adding to Tracy. If so I can add this to the manual and do what else needs to be done. A test application would be nice to write as well. 

Any feedback in the code would be really appreciated.

I guess I was also looking for some explanation on what as the best practices on using the Serial and Lock Free Queues. I noticed that OpenGL uses lock free and Vulkan uses serial. Is there a particular reason for that?

image


wolfpld:
> I would like to know if this is something you would be interested in adding to Tracy

Yes, this is very interesting.

> the UI will show OpenCL events under Vulkan Context

See 54a02935.

> If so I can add this to the manual and do what else needs to be done. A test application would be nice to write as well.

Manual, test application, and an entry in the `AUTHORS` file, so I remember who to harass about OpenCL in the future ;)

> I guess I was also looking for some explanation on what as the best practices on using the Serial and Lock Free Queues. I noticed that OpenGL uses lock free and Vulkan uses serial. Is there a particular reason for that?

OpenGL is (generally speaking) single threaded, so everything within given context will only happen on one thread. Lock free queue can be used, as it preserves order of events in a thread.

Vulkan is multithreaded and events for a single context may appear from multiple threads. For proper reconstruction of the timeline on the server a valid order is needed, and lock free queue doesn't synchronize events from different threads. Thus, serial queue is needed.

mcleary:
Excellent, I will continue this pull request with a test application and I will try to add a new context type to the UI. Thanks for clarifying the lock-free/serial queues question.

wolfpld:
Regarding timestamp retrieval, it _might_ be best to execute host callback when events complete (`clSetEventCallback` or something like that). This is performed as a PCIe interrupt and was suggested by a friend at Nvidia. Doing so would eliminate CPU/GPU clock synchronization problems.

mcleary:
Using `clSetEventCallback` was my first choice when implementing this, however, in my experience, the Nvidia platform doesn't guarantee the callback will be called immediately after event completion. I have seen cases where callbacks were being called in intervals of 100ms. This might have changed so I guess it is worth another shot.

sumtime1987:
https://github.com/wolfpld/tracy/pull/31#issuecomment-634777257

mcleary:
After experimenting with `clSetEventCallback` I think my current strategy to synchronize the timestamps is working fine, or perhaps I am just finding it difficult to modify the file to account for a callback instead of an explicit call to `TracyCLCollect`. I still think it is a valid suggestion but I would like to prepare a test application to be added to this PR first.

wolfpld:
That's ok, even bad result is a worthwhile result when searching for the best solution.

mcleary:
As the next step towards having this merged, I've added a test OpenCL application. It is as simple as it can get because the purpose is just to demonstrate how to annotate OpenCL applications. 

I've added it under `examples` but I can change it if necessary.

PR #32, opened 2020-05-28
Added support for horizontal scrolling.

jmanc3:
I saw the branch on bitbucket where it didn't seem like the author was going to fix his pull request for adding horizontal scrolling, so I made this one that does the same thing but removes all the duplicated code.


wolfpld:
Thanks!

PR #33, opened 2020-05-28
Minor fixes

graydon:
Hi! Thanks so much for this profiler, it's fantastic!

Here are a few small fixes from my end: a little .gitignore expansion so it can be used as a submodule in an automake project, some macros to handle cases that aren't currently handled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,8 @@ manual/t*.bbl
 manual/t*.blg
 profiler/build/win32/packages
 profiler/build/win32/Tracy.aps
-
 # include the vcpkg install script but not the files it produces
 vcpkg/*
 !vcpkg/install_vcpkg_dependencies.bat
+.deps/
+.dirstamp
--- a/Tracy.hpp
+++ b/Tracy.hpp
@@ -17,8 +17,11 @@
 #define ZoneScopedNC(x,y)
 
 #define ZoneText(x,y)
+#define ZoneTextV(x,y,z)
 #define ZoneName(x,y)
+#define ZoneNameV(x,y,z)
 #define ZoneValue(x)
+#define ZoneValueV(x,y)
 
 #define FrameMark
 #define FrameMarkNamed(x)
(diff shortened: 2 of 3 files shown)


wolfpld:
Why is `-fPIC` needed?

The new macros require a description in the manual.

graydon:
Updated macro names (to end with `V` since it didn't seem like using `N` suffix was correct) and added docs to manual.

Removed `-fPIC` change: this was a misinterpretation of a funny bug I figured out the root cause of this morning (might warrant mention in manual or building FAQ? I can do so if there's a good place to write it up): GNU make's default `CC` variable is `cc` and its default `CXX` variable is `g++` not `c++`. This means that on a computer with both `clang/clang++` and `gcc/g++` installed, but `cc` and `c++` linked to `clang` and `clang++` (which is a pretty common configuration these days) we will do a mixed-compiler build: the C++ files get built with `g++` but the C files with `cc` which is `clang`. This in turn causes emission of some position-independent relocs in the C objects that require passing `-fPIC` to the final linking invocation of `g++`. Putting `-fPIC` in the `CFLAGS` was a bad workaround though; the correct fix is for the user to run `make CC=cc CXX=c++`, or possibly I could add `CC=cc` and `CXX=c++` to the Makefiles themselves?

(In general are patches to the docs concerning build/install/contribute instructions welcome? It took me a while to figure out a TeX distribution that could build the docs as well -- neither TeXLive nor MikTex seem to work -- finally found TeXmaker able to do so though I'm not sure if that's what you're using.)

wolfpld:
Changes look good. As for the compiler mishap, this seems to be a distribution or local system configuration problem. I verified that on my systems (Arch and Ubuntu) `cc`, `c++` and `g++` all point to gcc (clang is of course also installed). Therefore I'm not sure if this should be in the manual.

As for the patches to documentation in general, all improvements are open for discussion. I'm using MiKTeX windows distribution with Texmaker as an editor.

PR #36, opened 2020-06-03
Warning fixes

graydon:
Just a couple warning fixes -- a use-before-init warning on regs during cpuid and an ignored-result warning on fscanf in SysTime.


client/TracyProfiler.cpp
```diff
@@ -173,6 +173,10 @@ struct ThreadHandleWrapper
 #if defined __i386 || defined _M_IX86 || defined __x86_64__ || defined _M_X64
 static inline void CpuId( uint32_t* regs, uint32_t leaf )
 {
+    regs[0] = 0;
+    regs[1] = 0;
+    regs[2] = 0;
+    regs[3] = 0;
```
wolfpld, line 179: `memset` would be more appropriate.

graydon: Fixed.

PR #40, opened 2020-06-07
Direct3D 12 GPU Profiling

Xenonic:
This PR allows for minimal-markup Direct3D 12 GPU profiling support, mimicking the existing Vulkan GPU profiling API. Note that other implementations require additional markup and context tracking for individual command allocators, while this implementation does not.

- [X] Working prototype.
- [X] Completed Vulkan-like API.
- [X] Implement semi-automatic N-buffered synchronization.
- [X] Update manual.
- [X] Test colored zones.
- [X] Test multiple command queues.
- [X] Test small query sizes.
- [X] Test nested zones.
- [X] Test callstack path.
- [X] Test query overflow.
- [X] Test double and triple buffered rendering.
- [X] Test compute and copy queues.


Xenonic:
Due to a current limitation in the profiler, out-of-order execution is not realistically possible and has been reverted. Changes in the backend to make this possible are being looked into, and should not require any additional changes in this API.

PR #52, opened 2020-06-20
Add ___tracy_init_thread function

### LINKED ISSUE #50: [C API/UNIX] rpmalloc intializers will fail to run for non-main threads leading to SIGABRT
I’m using the C tracy API. As an experiment I called the first function I came across – `___tracy_alloc_srcloc` and the application immediately aborted. Attached is the stack-trace:

```
(gdb) bt
#0  0x000055555557fea3 in tracy::_memory_allocate_small (heap=0x0, size=32) at tracy/client/tracy_rpmalloc.cpp:1270
#1  0x0000555555580134 in tracy::_memory_allocate (heap=0x0, size=32) at tracy/client/tracy_rpmalloc.cpp:1343
#2  0x00005555555815af in tracy::rpmalloc (size=32) at tracy/client/tracy_rpmalloc.cpp:2135
#3  0x00005555555660a0 in tracy::tracy_malloc (size=32) at tracy/common/TracyAlloc.hpp:16
#4  0x000055555557ba51 in tracy::Profiler::AllocSourceLocation (function=0x55555561301f "test_fn", source=0x555555613013 "<no-source>", line=0) at tracy/client/TracyProfiler.hpp:489
#5  ___tracy_alloc_srcloc (line=0, source=0x555555613013 "<no-source>", function=0x55555561301f "test_fn") at tracy/client/TracyProfiler.cpp:3068
<snip>
#19 0x00007ffff7dccedd in start_thread () from /nix/store/jx19wa4xlh9n4324xdl9rjnykd19mmq3-glibc-2.30/lib/libpthread.so.0
#20 0x00007ffff7ce3a4f in clone () from /nix/store/jx19wa4xlh9n4324xdl9rjnykd19mmq3-glibc-2.30/lib/libc.so.6

```

TracyClient was built as such:

```
"g++" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-Wall" "-Wextra" "-DTRACY_ENABLE" "-o" "out/tracy/TracyClient.o" "-c" "tracy/TracyClient.cpp
"ar" "crs" "out/libtracy-client.a" "out/tracy/TracyClient.o"
```

And then linked directly to the test executable that invokes this function from a thread. The thread is created by a test framework.

---

From spelunking the code base I found that the `rpmalloc_thread_initialize` is set-up to be called through a TLS initializer – and indeed it is, at least for the main thread:

```
(gdb) br RPMallocInit::RPMallocInit()
Breakpoint 1 at 0x1e638: file tracy/client/TracyProfiler.cpp, line 143.
(gdb) run
Breakpoint 1, tracy::RPMallocInit::RPMallocInit (this=0x55555564da70 <tracy::s_rpmalloc_init>) at tracy/client/TracyProfiler.cpp:143
(gdb) bt
#0  tracy::RPMallocInit::RPMallocInit (this=0x55555564da70 <tracy::s_rpmalloc_init>) at tracy/client/TracyProfiler.cpp:143
#1  0x00005555555948b1 in __static_initialization_and_destruction_0 (__initialize_p=1, __priority=102) at tracy/client/TracyProfiler.cpp:987
#2  0x00005555555950c1 in _GLOBAL__sub_I.00102_TracyClient.cpp(void) () at tracy/libbacktrace/elf.cpp:3512
#3  0x0000555555612e85 in __libc_csu_init ()
#4  0x00007ffff7c0fd1a in __libc_start_main () from /nix/store/jx19wa4xlh9n4324xdl9rjnykd19mmq3-glibc-2.30/lib/libc.so.6
#5  0x000055555556556a in _start () at ../sysdeps/x86_64/start.S:120

```

However, it does not appear to be called again for the newly created threads. And that's why I ultimately encounter a crash.

---

The manual does not appear to indicate that the client library must be used in the main thread, or that there are any initialization routines for the user to call, or that it is necessary to specify certain defines (such as `ENABLE_PRELOAD`) to use the client library in a multi-threaded program.

nagisa:
This function would need to be called for each non-main thread before calling
the `___tracy_alloc_`, and possibly in the future, other `___tracy` functions.

Alternative way to achieve this could be initializing the allocator
transparently in the `___tracy_alloc_*` calls.

Fixes #50

### DIFF
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -3068,6 +3068,13 @@ TRACY_API void ___tracy_emit_message_appinfo( const char* txt, size_t size ) { t
 TRACY_API uint64_t ___tracy_alloc_srcloc( uint32_t line, const char* source, const char* function ) { return tracy::Profiler::AllocSourceLocation( line, source, function ); }
 TRACY_API uint64_t ___tracy_alloc_srcloc_name( uint32_t line, const char* source, const char* function, const char* name, size_t nameSz ) { return tracy::Profiler::AllocSourceLocation( line, source, function, name, nameSz ); }
 
+// thread_locals are not initialized on thread creation. At least on GNU/Linux. Instead they are
+// initialized on their first ODR-use. This means that the allocator is not automagically
+// initialized every time a thread is created. As thus, expose to the C API users a simple API to
+// call every time they create a thread. Here we can then put all sorts of per-thread
+// initialization.
+TRACY_API void ___tracy_init_thread(void) { (void)tracy::s_rpmalloc_thread_init; }
+
 #ifdef __cplusplus
 }
 #endif
--- a/TracyC.h
+++ b/TracyC.h
@@ -81,6 +81,7 @@ struct ___tracy_c_zone_context
 // This struct, as visible to user, is immutable, so treat it as if const was declared here.
 typedef /*const*/ struct ___tracy_c_zone_context TracyCZoneCtx;
 
+TRACY_API void ___tracy_init_thread(void);
 TRACY_API uint64_t ___tracy_alloc_srcloc( uint32_t line, const char* source, const char* function );
 TRACY_API uint64_t ___tracy_alloc_srcloc_name( uint32_t line, const char* source, const char* function, const char* name, size_t nameSz );
 
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -1477,6 +1477,9 @@ couleur=black!5,
 logo=\bcbombe
 ]{Important}
 Since you are directly calling the profiler functions here, you will need to take care of manually disabling the code, if the \texttt{TRACY\_ENABLE} macro is not defined.
+
+Additionally, make sure you have called \texttt{\_\_\_tracy\_init\_thread} before calling the
+\textttt{alloc} functions for the first time on a new non-main thread.
 \end{bclogo}
 
 \subsubsection{Memory profiling}
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -713,6 +713,7 @@ With the aforementioned steps you will be able to connect to the profiled progra
 Manual instrumentation is best started with adding markup to the main loop of the application, along with a few function that are called there. This will give you a rough outline of the function's time cost, which you may then further refine by instrumenting functions deeper in the call stack. Alternatively, automated sampling might guide you more quickly to places of interest.
 
 \subsection{Handling text strings}
+\label{textstrings}
 
 When dealing with Tracy macros, you will encounter two ways of providing string data to the profiler. In both cases you should pass \texttt{const char*} pointers, but there are differences in expected life-time of the pointed data.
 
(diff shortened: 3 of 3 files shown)


nagisa:
Will fix the manual after I receive some affirmative feedback that this is the way to go.

client/TracyProfiler.cpp
```diff
@@ -3068,6 +3068,13 @@ TRACY_API void ___tracy_emit_message_appinfo( const char* txt, size_t size ) { t
 TRACY_API uint64_t ___tracy_alloc_srcloc( uint32_t line, const char* source, const char* function ) { return tracy::Profiler::AllocSourceLocation( line, source, function ); }
 TRACY_API uint64_t ___tracy_alloc_srcloc_name( uint32_t line, const char* source, const char* function, const char* name, size_t nameSz ) { return tracy::Profiler::AllocSourceLocation( line, source, function, name, nameSz ); }
 
+// thread_locals are not initialized on thread creation. At least on GNU/Linux. Instead they are
```
wolfpld, line 3071: Yup. MSVC initializes on thread creation, gcc/clang on first use.

wolfpld:
I would like to know the bigger context of how this is meant to be used. While the proposed change is well thought-out by itself, it may not be needed at all.

In the existing code, srcloc allocations are performed after an item is received from the lock free queue, which already handles TLS initialization. This brings the question, why would the order of doing things be different in your use case? Essentially, I would like to see some (pseudo)code of the intended final API usage.

nagisa:
Right, a good question.

I’m implementing an interface of a larger observability tool. My interaction with the tool mostly involves implementing an interface like this (crudely rewritten in C++, any errors are my own):

```c++
class Subscriber {
    // called when span is entered
    virtual void on_enter(Id& span_id, Context& ctx) = 0;
    virtual void on_exit(Id& span_id, Context& ctx) = 0;
    virtual void on_event(...) = 0;
}
```

various information about what span is comes from the `context` argument, however what `context` contains for e.g. a file name is not necessarily a `const char*` with a static lifetime (i.e. they do not necessarily live throughout the program execution), but rather something that can live for as long as `ctx`. `ctx` lifetime ends after `on_exit` is called.

Furthermore, the values are also stored as something resembling a [string_view](https://en.cppreference.com/w/cpp/string/basic_string_view) – and so the strings are not null-terminated. For these reasons I had to resort to using `___tracy_alloc_srcloc_name` in the first place, somewhat like this:

```c++
MySubscriber::on_enter(Id& span_id, Context& ctx) {
    auto span = ctx.span();
    string_view file = span.file();
    string_view name = span.name();
    auto line = span.line();
    unique_ptr c_file = make_c_string(file);
    ___tracy_alloc_srcloc_name(...); // crashes currently, because at this point rpmalloc might not be initialized for the calling thread.
}
```

There isn’t a convenient way for me to "just" call `___tracy_init_thread` on thread creation per se, but I can adjust `on_enter` to look more like this:

```c++
MySubscriber::on_enter(Id& span_id, Context& ctx) {
    // ...
    thread_local bool initialized_tracy_for_thread = false;
    if (!initialized_tracy_for_thread) { 
        ___tracy_init_thread(); 
        intiialized_tracy_for_thread = true;
    }
    ___tracy_alloc_srcloc_name(...)
}
```

I wouldn’t call this super convenient per-se, and I would prefer it if we adjusted `___tracy_alloc_srcloc_*` to do the initialization by itself, always. The change to do so can be as simple as changing `___tracy_alloc_srcloc_*` to refer to `tracy::s_rpmalloc_thread_init` much like how the propoosed `___tracy_init_thread` works now.

However, I looked at the generated code for such a change and it does introduce a function call in the generated assembly. As thus I wasn’t very sure if such a change would be acceptable. I doubt the overhead would would be perceptible, but I don’t know much enough about thread locals in C++ to say for sure.

For that reason I went with a separate function and tried to come up with more reasons to do it that way; like, for instance, allowing us to add additional per-thread initialization in the future.

wolfpld:
Ok, that makes sense. Regarding the manual - calling this function should only be needed when writing bindings, and not during typical usage. As such, please create another section on that topic (and maybe also document `___tracy_alloc_srcloc_*`).

BTW, the last code snippet may be pointless - a whole TLS data block is initialized on access to any TLS variable, so after you have checked the value of `initialized_tracy_for_thread` (but before you have called `___tracy_init_thread`) rpmalloc is already initialized.

nagisa:
I went ahead and adjusted the manual to extract documentation about the binding-oriented functions into a dedicated section.

wolfpld:
Typo: `fucntions return an uint64_t`

Issue #54, opened 2020-06-20
[import-chrome] how to approach importing metrics?


ethercrow:
I'd like to add a couple of global metrics like heap size and CPU load in my chrome-tracing files. Is there an appropriate API I should use in `import-chrome.cpp` to import those?

wolfpld:
No API other than what is available. How are these metrics structured? Are they appropriate for plotting (i.e. are time and value pairs)?

ethercrow:
The section `Counter Events` in the spec looks appropriate https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.msg3086636uqhttps://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.msg3086636uq

A counter event has a timestamp and one or more `(metric name, metric value)` pairs.

wolfpld:
Extended the API to accept plot data.

PR #58, opened 2020-06-21
Import chrome metrics

ethercrow:
Fixes https://github.com/wolfpld/tracy/issues/54

So I can see that data is imported but is not visible in a chart. I think I need to somehow convert timestamps for plot data, because ultimately these are stored in `PlotItem.time` which is `Int48`. Are those nanoseconds from the beginning of the file? Nanoseconds since Epoch don't fit.

The metric values, at least the min and max of those are displayed correctly in the tooltip.

Screen Shot 2020-06-21 at 11 33 38 AM

### DIFF
--- a/import-chrome/src/import-chrome.cpp
+++ b/import-chrome/src/import-chrome.cpp
@@ -118,6 +118,44 @@ int main( int argc, char** argv )
                 v["name"].get<std::string>()
             } );
         }
+        else if( type == "C" )
+        {
+            auto timestamp = int64_t( v["ts"].get<double>() * 1000 );
+            for( auto& kv : v["args"].items() )
+            {
+                bool plotFound = false;
+                auto& metricName = kv.key();
+                auto dataPoint = std::make_pair( timestamp, kv.value().get<double>() );
+
+                // The input file is assumed to have only very few metrics,
+                // so iterating through plots is not a problem.
+                for( auto& plot : plots )
+                {
+                    if( plot.name == metricName )
+                    {
+                        plot.data.emplace_back( dataPoint );
+                        plotFound = true;
+                        break;
+                    }
+                }
+                if( !plotFound )
+                {
+                    auto formatting = tracy::PlotValueFormatting::Number;
+
+                    // NOTE: With C++20 one could say metricName.ends_with( "_bytes" ) instead of rfind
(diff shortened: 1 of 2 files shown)


server/TracyWorker.cpp
```diff
@@ -400,8 +400,8 @@ Worker::Worker( const std::string& program, const std::vector<ImportEventTimelin
         plot->type = PlotType::User;
         plot->format = v.format;
 
-        double min = v.data.begin()->first;
-        double max = v.data.begin()->first;
+        double min = v.data.begin()->second;
```
ethercrow, line 403: This is supposed to be min and max value, not min and max timestamp, right?

wolfpld: Yup.

ethercrow:
```
Appeared at: 15:15:02
Last event at: 15:15:03
```

Hm, are these 15 *hours* into the profile?

wolfpld:
I think I forgot about adjustment of timestamps here:

```c++
    uint64_t mts = 0;
    if( !timeline.empty() )
    {
        mts = timeline[0].timestamp;
    }
    if( !messages.empty() )
    {
        if( mts > messages[0].timestamp ) mts = messages[0].timestamp;
    }
    for( auto& v : timeline ) v.timestamp -= mts;
    for( auto& v : messages ) v.timestamp -= mts;
```

Easy to fix.

ethercrow:
Thank you for the hint, seems to be working now:

Screen Shot 2020-06-21 at 1 37 38 PM

import-chrome/src/import-chrome.cpp
```diff
@@ -135,6 +173,9 @@ int main( int argc, char** argv )
     }
     for( auto& v : timeline ) v.timestamp -= mts;
     for( auto& v : messages ) v.timestamp -= mts;
+    for( auto& plot : plots )
+      for( auto& v : plot.data )
+        v.first -= mts;
```
wolfpld, line 178: Please use braces in non-oneliner case.

import-chrome/src/import-chrome.cpp
```diff
@@ -135,6 +173,9 @@ int main( int argc, char** argv )
     }
```
wolfpld, line 173: You need to add a check for minimum value of plots here.

PR #60, opened 2020-06-21
Check for null FILE handles in ReportTopology

nagisa:
On Linux `/sys` is not guaranteed to exist. When it doesn't, tracy
would crash with a SIGSEGV due to a nullptr dereference. Not actually
sure how well tracy clients will handle the case where this information
is missing, though...


wolfpld:
Data should be send only if fully valid.

nagisa:
Does something like this work, then?

PR #63, opened 2020-07-01
Added continuous scrolling.

jmanc3:
Added a feature I thought would be useful which is to be able to continuously scroll the view automatically so you always see the latest events. As soon as the user starts interacting with the view, the continuous scroll turns off and can be toggled on again by pressing the button. You can check out the vid below to see what I mean more clearly. 
https://drive.google.com/file/d/1UFwSTbA4FCKE18jUKeUnoj0WWvx9LByJ/view?usp=sharing

### DIFF
--- a/server/TracyView.cpp
+++ b/server/TracyView.cpp
@@ -146,6 +146,8 @@ View::View( const char* addr, int port, ImFont* fixedWidth, ImFont* smallFont, I
     s_instance = this;
 
     InitTextEditor( fixedWidth );
+
+    m_continuousScrollingFirstFrame = true;
 }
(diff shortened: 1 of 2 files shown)


wolfpld:
Looks cool, but is should be integrated with the already existing pause/resume functionality. The current mode of operation (showing last 3 frames) should be preserved and continuous scroll could be enabled when user zooms-out, pans the view, etc.

jmanc3:
Ah. Actually, I should've taken a closer look at all of Tracy's features. I hadn't seen that you could follow the latest 3 frames already!

PR #66, opened 2020-07-04
CSV export of zone statistics from .tracy file

kubouch:
This is a proof-of-concept implementation of a simple statistics export from a .tracy file into CSV. I'm not sure if it's welcome as a part of the repository but I thought it would be useful.

The idea is to implement a command-line equivalent of the "Statistics" and "Find zone" GUI windows that will output results in a CSV format. So, you can filter a zone's name and it will spit out columns such as mean, min, max, std, etc. for each matching zone.  The `--unwrap` flag reports each zone event separately, similar to the zone list dropdown in the "Find zone" window, so you can make you own statistics. The output is printed into stdout - you can redirect it into a file or use something like [nushell](https://github.com/nushell/nushell) to process it further.

You can toggle the self time by passing the `--self` flag. Introducing similar switches to get the running or cumulative time should be simple but it would require duplicating a lot of code from TracyView.cpp which I purposefully avoided.

The code is based mostly on `tracy::View::DrawStatistics()` and `tracy::View::DrawFindZone()` and uses [cxxopts](https://github.com/jarro2783/cxxopts) to handle CLI options.

#### TODO
- [x] Add an actual Windows build - I literally never program on Windows so it requires some setting up
- [x] Add CI build to AppVeyor & GitHub Workflow
- [x] Add manal entry
- [x] Add AUTHORS entry
- [x] Use getopt instead of cxxopts


wolfpld:
The idea seems fine.

Have you considered using getopt instead of cxxopts?

Note that CI failing on MSVC and Linux is a consequence of https://github.com/microsoft/vcpkg/issues/12062 and https://github.com/Cyan4973/xxHash/issues/412, respectively.

kubouch:
At one point I wanted multiple arguments for one option which I'm not sure I could do with getopt. But I removed it so I can switch to getopt since it's ~10x smaller than cxxopts.

wolfpld:
A portable getopt implementation is already included in the `capture` utility. Should be moved to (a subdirectory in) the root directory if you want to use it.

kubouch:
So I managed to get it working on Visual Studio 2019 and also fixed the getopt include for capture after moving it to root. This is my first time building C++ on Windows so please bear with me.

kubouch:
> So I managed to get it working on Visual Studio 2019 and also fixed the getopt include for capture after moving it to root. This is my first time building C++ on Windows so please bear with me.

There might be some extra dependencies since I copy-pasted them from the capture utility. However, I haven't found a convenient way to detect it.

PR #80, opened 2020-07-30
Manual: Clarify meaning of `size` argument in strings.

jimblandy:


wolfpld:
I think "should not include" wording would be less ambiguous.

PR #81, opened 2020-07-31
Factor out Unix build logic into shared files.

jimblandy:
This pulls out most of the repeated code from the various tools' build.mk files into two shared files, `common/unix.mk` and `common/src-from-vcxproj.mk`.


jimblandy:
Surely this one is right.

PR #85, opened 2020-08-07
Fix timestamp calculation in chrome tracing importer

kaiscode:
The current Chrome Tracing importer incorrectly checks if the "dur" field is an object when calculating the end timestamp. The "dur" field is specified to be microseconds, not an object. This results in correctly formatted chrome traces resulting in Tracy files where each zone is zero duration.

See Chrome Tracing "spec" here: https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.lpfof2aylapb


wolfpld:
Pinging @ethercrow

ethercrow:
Hello, I’m not at the computer currently, will take a look next week.

PR #86, opened 2020-08-07
Making find- and stat-zones stripes interleave always by changing DrawStripedRect

simonvanbernem:
Two `bool` parameters are added to DrawStripedRect:

`fix_stripes_in_screen_space `will align the stripes of the rect in screen space. This means that any stripe rects' stripes will always exactly "line up", no matter how they are positioned to each other. When it is false, the behaviour remains unchanged.
`inverted `will flip the empty and filled aread of a striped rectangle.

When drawing the striped zones in TracyView.hpp, one zone uses inverted and one doesn't. This leads to them interleaving, which I think looks way nicer than the random partial overlaps, that happen currently.

This is what it looks like now:
grafik

I tried to match the code style, but chose more verbose names and added some comments to make it understandable. Feel free to delete the comments and pick shorter names to get it in line with the rest of the codebase.

### DIFF
--- a/server/TracyImGui.hpp
+++ b/server/TracyImGui.hpp
@@ -184,9 +184,9 @@ namespace tracy
         const auto rw = x1 - x0;
         const auto rh = y1 - y0;
         const auto cnt = int( ( rh + rw + sw*2 ) / ( sw*2 ) );
-        auto v0 = ImVec2( x0, y0 - rw);
+        auto v0 = ImVec2( x0, y0 - rw );
 
-        if (fix_stripes_in_screen_space)
+        if ( fix_stripes_in_screen_space )
         {
             const auto window_width = double( ImGui::GetWindowHeight() );
             const auto flipped_v0y = window_width - v0.y; //we transform into a y-is-up coordinate space to achieve upper-left to lower-right stripes. If we didn't, we would calculate values for lower-left to upper-right
--- a/server/TracyImGui.hpp
+++ b/server/TracyImGui.hpp
@@ -188,16 +188,16 @@ namespace tracy
 
         if (fix_stripes_in_screen_space)
         {
-            const auto window_width = double(ImGui::GetWindowHeight());
+            const auto window_width = double( ImGui::GetWindowHeight() );
             const auto flipped_v0y = window_width - v0.y; //we transform into a y-is-up coordinate space to achieve upper-left to lower-right stripes. If we didn't, we would calculate values for lower-left to upper-right
 
             const auto manhatten_distance = x0 + flipped_v0y;
-            const auto in_multiples_of_2_times_sw = int(manhatten_distance / (sw * 2));
+            const auto in_multiples_of_2_times_sw = int( manhatten_distance / ( sw*2 ) );
         
-            const auto floored_manhatten_distance = double(in_multiples_of_2_times_sw * sw * 2); //floor in terms of 2 * stripe width
+            const auto floored_manhatten_distance = double( in_multiples_of_2_times_sw*sw*2 ); //floor in terms of 2 * stripe width
 
-            const auto corrected_flipped_v0y = (floored_manhatten_distance - x0); //we transform back into y-is-down imgui space
-            v0.y = window_width - corrected_flipped_v0y - double(inverted * sw);
+            const auto corrected_flipped_v0y = ( floored_manhatten_distance - x0 ); //the corrected (floored) y respects the position of the stripes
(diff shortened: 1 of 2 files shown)


server/TracyView.cpp
```diff
@@ -3245,7 +3245,7 @@ void View::DrawZones()
     {
         const auto s = std::min( m_setRangePopup.min, m_setRangePopup.max );
         const auto e = std::max( m_setRangePopup.min, m_setRangePopup.max );
-        DrawStripedRect( draw, wpos.x + ( s - m_vd.zvStart ) * pxns, linepos.y, wpos.x + ( e - m_vd.zvStart ) * pxns, linepos.y + lineh, 5 * ImGui::GetTextLineHeight() / 15.f, 0x55DD8888 );
+        DrawStripedRect( draw, wpos.x + ( s - m_vd.zvStart ) * pxns, linepos.y, wpos.x + ( e - m_vd.zvStart ) * pxns, linepos.y + lineh, 5 * ImGui::GetTextLineHeight() / 15.f, 0x55DD8888, true, false );
```
wolfpld, line 3248: Is aligning in this place beneficial?

server/TracyImGui.hpp
```diff
@@ -183,8 +183,23 @@ namespace tracy
...
         const auto cnt = int( ( rh + rw + sw*2 ) / ( sw*2 ) );
+        auto v0 = ImVec2( x0, y0 - rw );
+
+        if ( fix_stripes_in_screen_space )
+        {
+            const auto window_height = double( ImGui::GetWindowHeight() );
+            const auto flipped_v0y = window_height - v0.y; //we transform into a y-is-up coordinate space to achieve upper-left to lower-right stripes. If we didn't, we would calculate values for lower-left to upper-right
+
+            const auto manhatten_distance = x0 + flipped_v0y;
+            const auto in_multiples_of_2_times_sw = int( manhatten_distance / ( sw*2 ) );
```
wolfpld, line 195: I'm not sure about this flooring. It will likely break things when `dpiScale` (main.cpp) isn't 1.

server/TracyView.cpp
simonvanbernem: Not really, I just changed it for consistency, but since it has a different column spacing anyway, it doesn't really matter.

server/TracyImGui.hpp
simonvanbernem: I'm not understanding how that would happen. This floores a mulitplier, not an actual screen-distance, so I can't see how DPI settings would cause a problem here. Also: I am using two monitors with different DPIs and didn't obsever any problems.

wolfpld: Dynamic adapting to DPI changes in runtime is not available in Tracy (it's still ongoing work in ImGui). What you see as adapting is scaling of the window bitmap on the OS level (assuming you're using Windows).

I'm not entirely sure what this variable is used for, so this might be nitpicking. Anyway, this is a close-up of stripes with `dpiScale` set to 1:

obraz

And this is with `dpiScale` forced to 1.35:

obraz

There is some misalignment, which _might_ be caused by ignoring the fractional part. I don't know.

simonvanbernem: Wait, but the flooring here is purely to find the place where a stripe should begin. What happens is basically:

`const auto in_multiples_of_2_times_sw = int( manhatten_distance / ( sw*2 ) );`
Calculate how many stripes fit in the manhatten distance. (e.g. 16.32). If this value was fractional, it means that the current manhatten distance is _inside_ of a stripe. We want to place the vertices on the edges of the stripes. We are on an edge if the multiple has no fractional part, so we floor to get to the lower edge of the stripe we are currently in.

`const auto floored_manhatten_distance = double( in_multiples_of_2_times_sw*sw*2 ); //floor in terms of 2 * stripe width`
Then, we multiply by sw*2 again, which gives us the manhattan distance that marks the start of the (in this case) 16th stripe. Note that this value might have a fractional part! If e.g. sw is 2.3, then the lower edge of the 16th stripe will be at 2.3 * 16 = 36.8.

(I just noticed the comment is on the wrong line, it should be on the previous)

And after that, we subtract x from the manhattan distance to get just the y part, and flip the y back to imgui space.

So the flooring is the necessary logical operation of finding the lower edge of the stripe. Without flooring this algorithm would not work.

simonvanbernem: You are of course right that there still is some misalignment, but I am quite sure that the flooring is not the cause.

PR #88, opened 2020-08-08
Add D3D11 Gpu Context/Zone

hugoam:
I've removed the dust on this old PR (https://bitbucket.org/Nukem9/tracy/commits/227d3e9d2b1518031af9fb0114abd37a3f7d4f7f)  and made it work with the current version of Tracy, because I needed to profile a D3D11 based application
Would you be interested in merging it in master ?
It's not completely clean yet (mostly I the macros are untested as I use the structures directly) so in any case don't press merge just yet :D But I'm curious to know if you're interested in upstreaming it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,36 @@
+cmake_minimum_required(VERSION 3.0)
+
+project(tracy)
+
+set(CMAKE_CXX_STANDARD 11)
+
+option(TRACY_ENABLE "Profiling is enabled" ON)
+option(TRACY_ON_DEMAND "Profiling enabled on demand" OFF)
+option(TRACY_NO_EXIT "Client executable does not exit until all profile data is sent to server" OFF)
+option(TRACY_NO_BROADCAST "Disable client discovery by broadcast to local network" OFF)
+
+add_library(tracy_client TracyClient.cpp)
+
+macro(set_option OPTION)
+    if(${OPTION})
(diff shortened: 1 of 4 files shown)


wolfpld:
Yes, of course I'm interested in merging this.

There was also a second pull request for D3D11: https://bitbucket.org/wolfpld/tracy/pull-requests/34

Before this can be merged, the manual and AUTHORS list has to be updated.

hugoam:
The second one you linked seems a bit simpler as it doesn't seem to use a disjoint query for each scope, but that could be potentially a problem maybe. Although the disjoint handling doesn't look perfect in the current PR either, and I'm not sure it would behave correctly in case of frequency change, I think it's hard to get perfectly right

hugoam:
Do you know what would be the correct way to handle D3D11 disjoint event ? The current code just does not handle them at all. Looking at other profiling libraries, one way is to do a cpu/gpu resync at that point. Can we use the `calibrate` event for that ? Looking at some example code, I think that can only be done through a pipeline stall though, is that imaginable ?

wolfpld:
> Do you know what would be the correct way to handle D3D11 disjoint event ?

I don't use D3D.

> Looking at other profiling libraries, one way is to do a cpu/gpu resync at that point. Can we use the GpuCalibration event for that ?

Not really. Calibration events are for recalibration, not for resyncing.

PR #94, opened 2020-08-14
Add GTK3 in the list of build requirements on Unix

mcleary:
When I tried building Tracy on another machine other than my laptop I found that GTK2.0 is needed, however, there is no mention of this dependency in the manual.

This PR adds this reference.

### DIFF
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install linux libraries
       if: ${{ matrix.os == 'ubuntu-20.04' }}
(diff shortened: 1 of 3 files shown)


wolfpld:
Is this a strict requirement for gtk2, or will gtk3 also work? I can't check this at the moment.

mcleary:
I believe gtk2 is a requirement due to this line in `build.mk`

```
else
	SRC2 += ../../../nfd/nfd_gtk.c
	INCLUDES += $(shell pkg-config --cflags gtk+-2.0)
	LIBS += $(shell pkg-config --libs gtk+-2.0) -lGL
endif
```

wolfpld:
This is a mistake then. It should be gtk3, as specified at https://github.com/mlabbe/nativefiledialog/:

> GTK3 dialog on Linux.

mcleary:
I will try to update the build script as well

PR #101, opened 2020-08-27
Add support for TRACY_ONLY_IPV4 macro to exclude listening on IPv6

graydon:
When running a tracy client inside kubernetes, IPv6 port-forwarding is not available, only IPv4. It is therefore desirable to disable listening on IPv6 and use only IPv4. This patch enables such behaviour.


wolfpld:
`TRACY_ONLY_IPV4` should be available not only as a compile-time option, but also as a runtime env variable check.

graydon:
Added env var (as well as for previous flag TRACY_ONLY_LOCALHOST).

graydon:
(updated to fix build failure)

PR #115, opened 2020-10-01
Make Tracy run natively on Wayland

annacrombie:
Hello, this commit lets me run Tracy on wayland.  

The display server detection is a bit basic, maybe you would prefer an environment variable switch?

GLFW gives errors about the following unsupported features:
```
Error 65548: Wayland: The platform does not support setting the input focus
Error 65548: Wayland: The platform does not support setting the window icon
Error 65548: Wayland: The platform does not support setting the window position
```

I had to add `glfwWindowHint(GLFW_ALPHA_BITS, 0);` to prevent transparent windows in Tracy from making the whole window transparent

glfw issue: https://github.com/glfw/glfw/issues/1434

I also had to remove `glfwWindowHint(GLFW_VISIBLE, 0);` to avoid this error:
```
xdg_wm_base@9: error 3: xdg_surface must not have a buffer at creation
```
which prevents Tracy from running.

glfw issue: https://github.com/glfw/glfw/issues/1268

### DIFF
--- a/profiler/build/unix/build.mk
+++ b/profiler/build/unix/build.mk
@@ -5,11 +5,12 @@ INCLUDES := $(shell pkg-config --cflags glfw3 freetype2 capstone) -I../../../img
 LIBS := $(shell pkg-config --libs glfw3 freetype2 capstone) -lpthread -ldl
 
 DISPLAY_SERVER := X11
-libwayland_client := $(shell pkg-config --libs --silence-errors wayland-client)
-ifeq ($(.SHELLSTATUS),0)
+
+ifdef TRACY_USE_WAYLAND
 	DISPLAY_SERVER := WAYLAND
-	LIBS += $(libwayland_client)
+	LIBS += $(shell pkg-config --libs wayland-client)
 endif
+
 CXXFLAGS += -D"DISPLAY_SERVER_$(DISPLAY_SERVER)"
 
 PROJECT := Tracy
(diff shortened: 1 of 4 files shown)


profiler/build/unix/build.mk
```diff
@@ -3,6 +3,15 @@ CXXFLAGS := $(CFLAGS) -std=c++17
 DEFINES += -DIMGUI_IMPL_OPENGL_LOADER_GL3W
 INCLUDES := $(shell pkg-config --cflags glfw3 freetype2 capstone) -I../../../imgui -I../../libs/gl3w
 LIBS := $(shell pkg-config --libs glfw3 freetype2 capstone) -lpthread -ldl
+
+DISPLAY_SERVER := X11
+libwayland_client := $(shell pkg-config --libs --silence-errors wayland-client)
```
wolfpld, line 8: Selection of target display server has to be more explicit. I have `wayland-client` installed through package dependencies even on headless machines.

annacrombie: I figured.  What about checking for `$WAYLAND_DISPLAY`?  This would only work inside a wayland session though, obviously.

annacrombie: Maybe there should be an override environment variable, like `TRACY_LINUX_DISPLAY_SERVER` + a check for `$WAYLAND_DISPLAY` if it is unset.

wolfpld: Depending only on `WAYLAND_DISPLAY` may have some unforseen consequences (e.g. some tool in the build chain may want to connect to the server (due to reasons), which isn't present, resulting in build failure). The second option looks good, but maybe with a shorter variable name. Make sure it's properly documented.

PR #118, opened 2020-10-28
Avoid caching very large files (that crash the UI and bloat traces).

benvanik:
[Unity builds](https://onqtam.com/programming/2018-07-07-unity-builds/) or generated source files can exceed a reasonable size and cause Tracy to crash in various ways (source file tokenization/imgui handling/etc). This change bails when a file is "too large" (heuristically chosen for the largest file I was able to load and similar to what tools like vscode do when warning about large files). If there's a place this could be added as a configuration variable (`TRACY_MAXIMUM_SOURCE_FILE_SIZE`) I can replace the inlined magic number with that to allow user control.

### DIFF
--- a/server/TracyWorker.cpp
+++ b/server/TracyWorker.cpp
@@ -49,6 +49,8 @@ static bool SourceFileValid( const char* fn, uint64_t olderThan )
     struct stat buf;
     if( stat( fn, &buf ) == 0 && ( buf.st_mode & S_IFREG ) != 0 )
     {
+        // Avoid caching very large files (such as those in unity builds).
+        if ( buf.st_size >= 2 * 1024 * 1024 ) return false;
         return (uint64_t)buf.st_mtime < olderThan;
     }
     return false;


benvanik:
Also, if you think this is worth calling out in the manual I'm happy to add it there, however I feel it's mostly a corner case where just not having a crash is nice :)

wolfpld:
There's definitely a need for some kind of global profiler configuration (e.g. https://bitbucket.org/wolfpld/tracy/issues/94/save-options), but it still has to be done. This would be a good place for modification of the maximum size.

A mention in the manual about the limit would be good to have.

Do you have call stacks for the crashes you are experiencing?

benvanik:
Fantastic question - you made me try this again without my change and I noticed that it may not have had anything to do with the size itself, but the tokenization code accessing out of bounds in cases where the file doesn't end in a linefeed:
https://github.com/wolfpld/tracy/blob/3976408550b73ed10e7a1e5eeb5e4fdf7b50265c/server/TracySourceView.cpp#L489-L490

I think I was conflating the crash with an out of memory resizing the lines/token list when instead it was bailing from the loop when `end - m_data < sz` failed and `end` was then pointing after the last valid byte of the read data. Tokenize handles this correctly (bumping begin until end) but the code immediately after that unconditionally dereferences `end`:
https://github.com/wolfpld/tracy/blob/3976408550b73ed10e7a1e5eeb5e4fdf7b50265c/server/TracySourceView.cpp#L491

it looks like the `if( end - m_data == sz ) break;` should be guarding those dereferences too.

The test files I was working with were manually truncated to various sizes, and it looks like the larger ones were just by chance all cut off mid-line with no trailing newline!

Guarding the dereferences causes even my 200MB monster file to load in tracy and work quite well. I'll drop this PR and send one for that fix instead :)

wolfpld:
Fixed in 0c489564.

BTW, not ending source files with a newline was undefined behavior in C++03, but isn't in C++11.

PR #123, opened 2020-11-12
Fix sampling on Android with default `su` command.

bjacob:
This makes sampling work for me on two configurations where it currently did not:
* Android 10, on a Pixel4 device (aarch64).
* Debian (linux kernel 5.7) (x86-64).

This also unifies most of the Linux code between the Android and other Linux cases. The only function that I didn't unify was `ProcessTraceLines`.

The main thing that this PR really does is to introduce some shared helper functions, `WriteBufferToSystemFile` and `ReadSystemFileWithFunction`, and use them for all the file I/O that we perform here. These functions try regular file I/O first, and if that fails, they do the file I/O in with `fork`/`exec`/{`su`,`sudo`}. Compared to the existing code, the main differences are:
1. All file I/O goes through these same helpers, while in the existing code, it was done differently for different files. Some files that did need the `fork`/`exec`/{`su`,`sudo`} dance did not use it, as noted below.
2. The `fork`/`exec`/{`su`,`sudo`} dance itself is fixed/updated to work on recent Linux distributions and Android versions.

This PR also removes the 'payload' stuff that we used to have on Android. It was an intermediate hoop used to perform such privileged file IO and it's not needed anymore.

Here are some things that weren't working in the old code, that this fixes:
* The `perf_event_open` syscall now tends to require `0` to be written to `/proc/sys/kernel/perf_event_paranoid`. We could probably also have tried to avoid that by deferring `perf_event_open` to a subprocess with root permissions, but that would have been much harder than deferring file I/O  (would have required a separate binary).
* Many current linux distributions don't let you do `su` to promote to root, instead the gateway for that is `sudo`.
* Current Android is the opposite: it only has `su`, not `sudo`. However, the syntax to run a `command` as root is `su root command` instead of `su -c command`.
* Another issue is that different unices have a different take on whether the `command` passed to `su`/`sudo` is going to be interpreted by a shell, or if it's going to be just `exec`'d. That means that the current code's reliance on the `>` redirection operator was not working everywhere - at least it wasn't working on Android. So instead I rewrote this code to use the `dd` command, which is able to perform any redirection to/from a file passed to it as a command-line argument.
* The old code was only doign the `fork`/`exec`/`su` dance for some of the file I/O, not for all of it. It was not doing it when it tried to read `/proc/$tid`. On current Android, `/proc` is mounted by default with `hidepid=2` meaning that most users can't see most of `/proc/$tid`, so now `/proc` is part of what also needs the `fork`/`exec`/`su` dance.

Notes:
* While debugging this, I found myself wishing that this code had some logging of error codes returned by the syscalls/IO that it performs, so I added some.
* This code (both the existing code and this new code) is relying on `su` or `sudo` obtaining root permissions without any user interaction. I made this more explicit (`sudo -n`) and added a very verbose error message that should be printed in the most common case where this is the issue, with some practical steps for the user to take.
* I did wonder at some point whether all that was worth it or we should just require the profilee to run as root. But it turns out that Android graphical applications can't run as root, even if you launch them from a root shell. So there really is a strong need for performing this dance. Fortunately, on Android it goes smootly: on any rooted android device, the `su` command succeeds without interaction, it never asks for authentication.
* It turns out that some of our I/O is failing because we try to read /proc/$tid for some TIDs of threads that don't exist anymore by the time we are launching the profiler UI. This seems like a preexisting issue, I let it continue to fail silently, except that now the `dd` command spews a line on stderr when that happens.
* I removed the hard requirement of `CLOCK_MONOTONIC_RAW` and added fallback to `CLOCK_MONOTONIC` primarily because it's one of the things that I ran into while debugging this and wondering why sampling wasn't working for me. This turned out not to be the issue (all recent linux have `CLOCK_MONOTONIC_RAW`) but I also think it's better this way. There isn't a very significant difference between the two for our purposes; `CLOCK_MONOTONIC_RAW` is only useful for long running processes (e.g. on a server) whose lifetime my extend across discrete time adjustments.
* The existing code was doing some `pthread_setschedparam(..., SCHED_FIFO, ...)`. I have preserved that where I didn't have a reason to touch it, but as I wanted different file IO code paths to share common helpers, it was not trivial to preserve that stricly - so the current PR does not preserve that 1:1. Let me know if you think it's important to preserve and we can do it with some complexification of these shared helpers.


wolfpld:
There are several problems with this PR.

> This makes sampling work for me on two configurations where it currently did not:
>
> *    Android 10, on a Pixel4 device (aarch64).
> *   Debian (linux kernel 5.7) (x86-64).

Both these platforms (i.e. Android 10 on arm64 and recent Linux on x64) are working as intended for many people. I fail to see what's the change that made sampling work for you here.

> This also unifies most of the Linux code between the Android and other Linux cases.

Well, there are some very good reasons why these code paths are separate.

> These functions try regular file I/O first, and if that fails, they do the file I/O in with fork/exec/{su,sudo}.

The manual explicitly states that the profiled application has to be run with elevated privileges. The profiled application should not try to gain more privileges than it already has.

> This PR also removes the 'payload' stuff that we used to have on Android. It was an intermediate hoop used to perform such privileged file IO and it's not needed anymore.

It wasn't. The manual states its purpose.

> The perf_event_open syscall now tends to require 0 to be written to /proc/sys/kernel/perf_event_paranoid.

Yes, as described in the manual.

> Many current linux distributions don't let you do su to promote to root, instead the gateway for that is sudo.

Then these distributions are broken by design and shouldn't be used.

> Current Android is the opposite: it only has su, not sudo. However, the syntax to run a command as root is su root command instead of su -c command.

```
% adb shell
$ su --help
MagiskSU 20.4(20400)

Usage: su [options] [-] [user [argument...]]

Options:
  -c, --command COMMAND         pass COMMAND to the invoked shell
(...)
```

> Another issue is that different unices have a different take on whether the command passed to su/sudo is going to be interpreted by a shell, or if it's going to be just exec'd.

The su man page explicitly states that command passed with `-c` will execute in shell.

> On current Android, /proc is mounted by default with hidepid=2 meaning that most users can't see most of /proc/$tid,

Again, as described in manual.

> While debugging this, I found myself wishing that this code had some logging of error codes returned by the syscalls/IO that it performs, so I added some.

The profiler must not output anything on stdout/stderr.

> on Android it goes smootly: on any rooted android device, the su command succeeds without interaction, it never asks for authentication.

It may ask the user to allow privilege elevation.

> It turns out that some of our I/O is failing because we try to read /proc/$tid for some TIDs of threads that don't exist anymore by the time we are launching the profiler UI. This seems like a preexisting issue

Yes, this works as it works because the kernel people don't care about sane interfaces, or documentation, only that their beloved `perf` works.

> There isn't a very significant difference between the two for our purposes; CLOCK_MONOTONIC_RAW is only useful for long running processes (e.g. on a server) whose lifetime my extend across discrete time adjustments.

You're making a couple wrong assumptions here:

* You think that Tracy is only used for short captures. It isn't.
* You believe that clock adjustments will happen only if a process is running for a long time. This is wrong. There is no relation between process lifetime and clock adjustments being made or not. You want to play russian roulette here and hope bad things won't happen to you.
* By changing the used clock you have desynchronized sampling from the rest of profiler.

> Let me know if you think it's important to preserve

Everything is to be assumed to be important, otherwise it wouldn't be there.


To sum this up, I'd advise you to carefully read the manual, then identify the exact issues you have with sampling (if any), then talk about how to approach fixing them.

bjacob:
Thanks for the helpful review and explanations.

Based on your comments, I will restart this patch from scratch to do only the minimal changes that I really need.

Here is the main problem that I need solving: the different `su` command on current vanilla Android. I understand that your Android device has [Magisk](https://magiskmanager.com/) but unless you want to make Magisk a requirement for Tracy on Android, we need to support the `su` command that is part of the default Android builds, and it says:
```
$ adb shell su --help
usage: su [WHO [COMMAND...]]

Switch to WHO (default 'root') and run the given COMMAND (default sh).

WHO is a comma-separated list of user, group, and supplementary groups
in that order.
```
So we really need to write `su root command` instead of `su -c command` on Android. Trying `-c` gives:
```
$ adb shell su -c ls
su: invalid uid/gid '-c'
```
And this `su: invalid uid/gid '-c'` is currently spewn by Tracy on Android.

I would like to better understand what you mean with
> The manual explicitly states that the profiled application has to be run with elevated privileges. The profiled application should not try to gain more privileges than it already has.

I have 2 questions about this:
1. If the profiled application is assumed to already run with elevated privileges, then why is TracySysTrace.cpp dealing with `su` at all?
2. Android graphical applications ("intents", as opposed to command-line binaries) never run as root. Even if one spawns then with `am` from a root shell, they still run unprivileged. Is tracy not supporting profiling of such applications / is tracy only supporting command-line programs on Android? My PR was in principle making that work, allowing Tracy to profile GUI apps.

Depending on the answers to the above, if Tracy wants to support profiling an unprivileged process on Linux, we may need to discuss the topic of `su`'ing on certain major Linux distributions default installs. This is explained for instance [here](https://www.wikihow.com/Become-Root-in-Ubuntu). When a non-root user runs `su -c command`, that fails on many current linux distro default installs because the root user doesn't have a password by default, so it's impossible to `su` into. Whence the need for `sudo`. As explained in that link, it's a simple step to create a root password, making `su` work as expected, but the majority of Linux users at this point probably don't have a root password.

I also have a question: if `su` is really expected to elevate privileges, then it needs authentication; what's the intended user experience here, and is this relying on some persistence of authentication so that `su` does not ask for the root password again every time it's run (which is multiple times per profiling session)?

wolfpld:
> Here is the main problem that I need solving: the different su command on current vanilla Android.

Yes, the various `su` implementations currently available should be supported. The current implementation and its limitations are an artifact of the setup I have, and not a deliberate dependency on Magisk.

> If the profiled application is assumed to already run with elevated privileges, then why is TracySysTrace.cpp dealing with su at all?

That's the only way of obtaining privilege elevation on Android. With careful examination of the source code you will see that the `su` code path is not present on non-Android Linux.

> Android graphical applications ("intents", as opposed to command-line binaries) never run as root. Even if one spawns them with am from a root shell, they still run unprivileged. Is tracy not supporting profiling of such applications / is tracy only supporting command-line programs on Android? My PR was in principle making that work, allowing Tracy to profile GUI apps.

See above. I use Tracy only in GUI applications on Android.

> if Tracy wants to support profiling an unprivileged process on Linux

I'm not interested in that, as the user can easily gain the needed privileges in a number of ways. How this is performed is not something a profiler should be concerned about. Android is of course an exception, due to reasons you stated.

bjacob:
Got it, thanks! I agree it's fine to require the profiled application to run as root on desktop Linux, just not on Android; I had indeed missed the fact that `su` is only used in the Android code paths.

With that I think I'll be able to make a minimal patch doing just the support for Android's default `su` syntax.

Question:
> Yes, the various su implementations currently available should be supported. 
To confirm, you want code to gracefully handle both `su root command` and `su -c command` flavors? Would you suggest trying one and then falling back to the other? It's going to add a little bit of code to implement this, so I want to check upfront that you think it's the better support/complexity compromise, than just supporting the default Android flavor (`su root command`).

wolfpld:
The `-c` option can be easily emulated by just running the shell with appropriate parameters, so only one invocation schema should be enough. Something like `su root sh -c command params`.

bjacob:
Sorry, I don't understand! On a `su` flavor that expects any `command` argument to be prepended with `-c`, it seems that
```
su root sh -c command params
```
would not interprete `sh` as the command, as it's not prepended with `-c` ?

bjacob:
I think the problem you're addressing with `su root sh -c command params` is the interpretation of the command by a shell, as opposed to raw `exec`'ing it. But it's not a problem that we have, because as you noted earlier, `su` actually interpretes its command on a shell (and I checked, even the Android one does).  The problem that we need solving here is the discrepancy between the traditional `su` that requires a `-c` flag before the command, and the Android `su that does not know about the `-c` flag.

wolfpld:
Yeah, disregard the previous comment, it doesn't do what I thought it does. Now that I have checked, it would seem that passing `-c` is required in my version of `su` for this to work (both on desktop and in Magisk):

```
chiron:/ # ls /mnt
appfuse asec expand media_rw obb product runtime sdcard secure user vendor
chiron:/ # su root ls /mnt
chiron:/ # su -c ls /mnt
appfuse asec expand media_rw obb product runtime sdcard secure user vendor
chiron:/ #
```

```
chiron:/ $ whoami
shell
chiron:/ $ su root ls
chiron:/ # whoami
root
```

bjacob:
Thanks for running this! I'm curious why the non-`-c` commands seem to terminate silently for you as opposed to returning an error - can you check if they at least have a return code, e.g.
```
su root ls /mnt || echo failure
```
?

I just need to know so I can implement the fallback logic accordingly (assuming that you want a fallback logic implemented).

wolfpld:
The command succeeds, as the user is changed. Hence, the return code is zero.

When a shell is executed with arguments, it expects that a file with a shell script is provided:

```
# zsh /usr/bin/ls
/usr/bin/ls:1: parse error near `)'
# su root /usr/bin/ls
/usr/bin/ls:1: parse error near `)'
```

Magisk su is very simple, so it probably skips some of that logic: https://github.com/topjohnwu/MagiskSU/blob/master/su.c

By looking at `to.command` there it is evident that `-c` passed to `su` is just forwarded to `sh`.

bjacob:
Here's a whole new implementation, restarted from scratch after our conversation.

Actually, Android's default `su` command does `exec` its command argument, does not pass it to `sh -c`, so we do have to insert `sh -c` in our command line in that case:

```
$ adb shell
flame:/ # su root 'echo hello | cat'
su: failed to exec echo hello | cat: No such file or directory
```

I was confused earlier because at some point, I ran this experiement instead:
```
$ adb shell su root 'echo hello | cat'
hello
```

But that was not sufficiently escaping the command string, as `adb shell`'s implementation eats one layer of escaping. Properly escaping reproduces the error:

```
$ adb shell su root '"echo hello | cat"'
su: failed to exec echo hello | cat: No such file or directory
```

Note: In one instance, it seemed to make sense to just do exec `dd` as opposed to creating a shell just to perform a redirection operator, so I did that, but let me know if you'd

client/TracySysTrace.cpp
```diff
@@ -796,11 +797,180 @@ static void SetupSampling( int64_t& samplingPeriod )
...
+            waitpid( pid, nullptr, 0 );
+        }
+    }
+    return success;
+}
+
+// Enum identifying a method for running a command as root.
+enum class HowToRunAsRoot {
+    // Our process is already root (getuid()==0). Nothing else is needed.
+    AlreadyRunningAsRoot,
```
wolfpld, line 873: Is there a scenario in which this case would be possible?

client/TracySysTrace.cpp
```diff
@@ -796,11 +797,180 @@ static void SetupSampling( int64_t& samplingPeriod )
...
+    // The way to run a command as root is: `su -c 'command'`.
+    // In this case, `command` is interpreted by a shell (not just exec'd).
+    SuDashC,
+    // The way to run a command as root is: `su root 'command'`.
+    // In this case, `command` is just exec'd. If any interpretation by a
+    // shell is needed, it needs to be done explicitly, like:
+    //   `su root sh -c 'command'`.
+    SuRoot,
+    // We don't know how to run a command as root on this device.
+    DontKnow
```
wolfpld, line 883: This should be handled as a failure to setup kernel tracing, in `SysTraceStart()`.

client/TracySysTrace.cpp
```diff
@@ -796,11 +797,180 @@ static void SetupSampling( int64_t& samplingPeriod )
...
+            dup2( write_end, STDOUT_FILENO );
+            close( write_end );
+            execlp( "su", "su", flag, "id", "-u", nullptr );
+            exit( EXIT_FAILURE );
+        }
+        else if( pid > 0 )
+        {
+            // parent
+            close( write_end );
+            char buf[8] = { 0 };
```
wolfpld, line 858: Use just `{}`, there's no need to explicitly init the first element to 0 (it will be initialized implicitly, like the rest of values in buffer).

wolfpld:
In addition to the comments above, please adapt code style (spaces, braces, etc.) to the one used in the rest of the code. A bit terser function/enum/variable names would also be welcome.

Using `dd` should be fine.

client/TracySysTrace.cpp
bjacob: Expanded the comment to explain that scenario.

client/TracySysTrace.cpp
bjacob: Done.

client/TracySysTrace.cpp
bjacob: Done.

bjacob:
> In addition to the comments above, please adapt code style (spaces, braces, etc.) to the one used in the rest of the code. A bit terser function/enum/variable names would also be welcome.

Done.

> Using `dd` should be fine.

Actually, I noticed stderr messages from `dd` saying `Text file busy`. This seems as if we might have a race with another process concurrently accessing this file, conceivably from not properly waiting for the child process to terminate, however I didn't find that trivial to debug (a slightly less trivial use of `waitpid` here did not make a difference). So given the potential bug here, I reverted it to the "cat > /data/systrace" approach, just prepended with "sh -c" so that we get a shell to interprete the ">" so that my earlier objection to that approach does not apply. In the old code, using the "su -c" command was relying on "su" to perform the same thing (exec sh) anyway.

bjacob:
One more thing: since our approach inherently relies on trying some `su` commands that might be incorrect w.r.t. the current `su` implementation, it is to be expected that there will be some `su` error messages on `stderr`. To hide that noise (and also because you mentioned that this code shouldn't output anything to stderr), I added:
```
            int null_fd = open( "/dev/null", O_WRONLY );
            dup2( null_fd, STDERR_FILENO );
```
in `TrySuCommandFlag`.

wolfpld:
I was thinking if the `su` check could be simplified and it turns out that both forms of invocation are supported by a "proper" `su` command (both the Magisk one and the coreutils one):

```
chiron:/ $ whoami
shell
chiron:/ $ su -c 'whoami'
root
chiron:/ $ su root sh -c 'whoami'
root
```

I think this PR can be reduced to just adding `root sh` as the first arguments of the `su` invocation. The case of already being root can be ignored, as the `su` will be essentially no-op then.

bjacob:
That's a great simplification. PR updated!

bjacob:
thanks for merging! btw, i thought you'd use "squash and merge" - i was keeping some micro history here only for this PR's discussion purposes, but up to you!

bjacob:
Thanks for merging this!

PR #129, opened 2020-11-19
Ensure that mappings have read permission before decoding symbols and reading code.

### LINKED ISSUE #125: Sampling on Android: few samples and no asm view.
Attached a `.tracy` trace saved from the Tracy profiler UI: [low-samping-freq-and-no-asm.tracy.zip](https://github.com/wolfpld/tracy/files/5556324/low-samping-freq-and-no-asm.tracy.zip)

**What looks like not working as intended:**

1. I can't see a disassembly view, only a C++ source view.

2. There seem to be few samples in here. Most 'source views' don't show any "time percentage" left margin. Those that do have a "time percentage" left margin tend to have only few samples, as in the attached screenshot which says 3 samples.

3. It keeps saying '0x0' in the line at the top of the source view, as if it couldn't read a symbol address (?), see screenshot below.

Screen Shot 2020-11-17 at 4 29 57 PM

**Details about the workload:**

Upon request I could try to share steps to reproduce that trace, but that would be a bit complex. The workload is a neural network inference using [IREE](https://github.com/google/iree). This is running on a Pixel4 Android 10 device, the CPU architecture is aarch64.

The workload is Tracy-instrumented as follows: source: [tracing.h](https://github.com/google/iree/blob/main/iree/base/tracing.h), [tracing.cc](https://github.com/google/iree/blob/main/iree/base/tracing.cc) .

Regarding the "no asm view" part of this issue, a relevant fact about this workload might be that it involves code compiled by direct usage of LLVM , not compiled from a C or C++ source.

@benvanik has had complete success with Tracy with the same workload but running on Windows. So if the problem is on our end then it has to be somehow in a part of our code that's different on android vs windows (?)

bjacob:
Fixes #125


client/TracyProfiler.cpp
```diff
@@ -2907,8 +2915,195 @@ void Profiler::HandleSymbolQuery( uint64_t symbol )
...
+        if (!GetMappingsRange(mappings, start_address, end_address, &start_iter, &end_iter))
+        {
+            // Still not mapped? This shouldn't happen...
+            return false;
+        }
+    }
+    if (end_iter < start_iter) {
+        abort();
+    }
+    int k = 0;
```
bjacob, line 3085: oops, leftover debug helpers here, will remove

client/TracyProfiler.cpp
```diff
@@ -2907,8 +2915,195 @@ void Profiler::HandleSymbolQuery( uint64_t symbol )
...
+    mapping->perm_r = true;
+    return true;
+}
+
+// Attempts to make the specified address range readable.
+bool EnsureReadable( std::uintptr_t start_address, std::uintptr_t end_address )
+{
+    // Lock ensuring reentrancy of this function.
+    static std::mutex s_mutex;
+    const std::lock_guard<std::mutex> lock( s_mutex );
```
wolfpld, line 3066: This lock is not needed.

client/TracyProfiler.cpp
```diff
@@ -32,6 +32,14 @@
 #  include <sys/syscall.h>
 #endif
 
+#ifdef __ANDROID__
+#  include <sys/mman.h>
+#  include <cstdio>
+#  include <cstdint>
```
wolfpld, line 38: Use C headers (e.g. `stdio.h`) and `std::`-less types.

client/TracyProfiler.cpp
```diff
@@ -2907,8 +2915,195 @@ void Profiler::HandleSymbolQuery( uint64_t symbol )
...
+
+// Parses /proc/self/maps returning a vector<MappingInfo>.
+// /proc/self/maps is assumed to be sorted by ascending address, so the resulting
+// vector is sorted by ascending address too.
+MappingInfoVector ParseMappings()
+{
+    MappingInfoVector result;
+    FILE* file = fopen( "/proc/self/maps", "r" );
+    if( !file ) return result;
+    char line[256];
```
wolfpld, line 2955: I'd use at least 1KB here.

client/TracyProfiler.cpp
```diff
@@ -2907,8 +2915,195 @@ void Profiler::HandleSymbolQuery( uint64_t symbol )
...
+    }
+    return true;
+}
+                      
+
+// Attempts to make the specified `mapping` readable if it isn't already.
+// Returns true if and only if the mapping is readable.
+bool EnsureReadable( MappingInfo* mapping )
+{
+    if( mapping->perm_r )
```
wolfpld, line 3035: The function interface tells me that `mapping` may be `nullptr` here. Change it to a reference to ensure a valid object is always passed to the function.

client/TracyProfiler.cpp
```diff
@@ -2907,8 +2915,195 @@ void Profiler::HandleSymbolQuery( uint64_t symbol )
...
+        {
+            // Still not mapped? This shouldn't happen...
+            return false;
+        }
+    }
+    if (end_iter < start_iter) {
+        abort();
+    }
+    int k = 0;
+    for (auto it = start_iter; it != end_iter; ++it)
```
wolfpld, line 3086: I don't think it would be possible to have one symbol span multiple memory mappings?

bjacob: ah, i was wondering about that. thinking more about it, i agree, and it does result in a great simplification.

client/TracyProfiler.cpp
bjacob: done

client/TracyProfiler.cpp
bjacob: done

client/TracyProfiler.cpp
bjacob: done

client/TracyProfiler.cpp
bjacob: done (wasn't sure if more than 1 thread might call this, and there was some mutex stuff in this file).

bjacob:
I've made a few further cleanups / simplifications to this code now.

bjacob:
I've significantly updated this PR. As I was looking into the next issues preventing Tracy from showing all of the symbols in my workload, I realized that they were just more of the same. Namely, we were also running into the same issue in TracyCallstack.cpp, in `DecodeSymbolAddress`, so I added a call to `EnsureReadable` there. This required moving code to a place where it would be available to both `TracyCallstack.cpp` and `TracyProfiler.cpp`. Since the latter `#includes` TracyCallstack.hpp, I figured that I could move `EnsureReadable` to TrackCallstack.  Let me know if you would prefer another location (should it be its own file? I'm a little scared of touching Tracy's build system.)

client/TracyProfiler.cpp
```diff
@@ -2889,9 +2897,162 @@ void Profiler::HandleParameter( uint64_t payload )
...
+inline MappingInfo* LookUpMapping(std::vector<MappingInfo>& mappings, uintptr_t address)
+{
+    // We assume mappings to be sorted by address, as /proc/self/maps seems to be.
+    // Construct a MappingInfo just for the purpose of using std::lower_bound.
+    MappingInfo needle;
+    needle.start_address = address;
+    needle.end_address = address;
+    // Comparison function for std::lower_bound. Returns true if all addresses in `m1`
+    // are lower than all addresses in `m2`.
+    auto Compare = []( const MappingInfo& m1, const MappingInfo& m2 ) {
```
wolfpld, line 2971: This should be `( const MappingInfo& m1, uintptr_t m2 )`. Then you don't need to construct `needle` struct above, and just pass `address` to `std::lower_bound`.

bjacob: Thanks a lot for this tip, I didn't know about this overload of `std::lower_bound`, the code is simpler this way.

PR #130, opened 2020-11-20
Sort threads by tid

eklitzke:
I'm profiling an application that has hundreds of threads, and this change makes it a lot easier for me to use Tracy because it makes it a lot easier for me to find threads to select/deselect.

### DIFF
--- a/server/TracyView.cpp
+++ b/server/TracyView.cpp
@@ -8992,7 +8992,6 @@ void View::DrawOptions()
                     m_threadOrder.insert( m_threadOrder.begin() + target, srcval );
                     m_threadOrder.erase( m_threadOrder.begin() + source );
                 }
-                SortThreads();
             }
         }
         ImGui::TreePop();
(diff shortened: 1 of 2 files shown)


wolfpld:
Please make it controllable by a checkbox (off by default) in the options menu. Actions that modify thread order on user input should disable it.

wolfpld:
The variable should be stored in `ViewData` (`TracyViewData.hpp`) and have proper save/load support in `TracyUserData.cpp`.

benvanik:
Having a UI option for thread sorting behavior would be useful for other situations, too! I've found the current sorting (reverse creation order?) hard to deal with, as in this example even though the main thread is created first and the workers are created in order they appear reversed in the UI: 
image

I haven't tried this PR and it may give me the order I'm looking for, but even if not if the sorting UI/setting gets added it'll be easier to add in additional modes (ascending/descending creation order, TID, name, etc).

PR #132, opened 2020-11-23
Fixes following PR #129

bjacob:
I added these inline keywords at some intermediate step where this code was in a header file, but now that it's back to this `.cpp` file they are not needed anymore.

### DIFF
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -2898,7 +2898,6 @@ void Profiler::HandleParameter( uint64_t payload )
 }
 
 #ifdef __ANDROID__
-namespace {
 // Implementation helpers of EnsureReadable(address).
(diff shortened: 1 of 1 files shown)


wolfpld:
While you're at it, you may add `static` specifiers to these functions.

bjacob:
OK, I've pushed another round of fixes.

Regarding `static`, notice how this was in an anonymous namespace. However, I understand that the style in this code was to use `static` instead, so I changed to that. The only exception if the `struct MappingInfo`, I keep an anonymous namespace just around it.

I noticed an actual bug, which this new commit fixes. Since `std::lower_bound` returns "the first element that is not less-than", my condition `iter->end_address <= address` was always false (otherwise, `*iter` would be less-than `address`). The relevant condition to check was `iter->start_address > address`, i.e. check if this "first element that is not less-than" might already have a too-high start address to be a match.

PR #137, opened 2020-11-27
Adding ZoneColor to set a dynamic color override to an existing zone.

benvanik:


common/TracyQueue.hpp
```diff
@@ -124,6 +125,10 @@ struct QueueZoneValidation
     uint32_t id;
 };
 
+struct QueueZoneColor {
+  uint32_t value;
```
wolfpld, line 129: This should be 3 bytes (alpha is ignored).

server/TracyEvent.hpp
```diff
@@ -207,6 +207,7 @@ struct ZoneExtra
     Int24 callstack;
     StringIdx text;
     StringIdx name;
+    uint32_t color;
```
wolfpld, line 210: Same here, use `Int24`.

server/TracyView.cpp
```diff
@@ -16561,8 +16561,12 @@ uint32_t View::GetRawZoneColor( const ZoneEvent& ev, uint64_t thread, int depth
...
     const auto sl = ev.SrcLoc();
     const auto& srcloc = m_worker.GetSourceLocation( sl );
-    const auto color = srcloc.color;
-    if( color != 0 && !m_vd.forceColors ) return color | 0xFF000000;
+    if( !m_vd.forceColors ) {
+      const auto custom_color = m_worker.GetZoneExtra( ev ).color;
+      if( custom_color != 0 ) return custom_color | 0xFF000000;
+      const auto color = srcloc.color;
+      if( color != 0 ) return color | 0xFF000000;
+    }
```
wolfpld, line 16569: Formatting.

server/TracyWorker.cpp
```diff
@@ -4738,6 +4741,22 @@ void Worker::ProcessZoneName()
     extra.name = StringIdx( GetSingleStringIdx() );
 }
 
+void Worker::ProcessZoneColor( const QueueZoneColor& ev )
+{
+    auto td = RetrieveThread( m_threadCtx );
+    if( !td || td->stack.empty() || td->nextZoneId != td->zoneIdStack.back() )
+    {
+        ZoneNameFailure( m_threadCtx );
```
wolfpld, line 4749: This is wrong.

server/TracyView.cpp
benvanik: sorry not seeing it - what would you change? (I tried to match the existing formatting)

server/TracyView.cpp
```diff
@@ -16561,8 +16561,12 @@ uint32_t View::GetRawZoneColor( const ZoneEvent& ev, uint64_t thread, int depth
 {
     const auto sl = ev.SrcLoc();
     const auto& srcloc = m_worker.GetSourceLocation( sl );
-    const auto color = srcloc.color;
-    if( color != 0 && !m_vd.forceColors ) return color | 0xFF000000;
+    if( !m_vd.forceColors ) {
+      const auto custom_color = m_worker.GetZoneExtra( ev ).color;
```
wolfpld, line 16565: `HasZoneExtra( ev ) && GetZoneExtra( ev )`

server/TracyView.cpp
wolfpld: Move the first { brace to a new line and use 4 spaces to indent instead of 2.

benvanik: ahh nevermind 4 space indent - got it!

common/TracyQueue.hpp
benvanik: ah - I didn't see Int24 being used with color [here](https://github.com/wolfpld/tracy/blob/4eb489508def47ca4c64af099050a3607245e0c6/server%2FTracyEvent.hpp#L167) so thought there may have been a reason - changed!

wolfpld: Source locations are low volume, so it doesn't matter that much if everything is optimally packed. Zone colors can be sent for each zone.

benvanik: It looks like Int24 isn't accessible from common/ (it's defined in server/TracyEvent.hpp) - should I just use uint8_t[3] and cast it around?

benvanik: nevermind, I see SourceLocation has a split r g b - I'll do the same.

PR #139, opened 2020-11-27
Remove potential warnings

Sahnvour:
This PR helps adoption of Tracy in codebases with high level of warnings.

The semi-colons inside macro definitions can cause warnings when the user would add one themself in client code, creating an empty statement. Not using semi-colons after macros in client code will fix it, but will break code formatting.


wolfpld:
How the existing code, which may not include semicolons at end of macros, is supposed to keep working?

Sahnvour:
It's not ?

I mean, if there is a precise point in having some of the macros with semicolons, and some without, I didn't get it and I'd like you to explain it.
This was surprising to me, and I don't think it's usual in either C nor C++ to have this mix, it feels like an error. If people end up depending on errors, it's legit to introduce a "breaking change" when correcting them. Especially given that it is not that hard to automate.

wolfpld:
> It's not ?

Why? Why should existing, working and valid code be broken?

> I don't think it's usual in either C nor C++ to have this mix

Of course. Nevertheless, currently both variants of using these macros are legitimate, and choosing only one to be valid *will* break someone's code.

> If people end up depending on errors, it's legit to introduce a "breaking change" when correcting them.

I agree. But please explain how empty statements are errors (preferably by citing a rule from the C++ standard). If I understand correctly, we're still talking about what is effectively fashion advice dispensed by a compiler in the form of warnings, which you opted to treat as errors. This could as well be about function bodies being too long, or variable names being too short.

Sahnvour:
In my opinion, empty statements should be considered an error (as in, erroneous code, wether the compiler emits a warning or an error is another story). Granted, they don't have any effect most of the time, but since they have no use I think watching out for them helps catch real errors.
For example, a `for ();` loop can happily compile while being wrong. This is more than fashion or style preference.

I think the public part of a project/library should strive for the broadest compability, not hindering above-normal use of compiler warnings.

> Why? Why should existing, working and valid code be broken?

To allow more people using this library. Backward compatibility is definitely important, but in this case the change is minor and largely manageable. Are there actual examples of code relying on these semicolons or is it only speculation ?

If that's still not acceptable to you, I would propose doing :

```cpp
#ifdef TRACY_REMOVE_TRAILING_SEMICOLONS
#define TRACY_TRAILING_SEMICOLON
#else
#define TRACY_TRAILING_SEMICOLON ;
#endif

#define TracyLockable( type, varname ) type varname TRACY_TRAILING_SEMICOLON
// etc.
```
This is not pretty but at least we can have both sides. It can also for example stay during a release cycle while advertising the change, so that people have time to adapt without being blocked.

wolfpld:
This won't be merged, sorry.

Hirezias:
Do you have some rational for that ?
Code quality and poor integration is clearly the main Tracy issue and Sahnvour proposal is little but free improvement in that field.

wolfpld:
For starters, when asked about a C++ standard rule quote, don't give me your opinion on what should or shouldn't be in the standard. That's just unprofessional.

More importantly, I have come to realization that this doesn't actually fix anything. Take a look at this code:
```c++
void f() { ZoneScoped }
```

This expands to:
```c++
void f() { whatever(); }
```

What you'd want will change it to:
```c++
void f() { ZoneScoped; }
```

Which will expand to:
```c++
void f() { whatever(); }
```

All is fine, right? Now consider a case where `TRACY_ENABLE` is *not* defined. The expansion will be:
```c++
void f() { ; }
```

Which is exactly what you want to avoid, but you can't anymore.

Lastly, if you feel that code quality or integration is bad in Tracy, I'm sure there are many other profilers for you to choose from.

Hirezias:
Producing a non-empty whit no effect instruction is not a great deal for a "professional" programmer.

PR #142, opened 2020-12-01
Optimize module caching on Windows

GenuineAster:
We are running Tracy in a context that hosts a .NET Core runtime, and that loads a million DLLs and a bunch of which are unknown and don't have file names, or can cause stack traces that don't have a module loaded (I'm not too sure on that one though).

Either way, this change allowed us to run tracy with stack trace collection in real-time, I think it's about 2-3x faster than before but our startup is non-deterministic so it's a bit hard to tell.


wolfpld:
- There is a lot of code duplication around `s_modCache->push_next()`. Please extract the common part out of the branches.
- You'll need to adapt to the existing code style.

PR #161, opened 2021-01-05
Fix shutdown hang (resolves #160)

### LINKED ISSUE #160: Tracy 0.7.4 can hang on shutdown on Linux as root
1. Create an application that calls `tracy::StartupProfiler()`, creates at least one profiler zone, waits a bit of time and then calls `tracy::ShutdownProfiler()`
2. Run the application as root on Linux (I was running on aarch64) so that systrace gets initialized

Expected: the application exits
Actual: the application hangs on shutdown

jkriegshauser:
This change resolves the shutdown hang in issue #160

### DIFF
--- a/client/TracySysTrace.cpp
+++ b/client/TracySysTrace.cpp
@@ -706,10 +706,10 @@ static void SetupSampling( int64_t& samplingPeriod )
         for( int i=0; i<s_numCpus; i++ ) s_ring[i].Enable();
         for(;;)
         {
+            if( !traceActive.load( std::memory_order_relaxed ) ) break;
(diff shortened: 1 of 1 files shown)


wolfpld:
How does the proposed fix work? `traceActive` is already checked out of the loop in line 789.

jkriegshauser:
> How does the proposed fix work? `traceActive` is already checked out of the loop in line 789.

Ah, I missed that. That fix isn't in the v0.7.4 tagged release, which is what I was working against.

wolfpld:
Fix already on master.

PR #165, opened 2021-01-16
D3D12 transient zones

Xenonic:
This PR implements the new transient GPU zones feature for Direct3D 12, as well as removing the old callstack code to follow the new conventions.


wolfpld:
Thanks, there are some people who will be very happy with these changes :)

PR #167, opened 2021-01-21
Fix call to clRetainEvent inside assert in OpenCLCtxScope

johner:
In `OpenCLCtxScope::SetEvent`, the necessary `clRetainEvent`
call was inside an assert, thus never called if `NDEBUG`
was defined.  This change asserts only on the return value
of the function, as in other parts of the code.

### DIFF
--- a/TracyOpenCL.hpp
+++ b/TracyOpenCL.hpp
@@ -125,7 +125,8 @@ namespace tracy {
                 if (eventInfo.phase == EventPhase::End)
                 {
                     // Done with the event, so release it
-                    assert(clReleaseEvent(event) == CL_SUCCESS);
+                    err = clReleaseEvent(event);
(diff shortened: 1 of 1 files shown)


wolfpld:
@mcleary can you take a quick look at this?

mcleary:
@johner You are absolutely right. Could you please do the same in `OpenCLCtxScope::Collect`, more specifically: https://github.com/wolfpld/tracy/blob/2eae80b9dc7d2b3de71c35329a2592b7ed11e770/TracyOpenCL.hpp#L128 

Thanks for pointing this out.

johner:
Whoops, thank you.  (The `setEvent` was causing a crash, this one just leaks so I missed it!)  Pushed that change as well.

PR #168, opened 2021-01-22
Improve import

### LINKED ISSUE #163: File name in viewer window title missing for chrome-imported traces
Currently when you open a saved trace (and you have multiple Tracy client windows open) it isn't possible to tell at a glance which file is open in which window.
image

This request is so that the filename could be displayed in the window title bar such as:
`Tracy Profiler 0.7.4 - saved_trace.tracy`

jkriegshauser:
Resolves #163 by adding the filename to the titlebar if the imported .tracy file does not have a name (which is the case for imported Chrome traces).

Also carries over thread names from Chrome traces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 x64
 Release
 Debug
+_build
+_compiler
+tools/*
 *.d
 *.o
 *.so
@@ -28,3 +31,7 @@ vcpkg/*
 !vcpkg/install_vcpkg_dependencies.bat
 .deps/
 .dirstamp
+.vscode/
+
+/_*/**
+/**/__pycache__/**
--- a/client/TracySysTrace.cpp
+++ b/client/TracySysTrace.cpp
@@ -706,10 +706,10 @@ static void SetupSampling( int64_t& samplingPeriod )
         for( int i=0; i<s_numCpus; i++ ) s_ring[i].Enable();
         for(;;)
         {
-            if( !traceActive.load( std::memory_order_relaxed ) ) break;
             bool hadData = false;
             for( int i=0; i<s_numCpus; i++ )
             {
+                if( !traceActive.load( std::memory_order_relaxed ) ) break;
                 if( !s_ring[i].HasData() ) continue;
                 hadData = true;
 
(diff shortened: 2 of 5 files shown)


wolfpld:
Instead of setting `m_captureName`, if empty on file load, the chrome-import version of `Worker` ctor should be setting it appropriately. This seems to be an omission, as `m_captureProgram` is being set properly there. For reference, this seems to be the missing part:

https://github.com/wolfpld/tracy/blob/20b2be3b777a47f784bf241df08762af2fd15d37/server/TracyWorker.cpp#L2696-L2702

Depending of how much extra work you want to put into this, you may simply assign `m_captureProgram` to `m_captureName`, or do some timestamp reading in the import utility and add it to the name string, etc.

jkriegshauser:
> Instead of setting m_captureName, if empty on file load, the chrome-import version of Worker ctor should be setting it appropriately.

I can do that; seems simple enough. But it also makes sense to fallback to _something_ to display in the title bar even if the capture name is empty. So keeping the `m_captureName` set-on-empty still makes sense, no?

wolfpld:
I guess it does.

pt., 22 sty 2021, 03:13 użytkownik jkriegshauser <notifications@github.com>
napisał:

jkriegshauser:
Alright, the latest push should write the output name into the file, but keeps the input name as the 'program' so that it can be viewed in the Info pane.

wolfpld:
It seems like keeping the output trace name is important for you - can you say why? The convention in Tracy is that traces are identified by the name of captured executable (in this case it should be the source chrome tracing file name), so that you can freely move or rename the trace file.

jkriegshauser:
Sure, basically it’s that the import-chrome utility is generally manually invoked and provides an opportunity to be more descriptive with the file name. In our case, the chrome profiles generated have an automated name with date/time such as `ct 2021-01-22 05-55-12.json` which when we invoke import-chrome we would change to something more descriptive like `load scene no cache.tracy` which is the name we would want displayed. If we wanted to use the same original file name then that would be what is displayed but it provides more options without having to rename the original file before doing import-chrome (which still needs an output file name specified). Also the original name is still preserved in my changes so that it can be referenced in the Info pane in the client.

PR #171, opened 2021-01-26
Improve OpenCL time calibration

johner:
When creating an `OpenCLCtx`, the current OpenCL time calibration calls `Profiler::GetTime` for the host timestamp, then the `GetDeviceTimestamp` function to compute a device time, by recording the end time of a small memory transfer operation.  The problem is to set up that transfer, the `GetDeviceTimestamp` function makes several OpenCL calls to create a command queue, memory buffer, etc., all of which can take > 100 microseconds.  This bakes in an extra delay in host/device timestamps that effectively moves all OpenCL timestamps forward in time by that delay, leading to most OpenCL events having negative Execution Delays.

This commit is closer to the Vulkan approach: all the code in the constructor, the memory operation is queued as non-blocking, and the host timestamp is read immediately after the `clWaitForEvents` call returns, which happens as soon as the event's status becomes `CL_COMPLETE` (similar to `vkQueueWaitIdle`).  As this status change is the same time that is read with the call to `clGetEventProfilingInfo(writeBufferEvent, CL_PROFILING_COMMAND_END, ....`, I think that is the closest these timestamps can be read together.

With this change I get Execution Delays of around +15 μs for the OpenCLVectorAdd example's kernel launches on Linux/RTX 2080 Super, which I believe is a more typical delay for kernel launch.

Also, the previous code was storing the `m_hostStartTime` into the context structure's `gpuTime` member, which is confusing, and required a `TimestampOffset` function to correct, even though I believe the profiler itself is doing the same offset internally if the `gpuTime` properly holds the GPU timestamp.

Further questions:
- Are the `m_hostStartTime` and `m_deviceStartTime` members still needed, or can we switch to local `tcpu` and `tgpu` variables as in the Vulkan code?
- Should the OpenCLCtx constructor take an optional `cl_command_queue` argument and avoid creating its own?  The current approach is probably fine, but the calling code has likely already created a command queue itself with the `CL_QUEUE_PROFILING_ENABLE` property, so why not allow passing it in? Avoids creating another queue, and seems the device timestamp should come from the same queue actually used in the application?

### DIFF
--- a/TracyOpenCL.hpp
+++ b/TracyOpenCL.hpp
@@ -66,13 +66,37 @@ namespace tracy {
         {
             assert(m_contextId != 255);
 
+            cl_int err = CL_SUCCESS;
+            cl_command_queue queue = clCreateCommandQueue(context, device, CL_QUEUE_PROFILING_ENABLE, &err);
+            assert(err == CL_SUCCESS);
+            uint32_t dummyValue = 42;
+            cl_mem dummyBuffer = clCreateBuffer(context, CL_MEM_WRITE_ONLY, sizeof(uint32_t), nullptr, &err);
+            assert(err == CL_SUCCESS);
+            cl_event writeBufferEvent;
+            err = clEnqueueWriteBuffer(queue, dummyBuffer, CL_FALSE, 0, sizeof(uint32_t), &dummyValue, 0, nullptr, &writeBufferEvent);
+            assert(err == CL_SUCCESS);
+            err = clWaitForEvents(1, &writeBufferEvent);
+
             m_hostStartTime = Profiler::GetTime();
-            m_deviceStartTime = GetDeviceTimestamp(context, device);
+
+            assert(err == CL_SUCCESS);
+            cl_int eventStatus;
+            err = clGetEventInfo(writeBufferEvent, CL_EVENT_COMMAND_EXECUTION_STATUS, sizeof(cl_int), &eventStatus, nullptr);
+            assert(err == CL_SUCCESS);
+            assert(eventStatus == CL_COMPLETE);
+            err = clGetEventProfilingInfo(writeBufferEvent, CL_PROFILING_COMMAND_END, sizeof(cl_ulong), &m_deviceStartTime, nullptr);
+            assert(err == CL_SUCCESS);
+            err = clReleaseEvent(writeBufferEvent);
+            assert(err == CL_SUCCESS);
+            err = clReleaseMemObject(dummyBuffer);
+            assert(err == CL_SUCCESS);
+            err = clReleaseCommandQueue(queue);
+            assert(err == CL_SUCCESS);
 
             auto item = Profiler::QueueSerial();
             MemWrite(&item->hdr.type, QueueType::GpuNewContext);
             MemWrite(&item->gpuNewContext.cpuTime, m_hostStartTime);
-            MemWrite(&item->gpuNewContext.gpuTime, m_hostStartTime);
+            MemWrite(&item->gpuNewContext.gpuTime, m_deviceStartTime);
             memset(&item->gpuNewContext.thread, 0, sizeof(item->gpuNewContext.thread));
             MemWrite(&item->gpuNewContext.period, 1.0f);
             MemWrite(&item->gpuNewContext.type, GpuContextType::OpenCL);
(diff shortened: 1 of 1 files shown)


wolfpld:
Pinging @mcleary

mcleary:
> This commit is closer to the Vulkan approach: all the code in the constructor, the memory operation is queued as non-blocking, and the host timestamp is read immediately after the `clWaitForEvents` call returns, which happens as soon as the event's status becomes `CL_COMPLETE` (similar to `vkQueueWaitIdle`). As this status change is the same time that is read with the call to `clGetEventProfilingInfo(writeBufferEvent, CL_PROFILING_COMMAND_END, ....`, I think that is the closest these timestamps can be read together.

That makes sense and it is probably a good idea.

> 
> With this change I get Execution Delays of around +15 μs for the OpenCLVectorAdd example's kernel launches on Linux/RTX 2080 Super, which I believe is a more typical delay for kernel launch.

I would like to test this code in other devices (amd, intel and arm) to see if this holds true. I can probably test it today and report back.

> 
> Also, the previous code was storing the `m_hostStartTime` into the context structure's `gpuTime` member, which is confusing, and required a `TimestampOffset` function to correct, even though I believe the profiler itself is doing the same offset internally if the `gpuTime` properly holds the GPU timestamp.

That was my mistake, thanks for pointing it out.

> 
> Further questions:
> 
> * Should the OpenCLCtx constructor take an optional `cl_command_queue` argument and avoid creating its own?  The current approach is probably fine, but the calling code has likely already created a command queue itself with the `CL_QUEUE_PROFILING_ENABLE` property, so why not allow passing it in? Avoids creating another queue, and seems the device timestamp should come from the same queue actually used in the application?

This was my first approach on writing the `OpenCLCtx`, the problem is that while it is likely that a queue was already created, for large applications, more than one queue (which one should the user pass to the OpenCLCtx?) can be used by the application and the queue might not necessarily have profiling capabilities turned on. While a queue with profiling turned on is required to collect timing information, such a queue shouldn't be a requirement when instantiating the OpenCLCtx, in my opinion.

johner:
>I would like to test this code in other devices (amd, intel and arm) to see if this holds true. I can probably test it today and report back.

Thank you, that will be interesting to see.  I have tested with the Intel CPU driver, and get positive Execution Delays, generally a bit longer than with the NVIDIA GPU driver.  I think that is to be expected.

>This was my first approach on writing the OpenCLCtx, the problem is that while it is likely that a queue was already created, for large applications, more than one queue (which one should the user pass to the OpenCLCtx?) can be used by the application and the queue might not necessarily have profiling capabilities turned on. While a queue with profiling turned on is required to collect timing information, such a queue shouldn't be a requirement when instantiating the OpenCLCtx, in my opinion.

I don't have a strong opinion on this issue, but either way I absolutely agree it should be optional - maybe a new `TracyCLContextQ` macro?  It's likely that any application creating an `OpenCLCtx` will _at some point_ have at least one queue with profiling enabled (else `Collect()` would assert or yield garbage under `NDEBUG`.).  But again, not a strong opinion.

mcleary:
@johner I managed to run your code on my Intel CPU/GPU and AMD GPU and I get very good results, similar to what you reported, arount 15μs. So I think we can move forward with your changes.

> Are the m_hostStartTime and m_deviceStartTime members still needed, or can we switch to local tcpu and tgpu variables as in the Vulkan code?

This won't be necessary anymore since you managed to pretty much remove the overhead of acquiring a device time sample. Would you be able to remove these members?

As a side note, I originally tried to use [clGetDeviceAndHostTimer](https://www.khronos.org/registry/OpenCL/sdk/2.1/docs/man/xhtml/clGetDeviceAndHostTimer.html) but the samples that I was getting from the device were completely bonkers when I tried. Maybe I can try this again in the future.

johner:
>@johner I managed to run your code on my Intel CPU/GPU and AMD GPU and I get very good results, similar to what you reported, arount 15μs. So I think we can move forward with your changes.

Good to hear, thank you.

>This won't be necessary anymore since you managed to pretty much remove the overhead of acquiring a device time sample. Would you be able to remove these members?

Yes, I've pushed a commit with that change.

>As a side note, I originally tried to use `clGetDeviceAndHostTimer` but the samples that I was getting from the device were completely bonkers when I tried. Maybe I can try this again in the future.

Well that function seems useful and relevant!  I wasn't even aware of it, as we tend to stick to the 1.2 API (MacOS and NVIDIA ya know).  Looks like it's [optional](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_device_and_host_timer_synchronization) in the long-awaited 3.0 drivers, so we could at least properly query for support whenever(?) those become available.

mcleary:
@wolfpld This is probably good to be merged now, after the checks of course.

wolfpld:
Ok, thanks.

PR #178, opened 2021-02-13
Add IsActive accessor to ScopedZone.

johner:
We have some expensive ZoneText we would like to generate only if the current zone is active.  Since we're using `TRACY_ON_DEMAND`, we would prefer querying the zone object if it's active rather than making another call to `GetProfiler().IsConnected()`.


wolfpld:
Please also add a short mention of this feature to the manual.

johner:
I have not yet added this, but should there be macros in Tracy.hpp for this accessor? e.g.
`#define ZoneIsActive() ___tracy_scoped_zone.IsActive();`
`#define ZoneIsActiveV( varname ) varname.IsActive();`

wolfpld:
This would probably be a good idea, but you're the one who's using this feature :)

johner:
I was looking for where to add info to the manual for this small feature and realized section "3.4 Marking zones" was probably best, but that section describes all macros, which made me think this should be available as macros as well.  I switched my local code to use macros instead of something like `__tracy_scoped_zone.IsActive()` and preferred it, so I think it's a reasonable addition.  I'll push that commit as well.

PR #181, opened 2021-02-23
D3D11  support

neortls007idev:
I was in urgent need of D3D11 support for my program so I decided to fork https://github.com/wolfpld/tracy/pull/88 and made some minor changes like fixing the D3D11 Tracy macros for ZonedScope, Creation and Destruction of the Context and renaming the classes for consistency with D3D12 implementation.


common/TracyQueue.hpp
```diff
@@ -302,7 +302,8 @@ enum class GpuContextType : uint8_t
     OpenGl,
     Vulkan,
     OpenCL,
-    Direct3D12
+    Direct3D12,
+    Direct3D11
```
wolfpld, line 306: The `GpuContextNames` table in `TracyView.cpp` is not updated.

Tracy.hpp
```diff
@@ -103,6 +103,10 @@
 #include "client/TracyProfiler.hpp"
 #include "client/TracyScoped.hpp"
 
+#define TracySourceLocation( varname ) static const tracy::SourceLocationData varname { nullptr, __FUNCTION__,  __FILE__, (uint32_t)__LINE__, 0 };
+#define TracySourceLocationN( varname, name ) static const tracy::SourceLocationData varname { name, __FUNCTION__,  __FILE__, (uint32_t)__LINE__, 0 };
+#define TracySourceLocationNC( varname, name, color ) static const tracy::SourceLocationData varname { name, __FUNCTION__,  __FILE__, (uint32_t)__LINE__, color };
+
```
wolfpld, line 109: This should be removed.

CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+endif()
+
+message(STATUS "TRACY_PORT: ${TRACY_PORT}")
+target_compile_definitions(tracy_client PUBLIC TRACY_PORT=${TRACY_PORT})
+
+target_include_directories(tracy_client PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
+
+if (UNIX)
+    target_link_libraries(tracy_client PRIVATE ${CMAKE_DL_LIBS})
+endif()
```
wolfpld, line 40: This should be removed.

wolfpld (changes requested):
The manual has to be updated to explain how all this should be used.

neortls007idev:
Thanks will fix them all in 48 hours.

CMakeLists.txt
wolfpld: (By "this" I mean the entire file.)

neortls007idev:
@wolfpld not sure what the capstone.h file is for and I think it comes from vcpkg as far as I could tell after testing for build on my local machine. Can you shed some light on why the build fails there?

wolfpld:
Please ignore the failed MSVC build. There were incompatible changes in vcpkg _again_.

neortls007idev:
There are certain things I haven't had the time to test like tracking multiple deferred contexts or even immediate context migration over multiple threads. So, Should I mention that as a note or something ?  Also the context Owner thread is never shown but in my case I know the my main thread owns it so I am not sure if this is something that should be mentioned in the manual documentation. 
Tracy

wolfpld:
All the shortcomings should be described in the manual. As for the lack of displayed thread, you need to make sure proper thread identifiers are set in the GPU context creation message and all subsequent GPU zone messages, and that it is handled properly server-side. There are two modes of operation for GPU context, one is single-threaded (used for example in OpenGL), and the second is multithreaded (used for example in Vulkan). Maybe you are using the `TRACY_D3D11_NO_SINGLE_THREAD` define incorrectly (I don't know what it is for)?

https://github.com/wolfpld/tracy/blob/c12de1b/server/TracyWorker.cpp#L5303-L5315
https://github.com/wolfpld/tracy/blob/c12de1b/server/TracyView.cpp#L2922

neortls007idev:
I might need to dig deeper on how it works first, since my goal at the moment was just making it work and being able to do basic profiling and fixing whatever issues the pull request had in it to make it work.

PR #186, opened 2021-03-16
replace __FUNCTION__ with __func__ from c99 std

annacrombie:
https://gcc.gnu.org/onlinedocs/gcc/Function-Names.html

`__FUNCTION__` gives warnings with `-Wpedantic` as it is not part of the c standard.

Also seems to be well supported on windows: 
https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160


wolfpld:
There are subtle differences how these macros work across different compilers.

```c++
struct A
{
    static void B( int c )
    {
        printf( "__FUNCTION__ %s", __FUNCTION__ );
        printf( "__func__ %s", __func__ );
    }
};
```

While on gcc/clang both `__FUNCTION__` and `__func__` would print `B`, on MSVC `__FUNCTION__` prints `A::B` and `__func__` prints just `B`.

Moreover, I don't seem to be able to trigger this warning?

```
[11:46 wolf@oberon:~]% cat a.cpp
#include <stdio.h>

int main()
{
    printf( "%s\n", __FUNCTION__ );
}
[11:46 wolf@oberon:~]% g++ a.cpp -pedantic -Wall -Wpedantic
[11:46 wolf@oberon:~]% g++ --version
g++ (GCC) 10.2.0
```

annacrombie:
> subtle differences how these macros work

That's a shame.  In C-only programs however, the problem above wouldn't exist since there are no namespaced functions in C.

> I don't seem to be able to trigger this warning.

Maybe try gcc?

```
$ cat test.c
#include <stdio.h>

int
main(void)
{
	printf("%s\n", __func__);
	printf("%s\n", __FUNCTION__);
}
$ gcc --version
gcc (Alpine 10.2.1_pre2) 10.2.1 20210313
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ gcc -Wpedantic ./test.c
./test.c: In function 'main':
./test.c:7:17: warning: ISO C does not support '__FUNCTION__' predefined identifier [-Wpedantic]
    7 |  printf("%s\n", __FUNCTION__);
      |                 ^~~~~~~~~~~~
```

wolfpld:
Ah yes, I didn't notice this change was only for the C API header.

PR #187, opened 2021-03-16
update gl3w

annacrombie:
Tracy recently stopped building on my system, complaining of a missing `glXGetProc` during the final linking step.  I haven't looked into exactly how my system changed to break the build, and I'm aware that this is a rather heavy-handed solution, so feel free to disregard it.  

On the other hand, there have been a lot of changes to gl3w since 2017, and I thought you may be interested.

The pr where I believe my specific issue was fixed in gl3w is [here](https://github.com/skaslev/gl3w/pull/30).


wolfpld:
I think the breakage you are observing might stem from #188, where you have removed linking with `libGL` if `TRACY_NO_FILESELECTOR` is defined.

wolfpld:
With this update linking with OpenGL can be removed, as the library is dynamically loaded.

annacrombie:
> I think the breakage you are observing might stem from #188, where you have removed linking with `libGL` if `TRACY_NO_FILESELECTOR` is defined.

Duh!  I can't believe I overlooked this yesterday...

PR #188, opened 2021-03-16
fix TRACY_NO_FILESELECTOR macro

annacrombie:
I tried to building without GTK, and discovered `TRACY_NO_FILESELECTOR` which helped a little, but was incomplete.  This PR adds a few more guards around usage of the file selector, and adds a build variable to control TRACY_NO_FILESELECTOR.  Additionally, linking of GTK is disabled when you pass this.  

I haven't updated the docs yet since I didn't want to do the work unless you seemed receptive to this feature.


wolfpld:
The reason why this was incomplete (mainly in `main.cpp`) was to enable embedding server into the profiled application, where you don't need the initial dialog, so you'd just directly use `View`.

Looks good, please update the docs.

PR #194, opened 2021-04-09
Fix centos shutdown crash

jkriegshauser:
This addresses an issue (mostly present in centos7) where if the tracy client source is compiled into a shared object with a version of GCC prior to 8.4, a crash on shutdown can occur as glibc tries to destruct the `ProfilerThreadData` on the main thread after the library has unloaded.

### DIFF
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -1041,11 +1041,58 @@ static ProfilerData& GetProfilerData()
 }
 #  endif
 
+// GCC prior to 8.4 had a bug with function-inline thread_local variables. Versions of glibc beginning with
+// 2.18 may attempt to work around this issue, which manifests as a crash while running static destructors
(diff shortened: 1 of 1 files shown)


wolfpld:
I did some work to avoid using system allocator during initialization in 40efbe852 (#196), and it may be good to do something similar here, as you are using `new` and `delete`.

jwdevel:
Indeed, I notice that even after jumping through some hoops described at #196, while I can then use `TracyAlloc`+`TracyFree` inside my `malloc` wrappers, I *cannot* use `ZoneScoped` et. al. because of a circular initialization problem coming from this code (my `ZoneScoped` does init, which calls `new` in this codepath (I am using GCC 8.3, so I hit it), and it calls `malloc` — infinite loop; stack overflow).

I may try switching this to use rpmalloc to see what happens ...

PR #195, opened 2021-04-09
Windows getenv fix

jkriegshauser:
From the code comment:

> unfortunately getenv() on Windows is just fundamentally broken.  It caches the entire environment block once on startup, then never refreshes it again.  If any environment strings are added or modified after startup of the CRT, those changes will not be seen by getenv().  This removes the possibility of an app using this SDK from programmatically setting any of the behaviour controlling envvars here.

### DIFF
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -1261,7 +1261,6 @@ Profiler::Profiler()
     CalibrateDelay();
     ReportTopology();
 
-    char buffer[32];
 #ifndef TRACY_NO_EXIT
     const char* noExitEnv = GetEnvVar( "TRACY_NO_EXIT" );
     if( noExitEnv && noExitEnv[0] == '1' )
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -175,6 +175,7 @@ struct ThreadHandleWrapper
 };
 #endif
 
+
 #if defined __i386 || defined _M_IX86 || defined __x86_64__ || defined _M_X64
 static inline void CpuId( uint32_t* regs, uint32_t leaf )
 {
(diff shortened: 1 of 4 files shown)


wolfpld:
What would be the use case for this?

There are two `getenv()` calls you have missed:
```
common/TracySocket.cpp:    const char* onlyLocalhost = getenv( "TRACY_ONLY_LOCALHOST" );
common/TracySocket.cpp:    const char* onlyIPv4 = getenv( "TRACY_ONLY_IPV4" );
```

client/TracyProfiler.cpp
```diff
@@ -234,7 +270,8 @@ static int64_t SetupHwTimer()
     CpuId( regs, 0x80000007 );
     if( !( regs[3] & ( 1 << 8 ) ) )
     {
-        const char* noCheck = getenv( "TRACY_NO_INVARIANT_CHECK" );
+        char buffer[32];
+        const char* noCheck = GetEnvVar( "TRACY_NO_INVARIANT_CHECK", buffer, sizeof(buffer) / sizeof(buffer[0]) );
```
wolfpld, line 274: I'd prefer to keep the `getenv()` interface, so that this silly buffer business can be omitted. You can just use static char buffer in `GetEnvVar()`, if needed.

jkriegshauser:
From Eric, who doesn't have a github account:
> The two locations in ‘TracySocket.cpp’ were omitted out of laziness.  I didn’t take the time to find a suitable common location to put the helper function so it could be shared between the two files.  I can move it to a shared location and get those other two locations as well. 
> 
> The parameter buffer was used to ensure there wouldn’t be thread safety issues.  Since these calls only seem to be used during initialization times, it’s likely safe to use a static buffer.
> 
> The specific use case here is to programmatically avoid the `exit(0)` call when the invariant TSC check fails.  Tracy is being loaded as part of a plugin in our case so an initialization failure in Tracy causing the entire process to exit prematurely was less than desirable.  Unfortunately because of Windows broken getenv() implementation, setting that envvar programmatically tp override the exit behaviour was impossible.

Will look at addressing both the TracySocket.cpp and the buffer issue.

jkriegshauser:
@wolfpld should be updated now.

PR #208, opened 2021-05-06
add ability to read .zst files in the chrome importer

c-cube:
This is useful because json traces can get big very quickly, so they're often compressed. Supporting `.gz` would also be useful.

(warning: I'm not a C++ programmer)


wolfpld:
Added a more efficient implementation in fa1a717e.

c-cube:
Thank you!

PR #209, opened 2021-05-06
fix: need to use zstd's streaming API

c-cube:
follow up to #208.

Sometimes the size isn't known, if the .zst file was produced via
streaming, so we also need to parse it using streaming. The code is not pretty, sorry.

### DIFF
--- a/import-chrome/src/import-chrome.cpp
+++ b/import-chrome/src/import-chrome.cpp
@@ -81,19 +81,52 @@ int main( int argc, char** argv )
             exit( 1 );
         }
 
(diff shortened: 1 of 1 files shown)


wolfpld:
Please attach an example of such file.

c-cube:
(it's in a zip because github will not let me upload a .zst file)

[trace.json.zst.zip](https://github.com/wolfpld/tracy/files/6436967/trace.json.zst.zip)

wolfpld:
Applied as 7f0191d1.

PR #211, opened 2021-05-10
chrome-import: handle multiple pids

### LINKED ISSUE #210: support multiple pid in chrome-import
The catapult file format supports traces with multiple processes: messages can have both `tid` and `pid`. Would it be possible to encode the pair `tid, pid` into a single `uint64_t` when importing such traces?

c-cube:
close #210 

also I improved a bit the way `args` is handled.

### DIFF
--- a/import-chrome/src/import-chrome.cpp
+++ b/import-chrome/src/import-chrome.cpp
@@ -133,10 +133,29 @@ int main( int argc, char** argv )
     printf( "\33[2KParsing...\r" );
     fflush( stdout );
 
+    struct PidTid {
+      uint64_t tid;
+      uint64_t pid;
+      uint64_t both; // fake thread id, unique within Tracy
+    };
+
+    std::vector<PidTid> tids;
     std::vector<tracy::Worker::ImportEventTimeline> timeline;
     std::vector<tracy::Worker::ImportEventMessages> messages;
     std::vector<tracy::Worker::ImportEventPlots> plots;
     std::unordered_map<uint64_t, std::string> threadNames;
+
+    const auto getTid = [&](uint64_t pid, uint64_t tid) -> uint64_t {
+        for ( auto &pair : tids ) {
+            if ( pair.pid == pid && pair.tid == tid ) {
+                return pair.both;
+            }
+        }
+
+        const auto result = tids.size();
+        tids.emplace_back(PidTid {.tid=tid, .pid=pid, .both=result});
+        return result;
+    };
 
     if( j.is_object() && j.contains( "traceEvents" ) )
     {
@@ -162,10 +181,15 @@ int main( int argc, char** argv )
             }
         }
 
+        uint64_t pid = 0;
+        if ( v.contains( "pid" ) ) {
+            pid = v["pid"].get<uint64_t>();
+        }
+
         if( type == "B" )
         {
             timeline.emplace_back( tracy::Worker::ImportEventTimeline {
-                v["tid"].get<uint64_t>(),
+                getTid(pid, v["tid"].get<uint64_t>()),
                 uint64_t( v["ts"].get<double>() * 1000. ),
                 v["name"].get<std::string>(),
                 std::move(zoneText),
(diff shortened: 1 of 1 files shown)


import-chrome/src/import-chrome.cpp
```diff
@@ -133,11 +133,30 @@ int main( int argc, char** argv )
...
 
+    const auto getTid = [&](uint64_t pid, uint64_t tid) -> uint64_t {
+        for ( auto &pair : tids ) {
+            if ( pair.pid == pid && pair.tid == tid ) {
+                return pair.both;
+            }
+        }
+
+        const auto result = tids.size();
+        tids.emplace_back(PidTid {.tid=tid, .pid=pid, .both=result});
```
wolfpld, line 156: This is a C++20 feature.

import-chrome/src/import-chrome.cpp
```diff
@@ -158,14 +177,23 @@ int main( int argc, char** argv )
...
+                    val.is_string() ?
+                    val.get<std::string>() : val.dump();
+                zoneText += kv.key() + ": " + s + "\n";
             }
         }
 
+        uint64_t pid = 0;
+        if ( v.contains( "pid" ) ) {
+            pid = v["pid"].get<uint64_t>();
+        }
```
wolfpld, line 191: Why is PID extracted here, but TIDs are inside the `if` bodies?

import-chrome/src/import-chrome.cpp
```diff
@@ -158,14 +177,23 @@ int main( int argc, char** argv )
         {
             for ( auto& kv : v["args"].items() )
             {
-                zoneText += kv.key() + ": " + kv.value().dump() + "\n";
+                const auto val = kv.value();
+                const std::string s =
+                    val.is_string() ?
+                    val.get<std::string>() : val.dump();
+                zoneText += kv.key() + ": " + s + "\n";
```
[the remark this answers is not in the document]
c-cube: In case you have a rich string in the argument (say with newlines inside), this decodes the json string to turn "\\n" into a regular '\n' that will be rendered on multiple lines in Tracy.

import-chrome/src/import-chrome.cpp
c-cube: oh, my bad. does `push` work?

import-chrome/src/import-chrome.cpp
c-cube: good question, I mostly followed the existing structure of the code. I can fix that.

import-chrome/src/import-chrome.cpp
```diff
@@ -133,11 +133,30 @@ int main( int argc, char** argv )
...
+      uint64_t both; // fake thread id, unique within Tracy
+    };
+
+    std::vector<PidTid> tids;
     std::vector<tracy::Worker::ImportEventTimeline> timeline;
     std::vector<tracy::Worker::ImportEventMessages> messages;
     std::vector<tracy::Worker::ImportEventPlots> plots;
     std::unordered_map<uint64_t, std::string> threadNames;
 
+    const auto getTid = [&](uint64_t pid, uint64_t tid) -> uint64_t {
```
wolfpld, line 148: This has to be better named. There are two different meanings for `tid`. At the same time, `tid` and `both` are used as aliases.

import-chrome/src/import-chrome.cpp
wolfpld: You should do `tid.emplace_back( PidTid { tid, pid, result } );`, in the struct order. Designated initializers are not in the standard.

import-chrome/src/import-chrome.cpp
wolfpld: It certainly looks like `getTid()` could have just one json parametr and extract tid and pid itself.

import-chrome/src/import-chrome.cpp
```diff
@@ -133,11 +133,39 @@ int main( int argc, char** argv )
...
+
+        const auto real_tid = val["tid"].get<uint64_t>();
+
+        for ( auto &pair : tid_encoders) {
+            if ( pair.pid == pid && pair.tid == real_tid ) {
+                return pair.pseudo_tid;
+            }
+        }
+
+        const auto pseudo_tid = tid_encoders.size();
```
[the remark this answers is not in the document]
c-cube: that's fair. I guess we can keep the original tid if there's no pid.

c-cube:
Thank you!

PR #217, opened 2021-05-14
 I encountered some bugs when trying to use tracy in UE, the above is my fix

King19931229:
Most of the bug I encounter is rpmalloc initializing problem in android, especially when I use delay init. Other are compile problem in IOS
Also, I add support on python profile. The code is quite straightforward


wolfpld:
There are things here that can be merged right away, while other need discussion. Please split your changes appropriately (for example: warning fixes, python profiling, rpmalloc fixes) and submit as separate PRs.

PR #221, opened 2021-05-20
The value for paramter 'compressed_ttf_size' passed to 'AddFontFromMemoryCompressedTTF' seems to be a typo.

RichardUSTC:
The value for paramter 'compressed_ttf_size' passed to 'AddFontFromMemoryCompressedTTF' seems to be a typo.


wolfpld:
It's interesting that it seems to be working as intended.

PR #222, opened 2021-05-20
build with gcc 11

xxxbxxx:
When compiling with gcc 11 I get:

```
In file included from ../../../server/TracySourceContents.cpp:1:
../../../server/TracySourceContents.hpp:36:5: error: 'size_t' does not name a type
   36 |     size_t m_dataSize;
      |     ^~~~~~
../../../server/TracySourceContents.hpp:8:1: note: 'size_t' is defined in header '<cstddef>'; did you forget to '#include <cstddef>'?
    7 | #include "TracySourceTokenizer.hpp"
  +++ |+#include <cstddef>
    8 | 
```

and
```
In file included from ../../../server/TracySourceTokenizer.cpp:1:
../../../server/tracy_robin_hood.h: In member function 'size_t tracy::detail::Table<IsFlat, MaxLoadFactor100, Key, T, Hash, KeyEqual>::calcMaxNumElementsAllowed(size_t) const':
../../../server/tracy_robin_hood.h:2122:52: error: 'numeric_limits' is not a member of 'std'
 2122 |         if (ROBIN_HOOD_LIKELY(maxElements <= (std::numeric_limits<size_t>::max)() / 100)) {
      |                                                    ^~~~~~~~~~~~~~
```

### DIFF
--- a/server/TracySourceContents.hpp
+++ b/server/TracySourceContents.hpp
@@ -2,6 +2,7 @@
 #define __TRACYSOURCECONTENTS_HPP__
 
 #include <stdint.h>
+#include <stddef.h>
 #include <vector>
 
 #include "TracySourceTokenizer.hpp"
--- a/server/tracy_robin_hood.h
+++ b/server/tracy_robin_hood.h
@@ -47,6 +47,7 @@
 #include <string>
 #include <type_traits>
 #include <utility>
+#include <limits>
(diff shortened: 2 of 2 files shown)


0xBYTESHIFT:
Can this be added to the v0.7.8 tag? As for now, the profiler is broken because of this in the mentioned tag.

PR #223, opened 2021-05-21
add TRACY_CALLSTACK_IGNORE_INLINES

xxxbxxx:
In some cases, during a profiling session Tracy appears to be 'stuck' with a large backlog of requests for a very long time that doesn't seem to ever reduce. 

I think it's from two causes:
 * the client tries to process all the queries it received before sending the results back -> the profiler is blocked while the results accumulate on the client.

 * this situation occurs because the Win32 dbghelp SymAddrIncludeInlineTrace() call is very slow, several orders of magnitude slower than the simple ones.  (the client goes from overall answering 20000 queries per second to 70)

I don't know if it's the correct approach, but here are a pair of changes to flush once in a while to accumulated query results and add a define to disable the inline resolution code.

### DIFF
--- a/client/TracyCallstack.cpp
+++ b/client/TracyCallstack.cpp
@@ -81,6 +81,7 @@ extern "C"
     t_RtlWalkFrameChain RtlWalkFrameChain = 0;
 }
 
+#ifndef TRACY_CALLSTACK_IGNORE_INLINES
 #if defined __MINGW32__ && API_VERSION_NUMBER < 12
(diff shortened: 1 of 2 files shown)


wolfpld:
Added the macro (with a different name) as d7541bbdba, 621d1b03cd.

I'll take a look at the query response limiter some time later.

wolfpld:
Added limiter in 1e6aedf9.

PR #224, opened 2021-05-21
Use file:line when comparing traces

keur:
When comparing traces, where multiple classes share the same zone
names, the behavior prior to this patch was to auto-select the first
matching zone name in the other trace. Instead, find the most correct
zone by using filename and line number.


wolfpld:
Maybe a graded solution would be better, where you first compare function name + file + line, then if no exact match is found you'd only compare function name + file, and when there's still no match, you'd fall back to the old behavior?

keur:
i agree that's a good idea. is this what you had in mind cc @wolfpld

Issue #159, opened 2021-01-05
Re-entering zones through recursion inflates total time in zone statistics


rokopt:
If a `ZoneScoped` surrounds a recursive call, the time spent in child zones of the recursive call appears to be accounted to the parent once for each occurrence of the parent on the stack.  For example, if I run the following (Catch-2) test:

```
static void
child()
{
    ZoneScoped;
    sleep(1);
}

static void
parent(size_t numRecursions)
{
    ZoneScoped;
    if (numRecursions == 0)
    {
        child();
    }
    else
    {
        parent(numRecursions - 1);
    }
}

TEST_CASE("Tracy recursive zones")
{
    sleep(5); // allow time for Tracy handshake
    parent(10);
}
```

When I bring up "Statistics", the "self time" shows what I'd expect -- 1s in `child()` and practically nothing in `parent()` -- but when I un-check "self time" to show the total time, `parent()` shows 11s.

wolfpld:
This is how inclusive time counting works, which you explicitly requested by disabling the "self time" option.

rokopt:
Thanks for the reply.  I'll close the issue.

rokopt:
Is there some other statistic I could examine, or zone annotation that I could use, which would show ~1s for `parent()` in this case?  I'd like to know what latencies clients are getting when they call a function such as `parent()`, whose implementation might make some (possibly indirect) recursive calls which would induce the inclusive zone time time to reflect more than the wall clock time spent in the call from the client (and whose "self time" does not reflect the latency either because much of the time is spent in child zones).

wolfpld:
You can inspect the "child zones" list or "time distribution" for the parent zone in the zone information window.

rokopt:
> You can inspect the "child zones" list or "time distribution" for the parent zone in the zone information window.

Thanks.  The "child zones" list does show 1s in `parent`.  I think that view is only for single invocations, though -- there's not a summary form of that one, is there?

And (sorry to keep at this) the total time displayed in the "time distribution" for `parent` is also 11s -- is it not a bug for a single-zone time accumulator to show a time greater than the wall-clock time spent by a single-threaded program?  (It occurs to me that if I'd called `sleep()` directly from `parent()` in my example above instead of calling `child()`, then the `parent` zone would have a self time of 1s and an inclusive time of 11s.  If there's only one zone in the entire system, shouldn't the self and inclusive time for that one zone be equal?)

[...]

PR #230, opened 2021-06-05
Introduce "non-reentrant time" statistics view

rokopt:
This PR implements the enhancement proposed in #159 :  introducing to the Statistics window a radio button, "Non-reentrant time", which counts only the first appearance of any one zone on a thread's call stack.

Some notes:

- Per https://github.com/wolfpld/tracy/issues/159#issuecomment-767819118 , I haven't added this option to "Find Zone", only to "Statistics".
- Similarly, I haven't added this option to CSV export.
- For an explanation of the reason for and workings of the new `View::IsZoneReentry` methods, see https://github.com/wolfpld/tracy/issues/159#issuecomment-774457623 .


wolfpld:
Please fix the merge conflicts.

rokopt:
> Please fix the merge conflicts.

Fixed.

wolfpld:
I don't like the current UI for time selection. Please change to combo box, e.g.:

obraz

rokopt:
> Please change to combo box, e.g.:

I've pushed a change to do that; how does it look?

(It's verbose translating back and forth between the `int` that the combo box needs and the `enum class AccumulationMode`; I'm not sure whether you find it worth having the `enum class` anyway -- I could change that if you prefer.)

wolfpld:
> (It's verbose translating back and forth between the int that the combo box needs and the enum class AccumulationMode; I'm not sure whether you find it worth having the enum class anyway -- I could change that if you prefer.)

You can replace these switches with a pair of explicit casts.

wolfpld:
The "non-reentrant" option should not be available when displaying sampling statistics, as it is not implemented there.

> I've pushed a change to do that; how does it look?

Much better. I would additionally change:
- "Including children" to "With children".
- "Accumulation mode" to "Accumulation", "Timing", "Stats", or something like that (the shorter, the better).
- Add some additional spacing, so that the options are not clumped together:

```c++
ImGui::SameLine();
ImGui::Spacing();
ImGui::SameLine();
```

Probably would be good to add this also to the rest of checkboxes.

rokopt:
> You can replace these switches with a pair of explicit casts.

Okay, done.

rokopt:
> The "non-reentrant" option should not be available when displaying sampling statistics, as it is not implemented there.

Good catch -- I've suppressed it in that case.


> Much better. I would additionally change:
> 
> * "Including children" to "With children".
> * "Accumulation mode" to "Accumulation", "Timing", "Stats", or something like that (the shorter, the better).

Done.

> * Add some additional spacing, so that the options are not clumped together:
> 
> ```c++
> ImGui::SameLine();
> ImGui::Spacing();
> ImGui::SameLine();
> ```

Done.

> Probably would be good to add this also to the rest of checkboxes.

Okay, I've sprinkled spacing throughout some of the pre-existing options too.  I did that in a separate commit for easy reversion in case I didn't do it where you intended.  :)

server/TracyView.cpp
```diff
@@ -12269,9 +12269,29 @@ struct SrcLocZonesSlim
...
+    if ( m_statMode != 0 && m_statAccumulationMode == AccumulationMode::NonReentrantChildren )
+    {
+        m_statAccumulationMode = AccumulationMode::SelfOnly;
+    }
+    int accumulationMode = static_cast<int>( m_statAccumulationMode );
+    ImGui::Combo( "##accumulationMode", &accumulationMode, accumulationModeTable );
+    m_statAccumulationMode = static_cast<AccumulationMode>( accumulationMode );
+    ImGui::SameLine();
+    ImGui::Spacing();
+    ImGui::SameLine();
```
wolfpld, line 12292: I think it is invalid to force `SameLine` and `Spacing` here (and in the function prologue). It should be responsibility of the code calling this function to properly glue the item to its surroundings. Right now it seems that a `Separator` call was catched by `SameLine`:

obraz

rokopt: > I think it is invalid to force `SameLine` and `Spacing` here (and in the function prologue). It should be responsibility of the code calling this function to properly glue the item to its surroundings. Right now it seems that a `Separator` call was catched by `SameLine`:

Good eye.  How does the latest look?

rokopt:
Thank you for all your time and guidance and patience with this one!

PR #242, opened 2021-07-16
Fixing unbalanced tree pop in grouped GPU zones.

benvanik:


wolfpld:
Good catch. I think it would be better to move this outside the function.

benvanik:
Good idea - done!

PR #243, opened 2021-07-17
Indenting empty GPU zone tree nodes when grouped to keep zones aligned.

benvanik:
Before:
image

After:
image

image


wolfpld:
Can you do this also for CPU zones?
"Self time" should be indented too.

benvanik:
Done!

PR #257, opened 2021-08-29
Filter samples by zones source locations

xxxbxxx:
Give a view of the samples by "zone" to get further insight on the time spent in the instrumented zones without additional instrumentation.
(using all the samples from identical "source locations" of the whole capture)

not sure if it's done properly (I'm not very familiar with the code base, nor with the stl...)  but it seems to work for me.
~~it is limited to exclusive samples, couldn't get the inclusive to work.~~

note: It is also possible to implement along with other filtering options, but it is slow (need to examine all individual samples in the range and find the matching zone) and probably not that useful.

### DIFF
--- a/server/TracyEvent.hpp
+++ b/server/TracyEvent.hpp
@@ -268,6 +268,7 @@ enum { SampleDataSize = sizeof( SampleData ) };
 struct SampleDataRange
 {
     Int48 time;
+    uint16_t thread;
     CallstackFrameId ip;
 };
 
--- a/server/TracyView.cpp
+++ b/server/TracyView.cpp
@@ -10732,109 +10732,6 @@ void View::DrawFindZone()
                 m_findZone.scheduleResetMatch = true;
             }
             
-            if( m_findZone.samples.enabled && m_findZone.samples.scheduleUpdate && !m_findZone.scheduleResetMatch )
-            {
-                m_findZone.samples.scheduleUpdate = false;
-
-                const auto& symMap = m_worker.GetSymbolMap();
-                m_findZone.samples.counts.clear();
-                m_findZone.samples.counts.reserve( symMap.size() );
-
-                struct GroupRange {
-                    const FindZone::Group* group;
-                    Vector<short_ptr<ZoneEvent>>::const_iterator begin;
-                    Vector<short_ptr<ZoneEvent>>::const_iterator end;
-                };
-                Vector<GroupRange> selectedGroups;
-                selectedGroups.reserve( m_findZone.groups.size() );
-                for( auto it = m_findZone.groups.begin(); it != m_findZone.groups.end(); ++it )
-                {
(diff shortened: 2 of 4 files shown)


wolfpld:
You are building data structures when trace is loaded, but there is no machinery to create this data during a live capture. This will have to be added.

Since the UI has changed, please provide screenshots with the new options.

I'm not entirely sure how these changes can be used. Please think how to explain this in more detail, as an entry in the manual will be needed before this can be merged.

xxxbxxx:
> You are building data structures when trace is loaded, but there is no machinery to create this data during a live capture. This will have to be added.

mmm. indeed, will look into it.

> I'm not entirely sure how these changes can be used. Please think how to explain this in more detail, as an entry in the manual will be needed before this can be merged.

maybe not the best example, but
The idea is that when you can go from all the samples for the whole session:
samples_no_filter

to having some idea of what is going on in, say, the "viewport" zone. (lot's of dma to upload pixels to the gpu apparently...)
samples_by_zone1

It's not the same as just applying a range limit, since you'd get:  very few samples for the one zone (and samples from all unrelated threads ).

it is maybe redundant to filtering the samples child of some function if the zone actually matches a specific call, but it may be not so simple if the code path is shared with other moments of the frame or there are multiple zones in one function .

Do you think it is useful? or maybe it is redundant with some other feature or makes no sense?

wolfpld:
Are you limiting the scope to the single "viewport" zone you have highlighted? Or is this for all zones with this source location?

xxxbxxx:
all zones with this source location. 
(otherwise there would be very few samples - that you can already see by zooming and clicking on the dots.)

wolfpld:
I think in this case it may be more appropriate to move this functionality to the "find zone" window, and support all the narrowing down you can do there. There already is functionality to cache the results of such calculations, and it may be extended to also include finding samples.

xxxbxxx:
> I think in this case it may be more appropriate to move this functionality to the "find zone" window,

It indeed makes much more sense.

image

wolfpld:
I have taken a look at this and I have two additional functional requests:
1. Samples should be narrowed down to the "Found zones" selection, i.e. I want to see data only for the blue portion of the histogram.
obraz
2. Sample data is not displayed, when the histogram time-range selection is being made. Data is only updated when the selection is finished. Instead, the information should be presented live, as the time range is selected.
obraz

xxxbxxx:
>     2\. Sample data is not displayed, when the histogram time-range selection is being made. Data is only updated when the selection is finished. Instead, the information should be presented live, as the time range is selected.

I did this because it's a pretty slow with many zones and samples. 
But yeah, it still is usable and actually usefull, and you can hide the samples section if it's too slow.

xxxbxxx:
> I see that you have made additional commits. Is this ready for review?

I rebased on master. I don't have any other changes in mind.

PR #259, opened 2021-09-08
Add an option to switch between the default __FUNCTION__ and more verbose __PRETTY_FUNCTION__

Listwon:
[`__PRETTY_FUNCTION__`](https://gcc.gnu.org/onlinedocs/gcc/Function-Names.html) gives more verbose output with namespace context.

tracy_pretty_function
tracy_pretty_function2

### DIFF
--- a/Tracy.hpp
+++ b/Tracy.hpp
@@ -105,22 +105,28 @@
 #include "client/TracyProfiler.hpp"
 #include "client/TracyScoped.hpp"
 
-#if defined TRACY_HAS_CALLSTACK && defined TRACY_CALLSTACK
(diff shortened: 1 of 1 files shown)


wolfpld:
`namespace::method` is the intended result. It cannot be achieved on gcc/clang, as you either get just the function name, or the full decorated function signature.

This PR does misses discussion on which compilers are affected by the proposed change.

The changes are incomplete, these are not all places where `__FUNCTION__` is used.

There is no documentation for this change.

wolfpld:
This has been implemented in #497.

PR #265, opened 2021-09-18
Expose TRACY_MANUAL_LIFETIME APIs to C API clients 

nagisa:
These are extremely useful for ecosystems such as Rust. There are a
couple of reasons why:

1. Rust strongly advises against relying on life before/after main, as
   it is difficult to reason about. Most users working in Rust will
   generally be quite surprised when encountering this concept.
2. Rust and its package manager makes it easy to use packages (crates)
   and somewhat less straightforward to consider the implications of
   including a dependency.

   In case of the `rust_tracy_client` set of packages, I currently have
   to warn throughout the documentation of the package that simply
   adding a dependency on the bindings package is sufficient to
   potentially accidentally broadcast a lot of information about the
   instrumented binary to the broader world. This seems like a major
   footgun given how easy is it to forget about having added this
   dependency.

Ability to manually manage the lifetime of the profiler would be a great
solution to these problems.

---

Some things to think about:

1. We could approach the implementation by doing something along the lines of

```
#define StartupProfiler() ___tracy_startup_profiler()
#define ShutdownProfiler() ___tracy_shutdown_profiler()
```

or some such to avoid a wrapper function.

2. Shed colour: `___tracy_startup`, `___tracy_shutdown` sound as descriptive as the variants with `profiler` in it.

### DIFF
--- a/TracyC.h
+++ b/TracyC.h
@@ -101,6 +101,12 @@ struct ___tracy_c_zone_context
 // This struct, as visible to user, is immutable, so treat it as if const was declared here.
 typedef /*const*/ struct ___tracy_c_zone_context TracyCZoneCtx;
 
+
+#ifdef TRACY_MANUAL_LIFETIME
+TRACY_API void ___tracy_startup_profiler(void);
+TRACY_API void ___tracy_shutdown_profiler(void);
+#endif
+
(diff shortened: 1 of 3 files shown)


wolfpld:
In future please refrain from refering to what Rust advises, expects, or believes in. I have tried it once and it had a Strong Opinion that the variable names I elected to use in my own code were Wrong, and that I should immediately change them due to Important Reasons.

A simple "expose already existing functionality through the C API" is enough here.

PR #276, opened 2021-11-04
Removing explicit compiler define check for GL_TIMESTAMP

alexanderbock:
Some OpenGL binding frameworks (glbindings (https://github.com/cginternals/glbinding) in my example) are not defining the OpenGL constants using preprocessor defines but using enum values instead.  So even when I include the file before, this check will fail as it is not a preprocessor define.  Just defining it by hand before the TracyOpenGL.h include does not work either as it leads to further compile errors down the line because of name collisions

### DIFF
--- a/TracyOpenGL.hpp
+++ b/TracyOpenGL.hpp
@@ -31,10 +31,6 @@ public:
 
 #else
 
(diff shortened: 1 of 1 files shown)


wolfpld:
When you hate preprocessor so much, that you will resort to non-standard language features instead of include guards (which could be used to detect inclusion here...)

Issue #180, opened 2021-02-22
C API for graphics profiling


jesta88:
For C projects, it would be very useful to have a C API for Vulkan and OpenGL. I'm unsure how feasible this is for DirectX 12 though.

Issue #106, opened 2020-09-03
Fiber support


jkriegshauser:
Our task system uses Fibers. Since Tracy appears to get its own thread information this makes profiling with fibers problematic. Ideally we like to show a fiber as a thread. In order for this to work, we would need a way to override the current Thread ID that Tracy uses.

A few ideas:
- Additional `TracyCZone` functions that take a "context identifier" that could either be a thread ID or a fiber ID
- A `TracyCSetThreadIDOverride` function that overrides the current thread's ID with a given fiber ID until called again with 0.

wolfpld:
I don't use fibers. You will need to explain to me how fibers operate and what is their relation with threads. I'd like to know exactly what we're dealing with here, before deciding how to process further.

jkriegshauser:
Thanks for the reply. Fibers are like threads in that they have their own stack, registers, etc. But different in that the OS does not schedule them. Instead, they're more like co-routines (lua, python, etc.) where a thread must explicitly switch to them, and explicitly switch away from them.

When a task is queued to our task system, a Fiber is allocated for it. Once the task is finished, the Fiber is then returned to a pool waiting for a new task.

The task threads basically do the following:
1. Wait for a Fiber to be available
2. Switch context to the Fiber. The Fiber runs the task until either of the following happens:
    a. The Fiber needs to sleep or wait on a resource. A function is called so that the Thread switches off of the Fiber and puts it into a waiting queue.
    b. Or the task is finished. A function is called which makes the Thread switch off of a Fiber and puts it in the free pool.

Currently with Tracy, if a Thread needs to do 2.a. we end all of the current zones but remember them for later. When a Thread resumes a waiting Fiber, we begin again all of the zones that we remembered. This makes it look like all of the functions finished, and were called again later, and all of that time spent waiting isn't represented visually (which isn't ideal).

We'd like to be able to treat Fibers like Threads so that if a Thread isn't running a Fiber we can keep all of the zones pending and show that the Fiber is waiting.

wolfpld:
Thanks for the explanation, now I get the gist of it. I don't think fibers can be aliased to threads, due to the following reasons:

1. Possible conflicts between identifiers (a minor issue).
2. Call stack samples are mapped to threads, so you'd get no information about performance of code running in fibers.
3. You lose information about CPU core usage (threads with fiber payload can migrate or be suspended by the scheduler), which can be valuable information, e.g. if you have a hyperthreaded CPU.

What would be needed instead is:

[...]

PR #283, opened 2021-11-17
Implement a bare-bones C API for graphics profiling

expenses:
This would close https://github.com/wolfpld/tracy/issues/180.

I wrote this via a lot of trial-and-error, so it might need a few changes before being merged. I've been using this branch of the rust bindings for testing: https://github.com/expenses/rust_tracy_client/tree/gpu-testing - [specifically via running this example](https://github.com/expenses/rust_tracy_client/blob/gpu-testing/tracy-client/examples/showcase.rs). (CC https://github.com/nagisa/rust_tracy_client/issues/19)

### DIFF
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -3965,6 +3965,75 @@ TRACY_API uint64_t ___tracy_alloc_srcloc_name( uint32_t line, const char* source
     return tracy::Profiler::AllocSourceLocation( line, source, sourceSz, function, functionSz, name, nameSz );
 }
 
+TRACY_API void ___tracy_emit_gpu_zone_begin_alloc( const struct ___tracy_gpu_zone_begin_data data, int active )
+{
+    if( !active ) return;
+
+    {
+        TracyLfqPrepareC( tracy::QueueType::GpuZoneBeginAllocSrcLoc  );
+        tracy::MemWrite( &item->gpuZoneBegin.cpuTime, tracy::Profiler::GetTime() );
+        tracy::MemWrite( &item->gpuNewContext.thread, tracy::GetThreadHandle() );
+        tracy::MemWrite( &item->gpuZoneBegin.srcloc, data.srcloc );
+        tracy::MemWrite( &item->gpuZoneBegin.queryId, data.queryId );
+        tracy::MemWrite( &item->gpuZoneBegin.context, data.context );
+        TracyLfqCommitC;
+    }
+}
+
+TRACY_API void ___tracy_emit_gpu_time( const struct ___tracy_gpu_time_data data )
+{
+    {
+        TracyLfqPrepareC( tracy::QueueType::GpuTime );
+        tracy::MemWrite( &item->gpuTime.gpuTime, data.gpuTime );
+        tracy::MemWrite( &item->gpuTime.queryId, data.queryId );
+        tracy::MemWrite( &item->gpuTime.context, data.context );
+        TracyLfqCommitC;
+    }
+}
+
+TRACY_API void ___tracy_emit_gpu_zone_end( const struct ___tracy_gpu_zone_end_data data, int active )
+{
+    if( !active ) return;
+    {
+        TracyLfqPrepareC( tracy::QueueType::GpuZoneEnd );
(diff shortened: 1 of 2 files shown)


expenses:
I've been using this for a few days now and it seems to work perfectly for my uses.

wolfpld:
Using `TracyLfqPrepareC`/`TracyLfqCommit` will not work in most modern use cases. You may (and should) get away with this in case of OpenGL, which is single-threaded, but other APIs, like Vulkan or D3D12 allow submission of event from many threads. To handle this use `Profiler::QueueSerial()`/`Profiler::QueueSerialFinish()`, which will properly serialize events. Considering that the current code is valid for OpenGL, a second set of serialized functions should be provided in addition.

How are these functions supposed to be used? Is the user responsible for managing and quering the GPU state and timestamps?

client/TracyProfiler.cpp
```diff
@@ -3965,6 +3965,74 @@ TRACY_API uint64_t ___tracy_alloc_srcloc_name( uint32_t line, const char* source
     return tracy::Profiler::AllocSourceLocation( line, source, sourceSz, function, functionSz, name, nameSz );
 }
 
+TRACY_API void ___tracy_emit_gpu_zone_begin_alloc( const struct ___tracy_gpu_zone_begin_data data, int active )
+{
+    if( !active ) return;
```
wolfpld, line 3970: These checks are not neccessary. User can simply not call the function in case of `active == false`.

expenses:
> How are these functions supposed to be used? Is the user responsible for managing and quering the GPU state and timestamps?

Yes, that's my intention.

I've been using these bindings in a single-threaded Vulkan context, but it makes sense that this might not work when multithreading. Why not use `Profiler::QueueSerial()`/`Profiler::QueueSerialFinish()` for OpenGL as well though?

wolfpld:
> Why not use Profiler::QueueSerial()/Profiler::QueueSerialFinish() for OpenGL as well though?

You can safely use it, but it has to acquire a lock, instead of just using a lock-free queue. In extreme cases this can have a bit of an impact: https://github.com/wolfpld/tracy/issues/106#issuecomment-939382892

expenses:
Okay, I've added `*_serial` versions of the API and have briefly tested them.

wolfpld:
Looks good. Can you add a word or two to the documentation about these functions, how to use them, etc?

expenses:
> Looks good. Can you add a word or two to the documentation about these functions, how to use them, etc?

Sure, which bit in particular though? https://github.com/wolfpld/tracy/blob/master/manual/techdoc.tex?

wolfpld:
`tracy.tex` in the same directory. Chapter 3.13 (C API).

expenses:
Hmm unfortunately I don't know tex super well not have an editing setup for it. 

We probably just want something like this:

There are a set of C functions for GPU zones. Each function has a normal version and a thread-safe `_serial` version.

A GPU context can be created with one of the `___tracy_emit_gpu_new_context`/`___tracy_emit_gpu_new_context_serial` functions. You'll need to specify:

* `context` - a unique context id,
* `gpuTime` - an initial GPU timestamp,
* `period` - the timestamp period of the GPU. On Vulkan this is fetched via `vkGetPhysicalDeviceProperties`.
* `flags` - the flags to use. See `GpuContextFlags` in the source code.
* `type` - the GPU context type. See `GpuContextType` in the source code.

GPU contexts can be named with `___tracy_emit_gpu_context_name`/`___tracy_emit_gpu_context_name_serial`.

GPU zones can be created with `___tracy_emit_gpu_zone_begin_alloc`/`___tracy_emit_gpu_zone_begin_alloc_serial`. The `srcloc` field is the address of the source location data allocated via `___tracy_alloc_srcloc`/`___tracy_alloc_srcloc_name`. The `queryId` field is the id of the corresponding timestamp query. It should be unique on a per-frame basis.

GPU zones are ended via `___tracy_emit_gpu_zone_end`/`___tracy_emit_gpu_zone_end_serial`.

When the timestamps are fetched from the GPU (via `vkGetQueryPoolResults` on Vulkan), they must then be emitted via `___tracy_emit_gpu_time`/`___tracy_emit_gpu_time_serial`. After all timestamps for a frame are emitted, `queryId`s may be re-used.

wolfpld:
Ok, I will later add something based on the description you provided.

PR #288, opened 2021-11-26
Add fibers to TracyC.h

slembcke:
The new fiber APIs seem to work great! I added them to the C API and tried to follow your code style best as I could.

### DIFF
--- a/TracyC.h
+++ b/TracyC.h
@@ -74,8 +74,8 @@ typedef const void* TracyCZoneCtx;
 #define TracyCMessageLCS(x,y,z)
 
 #ifdef TRACY_FIBERS
-#  define TracyFiberEnter(fiber)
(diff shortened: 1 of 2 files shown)


TracyC.h
```diff
@@ -73,6 +73,11 @@ typedef const void* TracyCZoneCtx;
 #define TracyCMessageCS(x,y,z,w)
 #define TracyCMessageLCS(x,y,z)
 
+#ifdef TRACY_FIBERS
+#  define TracyFiberEnter(fiber)
+#  define TracyFiberLeave
```
wolfpld, line 78: These functions (and the second set below) should probably be named `TracyCFiber...`.

slembcke: Whoops... Not sure how I missed that. Should be updated now.

PR #290, opened 2021-12-06
Exporting cmake targets for install

Honeybunch:
I've been working on adding the TracyClient library to vcpkg: https://github.com/microsoft/vcpkg/pull/21826

This requires exporting the TracyClient library to make the library install-able and findable inside of vcpkg. As part of that PR it was suggested that I provide my small export patch upstream here for your consideration.


wolfpld:
@tksuoran, @lukasberbuer can you check if these changes don't break anything for you?

lukasberbuer:
Just tested.. works perfectly. Great addition! 👍

tksuoran:
This does not interfere with the way I use Tracy.

PR #291, opened 2021-12-07
Add CMake option TRACY_DLL for multidlls

Lectem:
The user can now easily support multi-dll setups when using CMake, by simply setting the `TRACY_DLL` option to ON.
This will create a .dll instead of static library, and any target linking tracy will automatically define `TRACY_IMPORTS`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,14 @@ project(Tracy LANGUAGES CXX)
 
 find_package(Threads REQUIRED)
 
-add_library(TracyClient TracyClient.cpp)
+option(TRACY_DLL "Should we build tracy for multiple dll support" OFF)
+
+if(TRACY_DLL)
+	add_library(TracyClient SHARED TracyClient.cpp)
+	target_compile_definitions(TracyClient PRIVATE TRACY_EXPORTS INTERFACE TRACY_IMPORTS)
+else()
+	add_library(TracyClient TracyClient.cpp)
+endif()	
 target_compile_features(TracyClient PUBLIC cxx_std_11)
 target_include_directories(TracyClient PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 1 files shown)


wolfpld:
Please take a look at #290 to see if it impacts this PR.

Honeybunch:
CMake has a built-in variable called BUILD_SHARED_LIBS that Tracy, by not specifying STATIC or SHARED currently interacts with

https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

Projects like vcpkg use BUILD_SHARED_LIBS when users select dynamically linked triplets

Lectem:
> Please take a look at #290 to see if it impacts this PR.

Sorry I didn't see that it addressed the shared library issue, it is indeed redundant.

> CMake has a built-in variable called BUILD_SHARED_LIBS that Tracy, by not specifying STATIC or SHARED currently interacts with https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html Projects like vcpkg use BUILD_SHARED_LIBS when users select dynamically linked triplets

You are correct! I somehow thought `BUILD_SHARED_LIBS` could cause issues when building with `add_subdirectory` (which was the main usage of this CMakeLists.txt before the `install` target was available), but you can just backup its value and override it locally before calling `add_subdirectory` if needed.

However please note that `TRACY_IMPORTS` (see https://github.com/wolfpld/tracy/pull/290/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR51 ) is `PUBLIC` but should probably be `INTERFACE`, since we don't want it defined when building the dll. (Thankfully tracy checks `TRACY_EXPORTS` first so it works, but we should probably not rely on it).

I'm closing this PR since the issue is solved using `BUILD_SHARED_LIBS`.

PR #294, opened 2021-12-08
Call SymLoadModuleEx for modules loaded after init (Fixes #293)

### LINKED ISSUE #293: (Windows) Tracy can not resolve symbols from renamed dlls
The project I'm working on has `vst3` modules loaded which are in fact simply `.dll`s with another extension. This seems to cause issues with the way tracy resolves symbols, which I find surprising since Visual Studio and Pix are resolving those correctly.

I'm assuming that somehow the `.pdb` is not loaded by DbgHelp even though the module does have the information.

Note that `dumpbin mymodule.vst3 /PDBPATH` does tell me it can find the pdb file `File Type: DLL PDB file found at 'D:\folder\mymodule.pdb'`.

Any pointers at what code would need to changed/fixed would be helpful.

Lectem:
Symbols were not loaded if the module is loaded dynamically after the `SymInitialize` call.
This may stall the symbol resolver thread a bit the first time a module is loaded.

### DIFF
--- a/client/TracyCallstack.cpp
+++ b/client/TracyCallstack.cpp
@@ -296,6 +296,8 @@ static const char* GetModuleName( uint64_t addr )
                     const auto res = GetModuleFileNameA( mod[i], name, 1021 );
                     if( res > 0 )
                     {
+                        // since this is the first time we encounter this module, load its symbols (needed for modules loaded after SymInitialize)
+                        SymLoadModuleEx(proc, NULL, name, NULL, (DWORD64)info.lpBaseOfDll, info.SizeOfImage, NULL, 0);
                         auto ptr = name + res;
                         while( ptr > name && *ptr != '\\' && *ptr != '/' ) ptr--;
                         if( ptr > name ) ptr++;


wolfpld:
`GetModuleName()` is called in `DecodeCallstackPtr()` before `SymFromAddr()`, but not before `SymQueryInlineTrace()`. Shouldn't this be corrected?

Lectem:
This could indeed lead to issues on the first symbol resolution of the module.
I renamed the function from `GetModuleName` to `GetModuleNameAndPrepareSymbols` and called it before `SymQueryInlineTrace`.

wolfpld:
I'm also wondering how this interferes with the manual section 3.11.1.2 Refreshing symbols list on Windows, which I totally forgot about.

Lectem:
As far as I know, there shouldn't be any issues with calling `SymLoadModule` twice, since .pdb files are unique and there checksum will be used (unless `SYMOPT_LOAD_ANYTHING` is specified to `SymSetOptions`, but this is not the case.) So I think this is pretty much safe.

However, this does mean that the section *Refreshing symbols list on Windows* needs a wording update. I think it is good to keep it there (because there are a few valid uses where the user may want to keep calling `SymLoadModule`manually, and thus be aware of `TRACY_DBGHELP_LOCK`), but perhaps we can write that tracy will attempt to load the symbols when it first encounters a module, but if extra care is needed, then the user can do it manually ?
It might also be good to mention that it might interfere with the applications own callstack resolving code (for example assertions).

wolfpld:
I think @avoroshilov had a similar use case. Andrey, can you look at these changes?

> But perhaps we can write that tracy will attempt to load the symbols when it first encounters a module, and that if extra care is needed, then the user can do it manually?
> It might also be good to mention that it might interfere with the applications own callstack resolving code (for example assertions).

Agree.

PR #303, opened 2021-12-19
Add new TRACY_ON_DEMAND_GPU_SYNC define

Lectem:
When using the tracy `TRACY_ON_DEMAND` mode, it is ok in most cases to drop measurements and do a GPU clock synchronization (that may stall, sometimes up to 10ms) during the first `tracyCollect`. This is not enabled by default in the `CMakeLists.txt` for backward compatibility and because it may be a bit intrusive.
This commit also makes the OpenGL tracy `TracyGpuZone*` a tiny bit more efficient by not calling the threadlocal `GetGpuCtx()` multiple times. It is also more resilient if no context has been declared on this thread. This means that the application will not crash if a context was used on different threads even though declared only on one (thus `GetGpuCtx().ptr == nullptr`). Tracy does not support  this scenario, so on one hand this helps users by not crashing, on the other it is an error that is now silent.

This is only one way to implement (the other would require changes to the communication protocol), and I'd understand if this was not getting merged.

### DIFF
--- a/TracyOpenGL.hpp
+++ b/TracyOpenGL.hpp
@@ -64,7 +64,7 @@ public:
 #  define TracyGpuZoneC( name, color ) TracyGpuNamedZoneC( ___tracy_gpu_zone, name, color, true )
 #  define TracyGpuZoneTransient( varname, name, active ) tracy::GpuCtxScope varname( __LINE__, __FILE__, strlen( __FILE__ ), __FUNCTION__, strlen( __FUNCTION__ ), name, strlen( name ), active );
 #endif
-#define TracyGpuCollect tracy::GetGpuCtx().ptr->Collect();
+#define TracyGpuCollect if(tracy::GetGpuCtx().ptr){tracy::GetGpuCtx().ptr->Collect();}
 
 #ifdef TRACY_HAS_CALLSTACK
 #  define TracyGpuNamedZoneS( varname, name, depth, active ) static constexpr tracy::SourceLocationData TracyConcat(__tracy_gpu_source_location,__LINE__) { name, __FUNCTION__,  __FILE__, (uint32_t)__LINE__, 0 }; tracy::GpuCtxScope varname( &TracyConcat(__tracy_gpu_source_location,__LINE__), depth, active );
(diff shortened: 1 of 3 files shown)


TracyOpenGL.hpp
```diff
@@ -173,6 +167,38 @@ class GpuCtx
...
+        MemWrite( &item->gpuNewContext.gpuTime, tgpu );
+        MemWrite( &item->gpuNewContext.thread, thread );
+        MemWrite( &item->gpuNewContext.period, period );
+        MemWrite( &item->gpuNewContext.context, m_context );
+        MemWrite( &item->gpuNewContext.flags, uint8_t(0) );
+        MemWrite( &item->gpuNewContext.type, GpuContextType::OpenGl );
+
+#if defined( TRACY_ON_DEMAND )
+        GetProfiler().DeferItem( *item );
+#elif defined( TRACY_ON_DEMAND_GPU_SYNC )
```
wolfpld, line 195: Is the condition order here the way you intended to be? `TRACY_ON_DEMAND_GPU_SYNC` does not make sense without `TRACY_ON_DEMAND`.

Lectem: Whoops sorry, forgot to repush the ammended commit before opening the PR.

Lectem: Do you prefer that I nest the `#ifdefs` ? or just in the correct order ?

wolfpld: I think that nesting is better at expressing the correct usage pattern.

Lectem: I updated all the checks to also check `TRACY_ON_DEMAND`.
The other option would have been to `#undef TRACY_ON_DEMAND_GPU_SYNC` if `TRACY_ON_DEMAND` was not defined.

wolfpld:
I have reverted this merge, as it did break my use case. The breakage was not in the profiler, but in the profiled application itself.

Lectem:
I think this may happen when reconnecting, as the current code may not detect the profiler had been disconnected+reconnected if it happens before the next call to `Collect`.
I'll try to think of a fix once I have a bit of free time.

PR #314, opened 2022-01-20
Fix OpenCLVectorAdd example

Dantali0n:
Add required library dependencies to CMake link target and fix trying to pass reference of lvalue

### DIFF
--- a/examples/OpenCLVectorAdd/CMakeLists.txt
+++ b/examples/OpenCLVectorAdd/CMakeLists.txt
@@ -11,4 +11,4 @@ add_library(TracyClient STATIC ../../TracyClient.cpp
 target_include_directories(TracyClient PUBLIC ../../)
 target_compile_definitions(TracyClient PUBLIC TRACY_ENABLE=1)
 
-target_link_libraries(OpenCLVectorAdd PUBLIC OpenCL::OpenCL TracyClient)
+target_link_libraries(OpenCLVectorAdd PUBLIC OpenCL::OpenCL TracyClient dl pthread)
--- a/examples/OpenCLVectorAdd/CMakeLists.txt
+++ b/examples/OpenCLVectorAdd/CMakeLists.txt
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 3.0)
 
 project(OpenCLVectorAdd)
 
+set(THREADS_PREFER_PTHREAD_FLAG ON)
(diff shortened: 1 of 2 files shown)


examples/OpenCLVectorAdd/CMakeLists.txt
```diff
@@ -11,4 +11,4 @@ add_library(TracyClient STATIC ../../TracyClient.cpp
 target_include_directories(TracyClient PUBLIC ../../)
 target_compile_definitions(TracyClient PUBLIC TRACY_ENABLE=1)
 
-target_link_libraries(OpenCLVectorAdd PUBLIC OpenCL::OpenCL TracyClient)
+target_link_libraries(OpenCLVectorAdd PUBLIC OpenCL::OpenCL TracyClient dl pthread)
```
wolfpld, line 14: How is this supposed to link on Windows?

Dantali0n: So we instead just want an example that does not actually compile unless people edit the files manually, or at least on Unix / Mac OS?

Do you have any other places in your code where you differentiate between OS to make a more informed compiling / linking decisions, might be better to copy that than to 'role my own'

mathisloge: for `dl` you should use `CMAKE_DL_LIBS` which does then have the correct library depending on the platform. e.g. on windows it defaults to [emtpy](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows.cmake#L19). 

see https://cmake.org/cmake/help/latest/variable/CMAKE_DL_LIBS.html

For pthreads it would be good to use the same standard cmake way. 
```cmake
set(THREADS_PREFER_PTHREAD_FLAG ON)
find_package(Threads REQUIRED)
```
and then 
```cmake
target_link_libraries(OpenCLVectorAdd PUBLIC OpenCL::OpenCL TracyClient ${CMAKE_DL_LIBS} Threads::Threads)
```


@wolfpld maybe the `cmake_minimum_required(VERSION 3.0)` should also reflect the minimum version of 3.10 of the main CMakeLists.txt

Dantali0n: Good suggestions, much cleaner solution than having to if else through cmake platforms. I have implemented the suggested changes

PR #324, opened 2022-02-11
Make Tracy usable in UWP applications

Kova93:
We want to use Tracy in our UWP application, but some Win32 APIs are unavailable on the platform, which have to be either substituted, or the features using them have to be disabled. Traditional Win32 apps are not affected by the change.

### DIFF
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -450,22 +450,22 @@ static const char* GetHostInfo()
 #  endif
     if( !GetVersion )
     {
-#    ifdef __MINGW32__
+#  ifdef __MINGW32__
         ptr += sprintf( ptr, "OS: Windows (MingW)\n" );
-#    else
+#  else
         ptr += sprintf( ptr, "OS: Windows\n" );
-#    endif
+#  endif
     }
     else
     {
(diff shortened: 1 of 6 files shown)


client/TracyProfiler.cpp
```diff
@@ -440,25 +443,29 @@ static const char* GetHostInfo()
...
-    if( !RtlGetVersion )
+#  ifdef TRACY_UWP
+    auto GetVersion = &::GetVersionEx;
+#  else
+    auto GetVersion = (t_RtlGetVersion)GetProcAddress( GetModuleHandleA( "ntdll.dll" ), "RtlGetVersion" );
+#  endif
+    if( !GetVersion )
     {
-#  ifdef __MINGW32__
+#    ifdef __MINGW32__
```
wolfpld, line 453: Is this (and following) indent change needed?

Kova93: No, that was just left there by mistake in https://github.com/wolfpld/tracy/pull/324/commits/a0a56eaf9eafc997cd3ba41ff121bd9f8a1779fc

Fixed: https://github.com/wolfpld/tracy/pull/324/commits/aa44cedcd2d539365982df6c8fb7cd818fae51b5

PR #325, opened 2022-02-11
Automatic timer fallback

BrychDaneel:
Hi. We've recently enabled TRACY_ON_DEMAND in our development build (since we measured that there are no critical performance issues about that). 
Problem we face is that sometimes we launch our binaries in virtual machine for CI tasks. And invariant TSC is unavailable there, so application crashes.  

Most convenient solution for us - firstly try to use TSC, and if it isn't available, fallback to std::chrono.

With this PR I propose following changes in logic of TRACY_TIMER_FALLBACK macros:

If TRACY_TIMER_FALLBACK is not enabled:
 * Try to use TSC 
 * Fail if it isn't available
 
If TRACY_TIMER_FALLBACK is enabled:
 * Try to use TSC 
 * If can't, use std::chrono

As a result, the same binaries may be runned with or without TSC, depending on a machine.

### DIFF
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -322,13 +322,10 @@ static void InitFailure( const char* msg )
     exit( 1 );
 }
 
-static bool checkHardwareSupportsInvariantTSC()
+static bool CheckHardwareSupportsInvariantTSC()
 {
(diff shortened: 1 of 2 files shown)


wolfpld:
Please adapt the code to follow the conventions.

BrychDaneel:
Ok, done.
I hope that I've fixed all style mismatches.

PR #339, opened 2022-03-09
Ignore uninitialized GPU contexts when rendering GPU events

zanshi:
This fixes an assertion crash when viewing GPU zones when no zones have been
recorded on a GPU context if the client is compiled with TRACY_ON_DEMAND.

Am I understanding this correctly? Is this a reasonable fix for the issue?

### DIFF
--- a/server/TracyView.cpp
+++ b/server/TracyView.cpp
@@ -19245,7 +19245,7 @@ uint64_t View::GetZoneThread( const GpuEvent& zone ) const
     {
         for( const auto& ctx : m_worker.GetGpuData() )
         {
-            assert( ctx->threadData.size() == 1 );
+            if ( ctx->threadData.size() != 1 ) continue;
             const Vector<short_ptr<GpuEvent>>* timeline = &ctx->threadData.begin()->second.timeline;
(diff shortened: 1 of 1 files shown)


wolfpld:
I'm not exactly sure what the problem here is. Can you provide a more detailed description of what is happening?

zanshi:
> I'm not exactly sure what the problem here is. Can you provide a more detailed description of what is happening?

Sure!

This is the assertion crash message:
```
../../../server/TracyView.cpp:17561: uint64_t tracy::View::GetZoneThread(const tracy::GpuEvent&) const: Assertion `ctx->threadData.size() == 1' failed.
```

We are trying to use TRACY_ON_DEMAND to be able to always have Tracy profiling accessible in our internal builds. We have multiple different OpenGL contexts that we profile using Tracy. One of these contexts is an "GlAsyncWorker" that's used to occasionally load resources into GPU memory. Thus, depending on when you connect to the client it's not guaranteed that the Tracy server has gotten any profile zones for that context.

This works:
all_contexts_active

This crashes without my fix if I hover over any zone on OpenGL context 2:
0_and_2_active_2

PR #341, opened 2022-03-10
Avoid accessing address 0 when reporting a failure.

teajay-fr:
When reporting a failure and the thread context is not valid, we shouldn't access invalid memory.

I'm not sure the "default" value 0 in the error reporting function (ZoneNameFailure and ZoneTextFailure) is appropriate but it's better than an access violation.

### DIFF
--- a/server/TracyWorker.cpp
+++ b/server/TracyWorker.cpp
@@ -5438,7 +5438,7 @@ void Worker::ProcessZoneText()
     auto td = RetrieveThread( m_threadCtx );
     if( !td )
     {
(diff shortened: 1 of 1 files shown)


wolfpld:
This code originally used `m_threadCtx`, which was deprecated in 3ede2a1, in order to properly support virtual fiber threads.

obraz

The `m_threadCtx` still tracks the originating thread and should probably be used here instead of the 0 value.

teajay-fr:
> The `m_threadCtx` still tracks the originating thread and should probably be used here instead of the 0 value.

That makes sense. I changed the code accordingly.

PR #342, opened 2022-03-10
Fixed missing value in initialisation of SymbolQueueItem.

sobolxxx:
On some compilers it is compilation error, it may be also potential crash ( uninitialised value ).

### DIFF
--- a/client/TracyProfiler.cpp
+++ b/client/TracyProfiler.cpp
@@ -3117,7 +3117,7 @@ void Profiler::SendCallstackAlloc( uint64_t _ptr )
 void Profiler::QueueCallstackFrame( uint64_t ptr )
 {
 #ifdef TRACY_HAS_CALLSTACK
-    m_symbolQueue.emplace( SymbolQueueItem { SymbolQueueItemType::CallstackFrame, ptr } );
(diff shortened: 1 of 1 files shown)


wolfpld:
Which compilers do you have in mind?

sobolxxx:
> 

Macos + ninja + llvm

wolfpld:
And what the error message you are seeing may be?

sobolxxx:
> 

_missing field 'extra' initializer_
Obviously it is because of -Wmissing-field-initializers

wolfpld:
So, you have explicitly asked the compiler to enable all the possible warnings, and then you asked it to treat these warnings as errors.

This is obviously incorrect, as the § 8.5.1.7 of the C++11 spec states:

> If there are fewer initializer-clauses in the list than there are members in the aggregate, then each member
not explicitly initialized shall be initialized from its brace-or-equal-initializer or, if there is no brace-or-equal-
initializer, from an empty initializer list (8.5.4). [ Example:
> struct S { int a; const char* b; int c; int d = b[a]; };
> S ss = { 1, "asdf" };
> initializes ss.a with 1, ss.b with "asdf", ss.c with the value of an expression of the form int{} (that
is, 0), and ss.d with the value of ss.b[ss.a] (that is, ’s’), (...) end example ]

PR #353, opened 2022-04-02
Add option to disable constexpr tokens in client markup.

biojppm:
When constexpr is enabled, these macros prevented client code from doing non-constexpr logic to retrieve strings and other properties.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,7 @@ set_option(TRACY_NO_FRAME_IMAGE  "Disable the frame image support and its thread
 set_option(TRACY_DELAYED_INIT "Enable delayed initialization of the library (init on first call)" OFF)
 set_option(TRACY_MANUAL_LIFETIME "Enable the manual lifetime management of the profile" OFF)
 set_option(TRACY_FIBERS "Enable fibers support" OFF)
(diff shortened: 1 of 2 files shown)


wolfpld:
I'm not sure I understand what you want to achieve here.

biojppm:
The problem: the use of `constexpr` keyword in macros such as `ZoneNamed()` prevents use of non-constexpr logic to retrieve the names. For example, this fails to compile:

```c++
struct profile_nodes
{
// a structure of nodes to be used at each profile point,
// each node containing a unique name, bool toggle and
// some other meta info on the profile point
};
profnode* getnode(int node) // not constexpr!
{
     // must be a static to avoid the static initialization order fiasco
     static profile_nodes nodes;
     return /*the node*/.
}

// elsewhere
void foo()
{
     ZoneNamed(getnode(kFoo)->name, foo);  // compile error
}
```
The compilation fails because `ZoneNamed()` would be instantiating a `constexpr` object from a non-`constexpr` parameter.

So this PR adds the option to remove `constexpr` from `ZoneNamed()`, which enables the code to compile successfully.

Hope this clarifies.

wolfpld:
You are looking for `ZoneTransient`.

PR #356, opened 2022-04-04
Add tool to CMake setup, improve vcpkg workflow and simple CMake Github Worklfow

theblackunknown:
## Description

This PR add `capture` and `profiler` tool to the existing CMake setup.
Moreover, vpckg workflow is improved by adding a [vcpkg manifest](https://vcpkg.io/en/docs/users/manifests.html) which simplify how to declare `vcpkg` dependencies and how to use them in a CMake project.
The above is tested through a simple CMake Github Workflow, which relies on a `CMakePresets.json` to centralize the cross-platform CMake setup (outcome example here: https://github.com/theblackunknown/tracy/actions/workflows/cmake.yml).

Changes Highlights:
- `profiler` (aka `Tracy`) can be built through the CMake setup
- `capture` can be built through the CMake setup
- Internal Tracy library are declared as CMake targets (e.g. `server`, `common`)
- CMake dependency management is improved to support cross-platform development (e.g. leverage `pkg-config` on Linux platform)
- To use CMake presets, it is required to declare an environment variable `VCPKG_ROOT` pointing to a `vcpkg` repository. 
- `capstone` dependency has been upgraded to a `5.0.0` pre-release version to allow a clean `vcpkg` integration. This is not yet (pending PR microsoft/vcpkg#23979) an official port so I have routed it to my own vcpkg registry for now.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,61 @@
+name: CMake
+
+on: push
+
+jobs:
+  build-and-install:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macOS-latest
+          - ubuntu-20.04
+          - windows-2022
+        include:
+          - os: windows-2022
+            platform: windows
+            triplet: x64-windows-static
+          - os: ubuntu-20.04
+            platform: linux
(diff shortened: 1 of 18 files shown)


theblackunknown:
Supersed by microsoft/vcpkg#24091 and #365

PR #361, opened 2022-04-08
Updating meson to perform a full build.

dave-juicelabs:


meson_options.txt
```diff
@@ -0,0 +1,19 @@
...
+option('tracy_no_exit', type : 'boolean', value : false, description : 'Client executable does not exit until all profile data is sent to server')
+option('tracy_no_sampling', type : 'boolean', value : false, description : 'Disable call stack sampling')
+option('tracy_no_verify', type : 'boolean', value : false, description : 'Disable zone validation for C API')
+option('tracy_no_vsync_capture', type : 'boolean', value : false, description : 'Disable capture of hardware Vsync events')
+option('tracy_no_frame_image', type : 'boolean', value : false, description : '"Disable the frame image support and its thread')
+option('tracy_delayed_init', type : 'boolean', value : false, description : 'Enable delayed initialization of the library (init on first call)')
+option('tracy_manual_lifetime', type : 'boolean', value : false, description : 'Enable the manual lifetime management of the profile')
+option('tracy_fibers', type : 'boolean', value : false, description : 'Enable fibers support')
+option('tracy_shared_libs', type : 'boolean', value : false, description : 'Builds Tracy as a shared object')
```
eli-schwartz, line 19: FWIW: when defining these options, you don't need to prefix it with the project name.

Meson has a different object type for variables vs. options and options are per-subproject. So these would now be `meson setup builddir -Dtracy:tracy_enable=false` and suchlike.

meson.build
```diff
@@ -1,5 +1,172 @@
...
+
+if get_option('tracy_manual_lifetime')
+  add_project_arguments('-DTRACY_MANUAL_LIFETIME', language : 'cpp')
+endif
+
+if get_option('tracy_fibers')
+  add_project_arguments('-DTRACY_FIBERS', language : 'cpp')
+endif
+
+tracy_shared_libs = get_option('tracy_shared_libs')
```
eli-schwartz, line 75: Meson has a builtin per-subproject option for this. `get_option('default_library')` can be one of `shared` or `static` or `both`

PR #362, opened 2022-04-08
make progress info and ANSI escapes conditional on being a TTY

bjacob:
This queries whether `stdout` is a terminal as opposed to another kind of file/pipe, and adjusts logging accordingly:
* Only terminals are likely to support ANSI escapes.
* Even without ANSI escapes, the data transfer progress info can make log files much larger in our experience (we run Tracy capture as part of continuous integration [here](https://buildkite.com/iree/iree-benchmark/builds)).

Notes:
* The Windows path (using `_isatty`) is untested - hope I got it right. I read Microsoft docs and saw that `<io.h>` was the right header.
* I think this actually **fixes a tiny bug** in the last ANSI escape sequence at the end of the main function in the failure case: `\033[31;1failed!` was missing the final `m` closing the escape sequence.
* Noticing: one of the places actually selects **black** foreground color.  Wouldn't that be invisible on many terminals where black is the background color?

### DIFF
--- a/capture/src/capture.cpp
+++ b/capture/src/capture.cpp
@@ -1,5 +1,6 @@
 #ifdef _WIN32
 #  include <windows.h>
+#  include <io.h>
 #else
 #  include <unistd.h>
 #endif
@@ -9,6 +10,7 @@
 #include <inttypes.h>
 #include <mutex>
 #include <signal.h>
+#include <stdarg.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


capture/src/capture.cpp
```diff
@@ -150,21 +153,25 @@ int main( int argc, char** argv )
...
-        if( mbps < 0.1f )
-        {
-            printf( "\33[2K\r\033[36;1m%7.2f Kbps", mbps * 1000.f );
-        }
-        else
-        {
-            printf( "\33[2K\r\033[36;1m%7.2f Mbps", mbps );
+        // Output progress info only if destination is a TTY to avoid bloating
+        // log files (so this is not just about usage of ANSI color codes).
+        if (isTTY) {
```
wolfpld, line 158: Please add a newline before `{`.

capture/src/capture.cpp
```diff
@@ -178,18 +185,27 @@ int main( int argc, char** argv )
...
     const auto t1 = std::chrono::high_resolution_clock::now();
 
+    const char *ansiReset = isTTY ? "\033[0m" : "";
+    const char *ansiBold = isTTY ? "\033[1m" : "";
+    const char *ansiBlack = isTTY ? "\033[30m" : "";
+    const char *ansiRed = isTTY ? "\033[31m" : "";
+    const char *ansiGreen = isTTY ? "\033[32m" : "";
+    const char *ansiYellow = isTTY ? "\033[33m" : "";
+    const char *ansiMagenta = isTTY ? "\033[35m" : "";
+    const char *ansiCyan = isTTY ? "\033[36m" : "";
```
wolfpld, line 195: Please move this code earlier (e.g. next to the `isTTY` initialization). Then please use these variables in the `if (isTTY) {` block, to improve readability. Maybe an alternate way of encoding colors would be better here? Something like:

```c++
#define AnsiRed "\033[31m"
...
const char *ansiRed = isTTY ? AnsiRed : "";
...
printf( AnsiRed "error" ... );
```

Of course the `%s` indirection would need to stay below, as it is dynamic, but in the code above the check is at entry into the block, so colors are always used there.

wolfpld:
> Noticing: one of the places actually selects **black** foreground color. Wouldn't that be invisible on many terminals where black is the background color?

Black color with bold attribute enabled should be displayed as dark grey, e.g.:

obraz

capture/src/capture.cpp
bjacob: done... went a bit farther, introduced `AnsiPrintf` function.

wolfpld: It looks fine, but please adapt this function to the coding style.

capture/src/capture.cpp
```diff
@@ -41,6 +43,51 @@ void SigInt( int )
...
+  }
+  else
+  {
+    // Just a normal printf.
+    va_list args;
+    va_start( args, format );
+    vfprintf( stdout, format, args );
+    va_end( args );
+  }
+}
```
wolfpld, line 89: Please also use 4 spaces for indentation.

bjacob: done

PR #363, opened 2022-04-12
Make `disconnect` atomic as it's written by a signal handler.

bjacob:
According to [cppreference](https://en.cppreference.com/w/cpp/utility/program/sig_atomic_t),

> Until C++11, which introduced [std::atomic](https://en.cppreference.com/w/cpp/atomic/atomic) and [std::atomic_signal_fence](https://en.cppreference.com/w/cpp/atomic/atomic_signal_fence), about the only thing a strictly conforming program could do in a signal handler was to assign a value to a volatile static std::sig_atomic_t variable and promptly return.

The `volatile` part seems important: without it, the program may not pick up updates to this value.

The `sig_atomic_t` type per se is probably not important in practice. The spec just needed a way to limit its atomicity commitment to suitably small types. `bool` is fine from that perspective.

### DIFF
--- a/capture/src/capture.cpp
+++ b/capture/src/capture.cpp
@@ -4,6 +4,7 @@
 #  include <unistd.h>
 #endif
 
+#include <atomic>
 #include <chrono>
 #include <inttypes.h>
 #include <mutex>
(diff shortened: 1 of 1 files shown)


wolfpld:
Tracy is targetting C++11, so using `std::atomic` would be preferred here. I think the relaxed memory order should be sufficient?

bjacob:
> Tracy is targetting C++11, so using `std::atomic` would be preferred here. I think the relaxed memory order should be sufficient?

Done. Added comments about why it's indeed sufficient. Note: relying on implicit zero-initialization of globals to get this atomic initialized as false while ensuring we don't introduce a global constructor here.

PR #365, opened 2022-04-13
Using vcpkg manifest for Windows

theblackunknown:
## Description

This PR adds a `vcpkg` [manifest](https://vcpkg.io/en/docs/users/manifests.html) and uses it for the Windows setup and build.
Visual Studio project are updated in a way that they will automatically build vcpkg dependencies and use them correctly when building Tracy sources.

Note that a valid `vcpkg` repository needs to be available for this to work.

More information here about the [Manifest MSBuild integration](https://vcpkg.io/en/docs/users/manifests.html#msbuild-integration).

### DIFF
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -1,10 +1,6 @@
 name: gcc
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: push
 
 jobs:
   build:
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -1,6 +1,10 @@
 name: gcc
 
-on: push
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
   build:
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -16,8 +16,6 @@ jobs:
     - uses: microsoft/setup-msbuild@v1.0.2
     - name: Integrate vcpkg
       run: vcpkg integrate install
-    - name: Build vcpkg libraries
-      run: vcpkg install freetype glfw3 capstone[arm,arm64,x86] --triplet x64-windows-static
     - name: Profiler GUI Debug
       run: msbuild .\profiler\build\win32\Tracy.vcxproj /property:Configuration=Debug /property:Platform=x64
     - name: Profiler GUI Release
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -1,6 +1,10 @@
 name: MSVC
 
-on: push
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
   build:
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -1,10 +1,6 @@
 name: MSVC
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: push
 
 jobs:
   build:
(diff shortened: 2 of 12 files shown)


theblackunknown:
@wolfpld Is this PR about using vcpkg manifest mode for Windows still of interest ?
If not I propose to close it to avoid letting it stale

wolfpld:
IIRC, there are some issues we have talked on Discord still left to resolve here. Otherwise, if this is working as I understood it, that is, you just open the solution in MSVC and the vcpkg dependencies are automatically resolved, then yes, this is something noteworthy. I'm not a big fan of having to specify a concrete vcpkg revision to use. This is *not* how libraries are intended to be used, regardless of what the masses think nowadays.

theblackunknown:
Thanks for those precisions.

> you just open the solution in MSVC and the vcpkg dependencies are automatically resolved
As of latest changes, this is what should be happening now yes.
The prior requirement is to have `vcpkg` set up somewhere and that `vcpkg integrate install` was called.

>  I'm not a big fan of having to specify a concrete vcpkg revision to use. This is not how libraries are intended to be used, regardless of what the masses think nowadays.
I have removed the `builtin-baseline` line in the manifest, which should use the current state of the `vcpkg` repository on the machine if I understand this correctly. 

As I have updated the CI to reflect those changes, can you allow it to run so that we can double check this on Github ?

wolfpld:
> As I have updated the CI to reflect those changes, can you allow it to run so that we can double check this on Github ?

Sure.

theblackunknown:
MSVC CI has passed and we can see it in actionsin the logs.
Not sure why gcc / ubuntu is failing though, this does not seem related

wolfpld:
I see you have removed the list of libraries to link with. Why so? There was some reason to not use the auto link list, but I don't remember what it was right now.

wolfpld:
Ok, we can leave it be as it is. Can you also update the vcpkg related chapters in the manual? I don't think these will be relevant anymore.

theblackunknown:
Sure, I'll take a moment to go through it later this week

theblackunknown:
I have added some details about what changed with vcpkg manifest to the manual.
I have kept the first way to use `vcpkg` (i.e. using `install_vcpkg_dependencies.bat`) as this is compatible and some people may prefer it.

wolfpld:
There are duplicate expansions in include paths: `..\..\..\vcpkg_installed\$(VcpkgTriplet)\$(VcpkgTriplet)\include\capstone`

And a propos `VcpkgTriplet`, how is it currently selected?

theblackunknown:
> There are duplicate expansions in include paths: `..\..\..\vcpkg_installed\$(VcpkgTriplet)\$(VcpkgTriplet)\include\capstone`

Yes, I thought that was a mistake but a maintainer explained to me this is intended. 
If I remember correctly this is a support a triplet selection for your solution, but then each project can override it as it is an editable project property. 
So first occurence map to the solution triplet, and the second one to the project triplet to avoid conflicting paths. 

cf. https://github.com/microsoft/vcpkg/pull/24126

wolfpld:
I hope they will be getting constant new bug reports about this.

---

> And a propos VcpkgTriplet, how is it currently selected?

This has not been answered. Tracy is built with static libraries, I can't see where this has been configured in the new setup?

theblackunknown:
> I hope they will be getting constant new bug reports about this.
> 
> > And a propos VcpkgTriplet, how is it currently selected?
> 
> This has not been answered. Tracy is built with static libraries, I can't see where this has been configured in the new setup?

The vcpkg triplet is actually part of the Visual Studio project in Tracy, for example see this line for `capture` https://github.com/wolfpld/tracy/blob/a53c8befdc914e58c0fbd1d159c11ff85d4f05df/capture/build/win32/capture.vcxproj#L18

It looks like all Visual Studio project have this correctly set up.

wolfpld:
The manual does not build because you have not escaped the underscore: `\texttt{VCPKG_ROOT}`.

wolfpld:
Please remove the last commit.

theblackunknown:
> Please remove the last commit.

I'll rebase and remove those.

theblackunknown:
Thanks for your patience with me @wolfpld !

PR #369, opened 2022-04-21
Fix column widths for hw counts and source location

bjacob:
Fixes:
* In the asm view, the source location column caused subsequent columns to be misaligned, and the asm jump arrows were often painted over the disassembled code.
* In both the source and asm views, the hw counters columns caused subsequent columns to be misaligned.

Note: I'm running the Tracy profiler on desktop Linux. I don't know to what extent the issues that I was seeing were specific to that.

Before: tracy-bad

After: tracy-good

### DIFF
--- a/server/TracySourceView.cpp
+++ b/server/TracySourceView.cpp
@@ -3069,6 +3069,7 @@ void SourceView::RenderLine( const Tokenizer::Line& line, int lineNum, const Add
     const bool showHwSamples = worker && m_hwSamples && worker->GetHwSampleCountAddress() != 0;
     if( showHwSamples )
     {
+        const auto startPos = ImGui::GetCursorScreenPos();
         if( hasHwData )
         {
             if( m_hwSamplesRelative )
@@ -3088,10 +3089,11 @@ void SourceView::RenderLine( const Tokenizer::Line& line, int lineNum, const Add
                 RenderHwLinePart( cycles, retired, branchRetired, branchMiss, cacheRef, cacheMiss, 0, 0, 0, 0, ts );
             }
         }
-        else
-        {
-            ImGui::ItemSize( ImVec2( 17 * ts.x, ts.y ) );
-        }
+        ImGui::SameLine( 0, 0);
+        const auto endPos = ImGui::GetCursorScreenPos();
+        const auto itemsWidth = ( endPos - startPos ).x;
+        const auto fixedWidth = 17 * ts.x;
+        ImGui::ItemSize( ImVec2( fixedWidth - itemsWidth, 0 ) );
         ImGui::SameLine( 0, ty );
     }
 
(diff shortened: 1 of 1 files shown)


wolfpld:
Your solution has some problems:

obraz

--

The layout should be correct regardless of what is displayed and the code tries to calculate this correctly, and at least on some platforms and locale settings, it succeeds. I'd rather see that the root cause of the problem is found.

bjacob:
Thanks, I have pushed another commit that fixes a definite issue in my code that might be what you were observing. In order for my approach to work, the `startPos` must be recorded just before we start printing the text, that is, after the `SmallColorBox`. Otherwise, there is no reason why the `ItemSize` of 32 character widths would be sufficient to space this column.

Does it work better now?

I didn't see a good way to fix the existing approach. One issue is that 'Before' screenshot shows lines falling out of alignment by a fractional multiple of character width, so we couldn't fix that by appending integral multiples of character width. Another possible issue (I'm not 100% sure) is that we weren't precislely accounting for the width of graphical (non-text) items such as SmallColorBox. In the end, trying to do accurrate book-keeping of already-drawn width to compensate precisely for it, seemed much harder than just recording the cursor position and compensating for the actual cursor position difference.

The approach in this PR is effectively relying on ImGui to tell us how wide was what we have drawn, instead of trying to implemment our own computation of that, trying to keep that in sync with what the ImGui drawing actually does.

bjacob:
Note: I have updated the "After" picture in the PR description. Following the above fix, the source-location column is now separated from the jump-arrows by about the width of the `SmallColorBox`.  Before, we weren't accounting for that and it was only accidental that on my machine the jump arrows appeared right next to the source-location without overlapping it.

wolfpld:
> at least on some platforms and locale settings

Ah, so this code uses `std::to_chars`. This function is designed to be as efficient as possible at the very specific little task it was designed to do. No locale nonsense, etc. As you can see at the top of `TracyPrint.hpp`, `std::to_chars` can be used only if a complicated web of `#ifdefs` is satisfied. If you would use an older compiler, which doesn't have this implemented, the code has to fallback to an equivalent `sprintf()`. This fallback happens to be working exactly as expected in my setup.

You will be probably wondering why the `std::to_chars` path is completely disabled at the slightest hint of `__GNUC__` being defined, instead of doing a proper check for the availablity of headers, as is done in the MSVC path. This is because some wondrous libstdc++ developer has decided to implement the `<charconv>` functionality only partially. In essence, the header is available, you can include it, but when you try to use the one function you need and which standard dictates had to be there, you only get an compilation error, because it's in another castle.

Due to this, `std::to_chars` cannot be ever used on Linux, because there is no way of checking if it is implemented. And there will always be some museum curator with a Debian LTS from two decades ago, who will encounter this exact issue.

> Does it work better now?

It looks ok now.

bjacob:
Thanks for the explanation! I actually didn't know about `std::to_chars` at all; now I'll also know to stay away from it in the near future. And it's good to know a reason why we had slightly different experiences with the above intermediate state of this PR.

wolfpld:
Don't be discouraged from using it, if it works for you (e.g. you don't have to support not-the-latest standard library, and so on), as it is much faster than going the printf way - by an order of magnitude.

PR #374, opened 2022-04-27
Adding TRACY_NO_CRASH_HANDLER flag to allow for disabling the crash handler

robertblaketaylor:
We've got cases where we need to disable the crash handler because it conflicts with our own crash handling code or causes an incompatibility in some of our integration environments (ex. under Unity).


robertblaketaylor:
Need to update my change

PR #375, opened 2022-04-27
Add flag to disable crash handler

robertblaketaylor:
We needed to add a flag to disable the crash handler as this was causing issues in our integration, maybe its useful for other folks as well?


wolfpld:
Please update the manual to describe the changes made.

robertblaketaylor:
Let me know if there is anything I'm missing.

PR #377, opened 2022-05-01
Fix build options, add TRACY_TIMER_FALLBACK option.

Net5F:
A few build option changes. Feel free to edit or let me know if any are undesirable or should be split into separate PRs.

Changes:
Renamed TRACY_NO_SYS_TRACE -> TRACY_NO_SYSTEM_TRACING to match the build flag name. Unlike the meson logic, the CMake logic directly maps the option name to the build flag that is injected. With the mismatched name, the flag wasn't being properly applied.

Added TRACY_TIMER_FALLBACK option to expose the same-named flag.

Moved signal.h include to get sigaction definition that was missing when TRACY_NO_CALLSTACK is defined.


wolfpld:
Can you provide more context for the relocation of `#include <signal.h>`?

Net5F:
> Can you provide more context for the relocation of `#include <signal.h>`?

Sure thing, when I enabled TRACY_NO_CALLSTACK on linux, I would get the following:
```
TracyProfiler.hpp:939:41: error: field ‘segv’ has incomplete type ‘tracy::sigaction’
```

I looked for where the include was coming from when TRACY_NO_CALLSTACK was not defined, but was unable to find it. Moving the `signal.h` include to the header resolved the missing definition so I moved on, but I don't have enough context to know if it's the best solution.

PR #380, opened 2022-05-04
Slight tweak to get client library working with mingw

Honeybunch:
Opening this as a draft because I'm sure there are some side effects here that I'm not exactly privy to.

Tested against:
```
gcc --version
gcc (GCC) 11.2.0
```

I was curious about getting tracy working with Mingw for a side project of mine. Dug in a bit and came up with this change to just disable the body of SetupVsync when compiling with mingw. That and adding some public dependencies to make it easier to consume from CMake and this worked find with a test project of mine. However I didn't really do a complete exhaustive test. I figured discussion might be warranted / there might be a good reason why this isn't fit for upstream. I came across some gitter chat archive where mingw was known to be broken and purposefully disabled on CI but that was in the context of Mingw 8.1.0 IIRC.

Thoughts? Ideas for other stuff I should test with this?

### DIFF
--- a/client/TracySysTrace.cpp
+++ b/client/TracySysTrace.cpp
@@ -265,7 +265,7 @@ void WINAPI EventRecordCallbackVsync( PEVENT_RECORD record )
 
 static void SetupVsync()
 {
-#if _WIN32_WINNT >= _WIN32_WINNT_WINBLUE && !__MINGW32__
+#if _WIN32_WINNT >= _WIN32_WINNT_WINBLUE && !defined(__MINGW32__)
     const auto psz = sizeof( EVENT_TRACE_PROPERTIES ) + MAX_PATH;
     s_propVsync = (EVENT_TRACE_PROPERTIES*)tracy_malloc( psz );
     memset( s_propVsync, 0, sizeof( EVENT_TRACE_PROPERTIES ) );
--- a/client/TracySysTrace.cpp
+++ b/client/TracySysTrace.cpp
@@ -265,7 +265,7 @@ void WINAPI EventRecordCallbackVsync( PEVENT_RECORD record )
 
 static void SetupVsync()
 {
-#if _WIN32_WINNT >= _WIN32_WINNT_WINBLUE
+#if _WIN32_WINNT >= _WIN32_WINNT_WINBLUE && !__MINGW32__
     const auto psz = sizeof( EVENT_TRACE_PROPERTIES ) + MAX_PATH;
     s_propVsync = (EVENT_TRACE_PROPERTIES*)tracy_malloc( psz );
     memset( s_propVsync, 0, sizeof( EVENT_TRACE_PROPERTIES ) );
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,11 @@ target_link_libraries(
         ${CMAKE_DL_LIBS}
 )
 
+# Public dependency on some libraries required when using Mingw
+if(WIN32 AND ${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")
+    target_link_libraries(TracyClient PUBLIC ws2_32 dbghelp)
+endif()
+
 if(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
     find_library(EXECINFO_LIBRARY NAMES execinfo REQUIRED)
     target_link_libraries(TracyClient PUBLIC ${EXECINFO_LIBRARY})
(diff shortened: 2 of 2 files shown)


wolfpld:
What does this fix? What is the original issue? Is mingw not providing some APIs which should be available at a given Windows API compatibility level?

Honeybunch:
Great question! Should have mentioned that to begin with. 

Without this change this is what happens if you try to compile with g++ & mingw on windows:
```
λ cmake --build build-mingw\
[1/2] Building CXX object CMakeFiles/TracyClient.dir/Debug/TracyClient.cpp.obj
FAILED: CMakeFiles/TracyClient.dir/Debug/TracyClient.cpp.obj
C:\Users\Arsine\scoop\apps\gcc\current\bin\g++.exe -DTRACY_ENABLE -DCMAKE_INTDIR=\"Debug\" -isystem C:/Users/Arsine/Git/tracy -g -MD -MT CMakeFiles/TracyClient.dir/Debug/TracyClient.cpp.obj -MF CMakeFiles\TracyClient.dir\Debug\TracyClient.cpp.obj.d -o CMakeFiles/TracyClient.dir/Debug/TracyClient.cpp.obj -c C:/Users/Arsine/Git/tracy/TracyClient.cpp
In file included from C:/Users/Arsine/Git/tracy/TracyClient.cpp:26:
C:/Users/Arsine/Git/tracy/client/TracySysTrace.cpp: In function 'void tracy::SetupVsync()':
C:/Users/Arsine/Git/tracy/client/TracySysTrace.cpp:303:5: error: 'EVENT_FILTER_EVENT_ID' was not declared in this scope; did you mean 'EVENT_FILTER_HEADER'?
  303 |     EVENT_FILTER_EVENT_ID fe = {};
      |     ^~~~~~~~~~~~~~~~~~~~~
      |     EVENT_FILTER_HEADER
C:/Users/Arsine/Git/tracy/client/TracySysTrace.cpp:304:5: error: 'fe' was not declared in this scope
  304 |     fe.FilterIn = TRUE;
      |     ^~
C:/Users/Arsine/Git/tracy/client/TracySysTrace.cpp:311:17: error: 'EVENT_FILTER_TYPE_EVENT_ID' was not declared in this scope; did you mean 'EVENT_FILTER_TYPE_SCHEMATIZED'?
  311 |     desc.Type = EVENT_FILTER_TYPE_EVENT_ID;
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~
      |                 EVENT_FILTER_TYPE_SCHEMATIZED
C:/Users/Arsine/Git/tracy/client/TracySysTrace.cpp:315:29: error: 'EVENT_ENABLE_PROPERTY_IGNORE_KEYWORD_0' was not declared in this scope
  315 |     params.EnableProperty = EVENT_ENABLE_PROPERTY_IGNORE_KEYWORD_0;
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ninja: build stopped: subcommand failed.
```
The mingw headers don't seem to define `EVENT_FILTER_EVENT_ID`. I'm not sure what the state of the Mingw `EnableTraceEx2` implementation is if that's the case too. But maybe I'm just missing a header somewhere?

wolfpld:
> The mingw headers don't seem to define `EVENT_FILTER_EVENT_ID`. I'm not sure what the state of the Mingw `EnableTraceEx2` implementation is if that's the case too. But maybe I'm just missing a header somewhere?

According to https://docs.microsoft.com/en-us/windows/win32/api/evntprov/ns-evntprov-event_filter_event_id `evntprov.h` should be included. Currently it isn't. Can you check if this would fix your issue?

Honeybunch:
Just including `evntprov.h` doesn't work because mingw's version of that header is missing the necessary type definitions: https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-headers/include/evntprov.h

client/TracySysTrace.cpp
```diff
@@ -265,7 +265,7 @@ void WINAPI EventRecordCallbackVsync( PEVENT_RECORD record )
 
 static void SetupVsync()
 {
-#if _WIN32_WINNT >= _WIN32_WINNT_WINBLUE
+#if _WIN32_WINNT >= _WIN32_WINNT_WINBLUE && !__MINGW32__
```
wolfpld, line 268: Please rewrite `!__MINGW32__` into a more verbose and explicit form. I don't know if this `!` is a negation requiring `__MINGW32__` to exist and how it will work when the macro is not defined (which is the main use case to consider here).

Honeybunch: Updated to `!defined(__MINGW32__)` which should be more explicit; we care about whether or not the __MINGW32__ define exists, not what its value is.

wolfpld:
Everything seems to be in order, but I am not allowed to merge a draft PR.

Honeybunch:
Bump now that the PR is no longer a draft

PR #383, opened 2022-05-04
(Do Not Pull) Fix up baseAddr by walking back from symAddr

bjacob:
Not a real pull request. Just to illustrate issue #384.


bjacob:
Note: https://github.com/iree-org/iree/issues/9717 is an instance where a variant of this patch doing `baseAddr -= 1;` instead of `-= 4` was found useful on x86.

PR #401, opened 2022-06-01
Add ability to filter callstacks in memory tab by inactive allocations.

thedmd:
image

Filtering by inactive allocations helps to pin point wasteful allocations in the app.


wolfpld:
Please update the manual.

thedmd:
Done

PR #412, opened 2022-06-17
Fix building on Windows without `VCPKG_ROOT` env var set

### LINKED ISSUE #409: Running install_vcpkg_dependencies.bat seems to no longer work
Running `vcpkg\install_vcpkg_dependencies.bat`, as in the manual's "Local installation within the project directory", on a fresh clone of [tracy (0b2e276)](https://github.com/wolfpld/tracy/commit/0b2e276dd2d96b475b38c737a7fc29bc74eaad67) results in the following error:
```
Error: In manifest mode, `vcpkg install` does not support individual package arguments.
To install additional packages, edit vcpkg.json and then run `vcpkg install` without any package arguments.
See https://github.com/Microsoft/vcpkg/tree/master/docs/users/manifests.md for more information.
Using manifest file at C:\Users\laurel\dev\tmp\tracy\vcpkg.json.
```
After locally reverting the merge of https://github.com/wolfpld/tracy/pull/365 it worked flawlessly:
```
Computing installation plan...
The following packages will be built and installed:
  * brotli[core]:x64-windows-static -> 1.0.9#2
  * bzip2[core]:x64-windows-static -> 1.0.8#2
    capstone[arm,arm64,core,x86]:x64-windows-static -> 5.0.0-rc2#1
    freetype[brotli,bzip2,core,png,zlib]:x64-windows-static -> 2.12.1
    glfw3[core]:x64-windows-static -> 3.3.7
  * libpng[core]:x64-windows-static -> 1.6.37#17
  * vcpkg-cmake[core]:x64-windows -> 2022-05-10#1
  * vcpkg-cmake-config[core]:x64-windows -> 2022-02-06#1
  * zlib[core]:x64-windows-static -> 1.2.12#1
...
```

theblackunknown:
# Description

This PR fixes #409 by adding path to where `scripts/install_vcpkg_dependencies.bat` will install dependencies if used.

## Details

To build Tracy projects on Windows, we rely on `vcpkg` to pull third party dependencies.
There are two ways to use `vcpkg` in Tracy:
1. Either you call `scripts/install_vcpkg_dependencies.bat`, then open Visual Studio project and build them.
2. You rely on `vcpkg` msbuild integration and you can simply open Visual Studio project and build them.

The 2. requires that you have previously clone `vcpkg` somewhere on your machine and set and environment variable `VCPKG_ROOT` to that location.
Whereas 1. previously did not have this requirement but it now has.

In #365 , I broke 1. by accident as all my tests were on machine where `VCPKG_ROOT` is set (this includes Github CI).
So in this PR I did test more carefully, restored and updated Visual Studio project file so they work both in 1. and 2.

I updated `scripts/install_vcpkg_dependencies.bat` to locally set `VCPKG_ROOT` and to not list which dependencies to pull: as we use `vcpkg` in manifest mode, all dependencies are now specified in a single place: `vcpkg.json`.

Last but not least, I added a CI job to msvc.yml so that those two scenarios are tested.

### DIFF
--- a/vcpkg/install_vcpkg_dependencies.bat
+++ b/vcpkg/install_vcpkg_dependencies.bat
@@ -1,14 +1,17 @@
 @echo off
+setlocal
 
 pushd %~dp0
 
 REM get vcpkg distribution
-if not exist vcpkg git clone https://github.com/Microsoft/vcpkg.git
+if not exist vcpkg git clone https://github.com/Microsoft/vcpkg.git || exit /b 1
 
 REM build vcpkg
-if not exist vcpkg\vcpkg.exe call vcpkg\bootstrap-vcpkg.bat -disableMetrics
+if not exist vcpkg\vcpkg.exe call vcpkg\bootstrap-vcpkg.bat -disableMetrics || exit /b 2
+
+set VCPKG_ROOT=%cd%\vcpkg
 
 REM install required packages
-vcpkg\vcpkg.exe install --triplet x64-windows-static
+vcpkg\vcpkg.exe install --triplet x64-windows-static || exit /b 3
 
 popd
--- a/vcpkg/install_vcpkg_dependencies.bat
+++ b/vcpkg/install_vcpkg_dependencies.bat
@@ -9,6 +9,6 @@ REM build vcpkg
 if not exist vcpkg\vcpkg.exe call vcpkg\bootstrap-vcpkg.bat -disableMetrics
 
 REM install required packages
-vcpkg\vcpkg.exe install --triplet x64-windows-static freetype glfw3 capstone[arm,arm64,x86]
+vcpkg\vcpkg.exe install --triplet x64-windows-static
 
 popd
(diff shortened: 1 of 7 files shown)


vcpkg/install_vcpkg_dependencies.bat
```diff
@@ -1,14 +1,17 @@
-@echo off
+rem @echo off
```
theblackunknown, line 1: Testing leftover, revert it

theblackunknown:
@wolfpld @laurelkeys this PR seems to be in a good shape, can you give it a round of testing and review so that we can solve #409 please ?

wolfpld:
There's no need to provide two sets of build artifacts. I'm also not sure if builing everything twice on MSVC is the right approach.

theblackunknown:
Indeed there is no need to publish two sets of build artifacts, however I see some value to build twice: one is testing `vcpkg integration` whereas the other is testing without it (which was what caused #409).

wolfpld:
Maybe it would be viable to just build a small subset of projects, which would be enough to test the functionality? What do you think?

theblackunknown:
> Maybe it would be viable to just build a small subset of projects, which would be enough to test the functionality? What do you think?

Great idea, let's restrict the second build job to `Tracy` and `capture`: those are the two most common ones from my experience.

PR #416, opened 2022-06-28
Fix Meson options to match build

kklobe:
Some options in `meson_options.txt` were out of sync with `meson.build`, causing errors during `meson setup ...`.

If it would be at all possible to get a minor release tag for this, then we can integrate the source archive directly in the Meson WrapDB repository with no patches required - thanks!


wolfpld:
https://github.com/wolfpld/tracy/releases/tag/v0.8.2-meson-hotfix

wolfpld:
Pushed.

kklobe:
Thank you for the incredibly speedy responses. Official Meson WrapDB is already live with the new version.

PR #434, opened 2022-07-27
fix tracy worker threads conflict

haCo77:
tracy [launches multiple threads](https://github.com/wolfpld/tracy/blob/cb58b4803a2beb36d4c6cf89593865af1ef6fec3/server/TracyWorker.cpp#L1814) exec ReconstructMemAllocPlot(). In that function, [it uses lock to coordinate memory allocations between threads, but forgot to include plot->data.reserve_exact() in the locked scope too, so sometimes two data have overlap in the allocated memory block](https://github.com/wolfpld/tracy/blob/cb58b4803a2beb36d4c6cf89593865af1ef6fec3/server/TracyWorker.cpp#L7112-L7119).


wolfpld:
Thanks.

PR #437, opened 2022-07-28
Meson fix

### LINKED ISSUE #436: Meson integration is broken
After changes in 06c7984a the meson integration no longer works.

@Net5F @tmayoff @kklobe @dave-juicelabs please take a look, at your leisure.

tmayoff:
This should fix the meson build, closes #436.

The only thing I'm not sure about is the public include dir, should it just be `public` making the include `#include <tracy/Tracy.hpp>` or `public/tracy` and `#include <Tracy.hpp>`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -92,59 +92,59 @@ endif
 threads_dep = dependency('threads')
 
 includes = [
-    'TracyC.h',
-    'Tracy.hpp',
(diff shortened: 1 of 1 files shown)


wolfpld:
Both include schemas are intended to be usable. I prefer to use `#include <tracy/Tracy.hpp>`.

PR #440, opened 2022-08-03
Using push/pop in all pragma pack pair to avoid potential padding bugs

Danielmelody:


public/common/TracySystem.cpp
```diff
@@ -106,7 +106,7 @@ std::atomic<ThreadNameData*>& GetThreadNameData();
 #endif
 
 #ifdef _MSC_VER
-#  pragma pack( push, 8 )
```
wolfpld, line 109: Do not change the coding style.

PR #444, opened 2022-08-12
Added profiler and deps to meson build

tmayoff:
Right now the only issue is that the profiler profiles itself when ran (relying on TracyClient, rather than the common files directly, I believe is the problem).


wolfpld:
I will need some explanation for what this is.

tmayoff:
This adds the profiler GUI to the meson build system to allow it be built through meson

tmayoff:
Also unfortunately I can no longer run the profiler at all (built via meson, or make), due to a failure in glfw (seems to be an nvidia driver thing I think) to I can't test this until it's fixed

wolfpld:
I do not understand the point of this. The existing cmake/meson integrations are meant to enable one-liner enablement of Tracy in external projects. But what is the point of building the GUI through meson? Why is only GUI build here, what about the rest of utilities? Also, this is essentially an alternative build system to the already existing one. I do not understand the expectation here, am I supposed to leave the build system that is currently working for me and start using this one out of the blue? If not, who will be responsible for maintining it? There are already changes on the repo (removal of Wayland build-time define) that make it not up-to-date. More are coming.

tmayoff:
The idea was easing my and my teams workflow a bit by keeping what we can withing meson, we can ensure the same version of tracy client and the profiler, and build everything together without needing anything outside our own repos. I was not aware the meson/cmake were never meant to be full build systems (I've seen in other projects where this was the case) I assumed just no one had extended it. I only had the GUI here as that's the utility we use the most. I wouldn't mind maintaining it, but I guess it's ultimately up to you if you want this in the project at all

PR #446, opened 2022-08-21
Add WASD panning/zooming functionality to timeline.

simplyWiri:
While doing this, I imlemented some clamping logic for scrolling outside the range of the timezone:
```c++
        nextTimelineRangeStart = std::max<int64_t>(std::min(nextTimelineRangeStart, lastKnownTime - 50), 0);
        nextTimelineRangeEnd = std::max<int64_t>(std::min(nextTimelineRangeEnd, lastKnownTime), 1);

        if ( nextTimelineRangeEnd - nextTimelineRangeStart <= 50 ) return;
        ZoomToRange( nextTimelineRangeStart, nextTimelineRangeEnd, !m_worker.IsConnected() || m_viewMode == ViewMode::Paused );
```

I was wondering if I should potentially add this logic to the mouse scrolling too?

Its currently a bit janky in that, if you zoom out with the mouse, beyond the bounds of the timeline - and then use the wasd keyboard navigation, it will 'snap' (still animated) into the bounds of the timeline.

Was unsure if being able to scroll outside the bounds of the timeline was intentional behaviour or not though - let me know.

### DIFF
--- a/server/TracyView.hpp
+++ b/server/TracyView.hpp
@@ -1,6 +1,7 @@
 #ifndef __TRACYVIEW_HPP__
 #define __TRACYVIEW_HPP__
 
+#include <array>
 #include <atomic>
 #include <functional>
 #include <memory>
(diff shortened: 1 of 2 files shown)


wolfpld:
Please use the existing code style.

> I was wondering if I should potentially add this logic to the mouse scrolling too?

Current behavior is intentional.

PR #450, opened 2022-08-28
Add right click jump to assembly loc in symbol view

simplyWiri:
I feel like intuitively this should be a left click operation, but that is likely a matter of preference - and consistency is probably more important here.

(I'd also argue shift+click is more natural for 'selecting' a line, rather than left click)


wolfpld:
What exactly is this doing?

simplyWiri:
If you right click on some line in the CPP source in the symbol view, the assembly panel will open (existing behaviour) + will scroll to the assembly associated with that line (new behaviour)

wolfpld:
> will scroll to the assembly associated with that line (new behaviour)

This is an already existing behavior. I don't understand what you are trying to achieve here.

simplyWiri:
I guess I hadn't tried right clicking the line prior - or some other change of mine locally had changed this. Apologies!

PR #451, opened 2022-08-29
Add assembly syntax colouring

simplyWiri:
image

Makes the asm view a little easier (or harder) on the eyes. I bear no attachment to the chosen colours - This is a pretty simple change, just adds some basic syntax highlighting for the asm, splitting registers/directives/opcodes.

There is still some work to be done in:
- [x] Formatting
- [x] Renaming `Tokenizer::AsmOperandString` to something more to the point
- [x] Implementing any feedback
- [x] Support for ARM assembly
- [x] Fix bug wherein the 'click' to select an element is currently the last token rendered. Entire group should be clickable

I also suspect there may be a desire to move the storage of all the assembly information into its own struct similar to `TracySourceContents.hpp`, however I am not planning to do this in this change.

Would like to hear any opinions you have regarding the implementation / suggestions regarding alternate approaches.

I'm also interested in any feedback regarding ideas for reducing the amount of space taken up in the margins for both the source / assembly views for things like hardware counters / src -> assembly line count.

### DIFF
--- a/server/TracySourceTokenizer.hpp
+++ b/server/TracySourceTokenizer.hpp
@@ -2,7 +2,9 @@
 #define __TRACYSOURCETOKENIZER_HPP__
 
 #include <stdint.h>
+#include <cstdint>
 #include <vector>
+#include <string>
 
 namespace tracy
 {
@@ -24,11 +26,28 @@ public:
         Special
     };
 
+    enum class AsmTokenColor : uint8_t
+    {
+        Mnemonic,       // no-op, padding
+        Label,          // no-op, padding
+        Default,        // '+', '[', '*', etc
+        SizeDirective,   // byte, word, dword, etc
+        Register,       // rax, rip, etc
+        Literal,        // 0x04, etc
+    };
+
     struct Token
     {
         const char* begin;
         const char* end;
         TokenColor color;
+    };
+
+    struct AsmToken
+    {
+        uint8_t beginIdx;
+        uint8_t endIdx;
+        AsmTokenColor color;
     };
 
     struct Line
@@ -38,9 +57,16 @@ public:
         std::vector<Token> tokens;
     };
 
+    struct AsmOperand
+    {
+        std::string string;
+        std::vector<AsmToken> tokens;
+    };
+
     Tokenizer();
 
     std::vector<Token> Tokenize( const char* begin, const char* end );
+    AsmOperand TokenizeAsmOperand( const char assemblyText[160] );
 
 private:
     TokenColor IdentifyToken( const char*& begin, const char* end );
--- a/server/TracySourceView.cpp
+++ b/server/TracySourceView.cpp
@@ -4,6 +4,7 @@
 
 #include <capstone.h>
 
+#include "TracySourceTokenizer.hpp"
 #include "imgui.h"
 #include "TracyCharUtil.hpp"
 #include "TracyColor.hpp"
@@ -875,7 +876,7 @@ bool SourceView::Disassemble( uint64_t symAddr, const Worker& worker )
                     }
                 }
             }
-            m_asm.emplace_back( AsmLine { op.address, jumpAddr, op.mnemonic, op.op_str, (uint8_t)op.size, leaData, jumpConditional, std::move( params ) } );
+            m_asm.emplace_back( AsmLine { op.address, jumpAddr, op.mnemonic, m_tokenizer.TokenizeAsmOperand( op.op_str ), (uint8_t)op.size, leaData, jumpConditional, std::move( params ) } );
 
 #if CS_API_MAJOR >= 4
             auto& entry = m_asm.back();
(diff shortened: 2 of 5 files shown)


wolfpld:
I'm not sure if tokenization of a string is the way to go here. For one, the current implementation only works on x86/x64. Capstone provides rich information about disassembly, have you considered using that information to obtain a generic solution that would work i.e. on ARM?

simplyWiri:
I had a brief look into this after it was suggested, I think for simplicity this approach would be favourable.

The current formatting for the operand string is implemented in:
https://github.com/capstone-engine/capstone/blob/master/arch/X86/X86IntelInstPrinter.c / https://github.com/capstone-engine/capstone/blob/master/arch/X86/X86ATTInstPrinter.c.

To move this internally would be to replicate this code, while it is unlikely to change, its a lot of code bloat for a relatively minor feature (note a lot could probably be skimmed by ignoring opcodes, and just doing operands). 

I also considered an alternate approach wherein it leverages information inside the opcode struct, however I'm unsure if there are any benefits to this approach (only one I can see would be dropping the list of register names), in return for the additional complexity.

server/TracySourceView.cpp
```diff
@@ -3843,69 +3843,48 @@ void SourceView::RenderAsmLine( AsmLine& line, const AddrStat& ipcnt, const Addr
...
-    {
-        TextColoredUnformatted( ImVec4( 1, 0.25f, 0.25f, isInContext ? 1.f : 0.5f ), buf );
-    }
-    else if( line.regData[0] != 0 )
-    {
-        bool hasDepencency = false;
-        int idx = 0;
-        for(;;)
-        {
-            if( line.regData[idx] == 0 ) break;
```
simplyWiri, line 3873: Also worth noting that I removed this code path ; Unsure how to visualize this if it is desirable to bring back

server/TracySourceTokenizer.hpp
```diff
@@ -2,15 +2,17 @@
...
 #include <vector>
+#include <string>
 
 namespace tracy
 {
 
 class Tokenizer
 {
 public:
-    enum class TokenColor : uint8_t
```
wolfpld, line 13: Leave the `TokenColor` name as it is.

server/TracySourceTokenizer.hpp
```diff
@@ -24,26 +26,50 @@ class Tokenizer
         Special
     };
 
-    struct Token
```
wolfpld, line 27: Same here.

server/TracySourceView.cpp
```diff
@@ -3843,69 +3843,48 @@ void SourceView::RenderAsmLine( AsmLine& line, const AddrStat& ipcnt, const Addr
...
+    int jumpLabel = 0;
     if( line.jumpAddr != 0 )
     {
         auto lit = m_locMap.find( line.jumpAddr );
         if( lit != m_locMap.end() )
         {
+            jumpLabel = lit->second;
             char tmp[64];
-            sprintf( tmp, ".L%" PRIu32, lit->second );
-            strcpy( buf+m_maxMnemonicLen, tmp );
```
wolfpld, line 3853: I have asked you before to use the existing code style. This change does nothing *but* changes the style to be non-conforming.

server/TracySourceView.cpp
wolfpld: Why was it removed?

simplyWiri: Was unsure how to make it immediately obvious _what_ was being represented. 

Dimming things out of focus is a familiar concept to most, but unsure how to represent this - I also was not sure exactly what this was representing / what the idea behind it was.

server/TracySourceView.cpp
simplyWiri: Can do; Is there a reason there is no clang-format file? These are artifacts of IDE fixups when copy pasting code around - very easy on my end to miss

server/TracySourceView.cpp
wolfpld: And what would this "this" be?

server/TracySourceView.cpp
wolfpld: > Is there a reason there is no clang-format file?

Because someone would have to make it.

> These are artifacts of IDE fixups

Use different IDE then.

> very easy on my end to miss

I would expect people to actually look at the changes they propose (i.e. by running a `git diff` before commiting changes). Otherwise it may look like they do not respect my time.

server/TracySourceView.cpp
simplyWiri: Wasn't sure how to make the visualisation make sense. I've added it back locally to just be not dimmed. I.e. the logic is just
```c++
    const bool isInContext = IsInContext( worker, line.addr );
    const auto selected = m_asmSelected == asmIdx;
    auto transitiveDependency = false;
    for(int i = 0; line.regData[i] != 0; i++) {
        transitiveDependency |= line.regData[i] & ( WriteBit | ReadBit );
    }

    const auto shouldFocus = (isInContext && m_asmSelected == -1) || selected || transitiveDependency;
    const auto& colorPalette = shouldFocus ? AsmSyntaxColors : AsmSyntaxColorsDimmed;
```

wolfpld:
No longer applicable.

PR #454, opened 2022-08-31
CMake: Add option to explicitly set lib type 

Krzmbrzl:
With this new option, it is now possible to explicitly build Tracy as a
shared or static library independent from the BUILD_SHARED_LIBS
variable, which always acts on a global scope (thus, affecting all CMake
targets).

If the options is not explicitly given, it will default to whatever
BUILD_SHARED_LIBS would indicate, leaving the default behavior
unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,11 +2,19 @@ cmake_minimum_required(VERSION 3.10)
 
 project(Tracy LANGUAGES CXX)
 
+option(TRACY_STATIC "Whether to build Tracy as a static library" NOT ${BUILD_SHARED_LIBS})
+
 find_package(Threads REQUIRED)
 
 set(TRACY_PUBLIC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/public)
 
-add_library(TracyClient "${TRACY_PUBLIC_DIR}/TracyClient.cpp")
+if(TRACY_STATIC)
+	set(TRACY_VISIBILITY "STATIC")
+else()
+	set(TRACY_VISIBILITY "SHARED")
+endif()
+
+add_library(TracyClient ${TRACY_VISIBILITY} "${TRACY_PUBLIC_DIR}/TracyClient.cpp")
 target_compile_features(TracyClient PUBLIC cxx_std_11)
 target_include_directories(TracyClient SYSTEM PUBLIC
     $<BUILD_INTERFACE:${TRACY_PUBLIC_DIR}>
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -62,7 +70,7 @@ set_option(TRACY_FIBERS "Enable fibers support" OFF)
 set_option(TRACY_NO_CRASH_HANDLER "Disable crash handling" OFF)
 set_option(TRACY_TIMER_FALLBACK "Use lower resolution timers" OFF)
 
-if(BUILD_SHARED_LIBS)
+if(NOT TRACY_STATIC)
     target_compile_definitions(TracyClient PRIVATE TRACY_EXPORTS)
     target_compile_definitions(TracyClient PUBLIC TRACY_IMPORTS)
```
davidebeatrici, line 75: Not related to the PR: I would expect this to be `INTERFACE` instead of `PUBLIC`.

https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

CMakeLists.txt
```diff
@@ -2,11 +2,19 @@ cmake_minimum_required(VERSION 3.10)
 
 project(Tracy LANGUAGES CXX)
 
+option(TRACY_STATIC "Whether to build Tracy as a static library" NOT ${BUILD_SHARED_LIBS})
```
ymd-stella, line 5: I think `NOT` is not available here.

Krzmbrzl: Why would you think that?
I tested on my system and before this change, Tracy would build as a shared lib by default, implying `BUILD_SHARED_LIBS=ON`. And after this change, Tracy still built as a shared lib by default. But if the `NOT` was ignored, I would have expected the default to change to a static lib

ymd-stella: I set BUILD_SHARED_LIBS=ON and got the following error.

```shell-session
CMake Error at 3rd/tracy/CMakeLists.txt:5 (option):
  option called with incorrect number of arguments: TRACY_STATIC Whether to
  build Tracy as a static library NOT ON
```

ymd-stella: This is a minimal example. The string `NOT` seems to be evaluated as OFF.

```
cmake_minimum_required(VERSION 3.10)
project(example)

# set(BUILD_SHARED_LIBS OFF)
# set(BUILD_SHARED_LIBS ON)
message("BUILD_SHARED_LIBS=" ${BUILD_SHARED_LIBS})
option(TRACY_STATIC "Whether to build Tracy as a static library" NOT ${BUILD_SHARED_LIBS})
message("TRACY_STATIC=" ${TRACY_STATIC})
```

Krzmbrzl: Oh interesting. So apparently `BUILD_SHARED_LIBS` is empty by default, which would explain my test results. 
Alright, I'll hand in a fix shortly. Thanks for bringing this to my attention :+1:

Krzmbrzl: https://github.com/wolfpld/tracy/pull/460

SpaceIm:
Creating a dedicated option doing the same thing as `BUILD_SHARED_LIBS` is considered bad practice (and it's convoluted).

Krzmbrzl:
It's not doing the same as `BUILD_SHARED_LIBS` though. The former has global effect, whereas the dedicated option is specific to Tracy. That makes it much easier to target a specific builld type of Tracy independent of the surrounding project's build type.

PR #467, opened 2022-09-10
Add missing TreePop in function which is called from inside a TreeNode

simplyWiri:
Alternatively this should should be moved into the sections of code which call this method.


wolfpld:
`DrawZoneList` does not know anything about a tree, so the alternative should be used.

simplyWiri:
> `DrawZoneList` does not know anything about a tree, so the alternative should be used.

https://github.com/wolfpld/tracy/blob/830e7162ebbaa4b52071d8ef8819c08399071e26/server/TracyView_FindZone.cpp#L245-L250

It currently acts as if it does. Happy to go with the alternate approach though, makes more sense to me too.

wolfpld:
Oh, ok.

PR #468, opened 2022-09-11
Windows exception handler allows other handlers to be called.

sherief:
The profiled app might install handlers to track crashes, write minidumps, etc. - this patch makes sure the app's exception handler is called when a crash happens while profiling with Tracy.


wolfpld:
How will the application be terminated now?

sherief:
It'll behave the same way it did without Tracy: if the app didn't install a handler then the OS will report the invalid operation error and terminate the process, but if the user _did_ install a handler it will run - in my case the handler was for writing a minidump, and this was causing builds in testing to not generate dumps when they crash under Tracy collection. 

This patch basically removes a behavioral change observed only when apps run with Tracy enabled. FWIW, I've been running with a version of it for a couple of years now (maybe more? It's from when this was back on BitBucket!)

PR #469, opened 2022-09-11
Initialize some variables in client

frobino:
I was trying to integrate Tracy in a project that is built using the following flags:
```
-Wno-comment -Wno-reorder -Wno-unused-variable -Wtype-limits -std=c++11 -fno-strict-aliasing -Wno-unused-but-set-variable -march=native -mfpmath=sse -mavx2 -DLV_HAVE_AVX2 -DLV_HAVE_AVX -DLV_HAVE_SSE -fvisibility=hidden -O3 -fno-trapping-math -fno-math-errno -DBUILD_TYPE_RELEASE -finstrument-functions -Werror -O3 -DNDEBUG
```

When doing that, the build procedure was failing due to issues similar to the following:
```
error: ‘<variableName>’ may be used uninitialized [-Werror=maybe-uninitialized]
```
The commit proposed in this pull request solves the issue by initializing some variables in the client.

It is possible to reproduce the issue by setting the flags above in the tracy/test Makefile, and building the test.

### DIFF
--- a/public/client/TracyCallstack.cpp
+++ b/public/client/TracyCallstack.cpp
@@ -661,7 +661,7 @@ static void InitKernelSymbols()
         while( *ptr != '\t' && *ptr != '\n' ) ptr++;
         const auto nameend = ptr;
         const char* modstart = nullptr;
-        const char* modend;
+        const char* modend = nullptr;
         if( *ptr == '\t' )
         {
             ptr += 2;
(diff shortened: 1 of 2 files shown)


wolfpld:
The "errors" you have found are not real problems. There are no paths possible in the program for the issues to manifest. Your static analyser is buggy and you have specifically asked for this behavior to be enabled.

On the other hand, the initialization you are asking to add has a very real impact on the code.

I will not be merging this, sorry.

PR #490, opened 2022-11-03
Fix compilation error on linux when TRACY_NO_CRASH_HANDLER defined

nwessing:
On Linux, when `TRACY_NO_CRASH_HANDLER` is defined, there is a compilation error here: https://github.com/threespace/tracy/blob/fix/compile_on_linux_without_crash_handlers/public/client/TracyProfiler.cpp#L1460 due to `TRACY_CRASH_SIGNAL` not being defined. This PR just makes sure that `TRACY_CRASH_SIGNAL` is defined no matter what.


wolfpld:
Resolved in a more correct way in e7ac54fb.

nwessing:
Thanks!

PR #492, opened 2022-11-09
Fix build when TRACY_HAS_CALLSTACK is not defined

veghbalazs:
The function `tracy::Callstack()` is unknown when `TRACY_HAS_CALLSTACK` macro is not defined, causing compile errors in platforms without callstack support (e.g. Universal Windows Platform in our case). Added guards around affected methods.

Another working alternative could be to introduce a dummy implementation in `TracyCallstack.hpp`: 
```cpp
#ifndef TRACY_HAS_CALLSTACK
namespace tracy
{
static tracy_force_inline void* Callstack( int depth )
{
    return nullptr;
}
}
#else 
// ...
```
Guarding seemed to be the more common practice in the codebase, please tell if you prefer the dummy option or any other alternative.

### DIFF
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -4172,6 +4172,7 @@ TRACY_API void ___tracy_emit_gpu_zone_begin( const struct ___tracy_gpu_zone_begi
     TracyLfqCommitC;
 }
 
+#ifdef TRACY_HAS_CALLSTACK
 TRACY_API void ___tracy_emit_gpu_zone_begin_callstack( const struct ___tracy_gpu_zone_begin_callstack_data data )
 {
     tracy::GetProfiler().SendCallstack( data.depth );
@@ -4183,6 +4184,7 @@ TRACY_API void ___tracy_emit_gpu_zone_begin_callstack( const struct ___tracy_gpu
     tracy::MemWrite( &item->gpuZoneBegin.srcloc, data.srcloc );
     TracyLfqCommitC;
 }
+#endif
 
 TRACY_API void ___tracy_emit_gpu_zone_begin_alloc( const struct ___tracy_gpu_zone_begin_data data )
 {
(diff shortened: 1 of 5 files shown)


wolfpld:
The intention is for code to be portable between platforms, so that if you use callstack macros or functions on a platform that doesn't support callstack capture, the functionality should behave as if non-callstack macros/functions were used. There should be redirection to callstack-less code paths in the `#else` branches for each `#ifdef` you have added. Changing the `Callstack()` function behavior might be a viable alternative, but it would require a review of all the places where it is used.

veghbalazs:
Got it. Compared to adding new `#else` branches, reviewing `tracy::Callstack()` looks like the simpler option. When not behind an existing `TRACY_HAS_CALLSTACK` check, it's always used as an input to [`Profiler::QueueSerialCallstack()`](https://github.com/wolfpld/tracy/blob/c0035a492f2af2f6ceb435b2f824e1cc43e1e00f/public/client/TracyProfiler.hpp#L246-L252). It passes on the pointer to [`Profiler::SendCallstackSerial()`](https://github.com/wolfpld/tracy/blob/c0035a492f2af2f6ceb435b2f824e1cc43e1e00f/public/client/TracyProfiler.hpp#L835), which falls back to noop when `TRACY_HAS_CALLSTACK` is not defined.

Therefore I think the dummy `tracy::Callstack()` implementation is safe, I'd move forward with that. Wdyt?

Call sites of `tracy::Callstack()` are already well guarded against actually getting executed when that macro is off. For example a client shouldn't use any `tracy::D3D11ZoneScope` constructor directly, there's `TracyD3D11ZoneS()` macro for the callstack version, which is redirected to callstack-less `TracyD3D11Zone()` on platforms lacking support. But, the callstack version of `tracy::D3D11ZoneScope` constructor is still part of the compilation, hence the build errors.

wolfpld:
2c033063 should fix this for you.

veghbalazs:
Great, thanks!

Issue #28, opened 2020-05-22
Custom source location macros


Manu343726:
Hi,

One of the changes I have to do to integrate tracy in my company was to improve the client `__FUNCTION__` based source location by `constexpr`-processing `__PRETTY_FUNCTION__` to get full qualified function names. The issue is that this introduces a change that I have to manually patch whenever I want to update the profiler server to get the latest features and fixes.

I was thinking that maybe tracy could open the door for a bit of customization of source location by allowing users to provide their own alternatives:

``` cpp

constexpr std::string_view ParseFunctionName(std::string_view funcname)
{
    return ...;
}

#define TRAYC_SOURCE_LOCATION_FUNCTION \
    ParseFunctionName(__PRETTY_FUNCTION__).data()

#include <Tracy.hpp>
```

and tracy uses `__FUNCTION__` by default (To keep the current behavior as default):

``` cpp
#ifndef TRACY_SOURCE_LOCATION_FUNCTION
    #define TRACY_SOURCE_LOCATION_FUNCTION __FUNCTION__
#endif

...

#define ZoneNamed( varname, active ) static const tracy::SourceLocationData TracyConcat(__tracy_source_location,__LINE__) { nullptr, TRACY_SOURCE_LOCATION_FUNCTION, ...)
```

The same could apply to the file and line info. With full customization users could decide using C++20 [`std::source_location`](https://en.cppreference.com/w/cpp/utility/source_location) (Or similar lib feature) if they want:

``` cpp
#define TRACY_SOURCE_LOCATION_FILE std::source_location::current().file_name()
#define TRACY_SOURCE_LOCATION_LINE std::source_location::current().line()
#define TRACY_SOURCE_LOCATION_FUNCTION std::source_location::current().function_name()
```

Would you be interested in a PR implementing this feature?
Thanks.

wolfpld:
Can you give some more information about your intended use case? I.e. what do you get by default and what your solution provides instead?

Manu343726:
Of course. Given these two functions:

``` cpp
namespace module1
{
    void f();

    class Entity
    {
        void f();
    };
}
```

`__FUNCTION__` gives `"f"` in both cases, but `__PRETTY_FUNCTION__` gives `"module1::f"` and `"module1::Entity::f"`. It also handles overloads (Since `__PRETTY_FUNCTION__` includes the full function signature).

wolfpld:
[...]

PR #497, opened 2022-11-25
More detailed function names under Clang and GCC

pshurgal:
# Why?
`__FUNCTION__` produces different function names under MSVC and Clang/GCC. Example:
```
struct EventSystem {
    void execute() {
        ZoneScoped;
        // ...
    }
};

struct GameplaySystem {
    void execute() {
        ZoneScoped;
        // ...
    }
};

struct RenderSystem {
    void execute() {
        ZoneScoped;
        // ...
    }
};

int main() {
    EventSystem es;
    GameplaySystem gs;
    RenderSystem rs;

    while(true) {
        es.execute();
        gs.execute();
        rs.execute();
    }
    return 0;
}
```

If the binary was built under MSVC in Tracy Profiler I would see zones named: `EventSystem::execute`, `GameplaySystem::execute`, `RenderSystem::execute`.
If the binary was built under GCC or Clang in Tracy Profiler I would see zones named: `execute`, `execute`, `execute`.

The PR fixes the broblem and makes source location info more informative.

### DIFF
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -801,6 +801,28 @@ Remember to set thread names for proper identification of threads. You should do
 
 Tracy will try to capture thread names through operating system data if context switch capture is active. However, this is only a fallback mechanism, and it shouldn't be relied upon.
 
+\subsubsection{Source location data customization}
+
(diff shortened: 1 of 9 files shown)


wolfpld:
Please see #28 and maybe #259.

pshurgal:
Okay, so what is you proposal? I think I can make https://github.com/wolfpld/tracy/issues/28 up to date, for example

wolfpld:
> I think I can make https://github.com/wolfpld/tracy/issues/28 up to date, for example

Give it a go.

pshurgal:
@wolfpld please have a look
I'm not sure if documentation is in correct place, what do you think about it?

wolfpld:
CI seems to be back to normal now, the manual needs fixing.

pshurgal:
Done

PR #499, opened 2022-11-28
Always fallback to using native thread name...

ReplayCoding:
...even when TRACY_ENABLE is defined.

### DIFF
--- a/public/common/TracySystem.cpp
+++ b/public/common/TracySystem.cpp
@@ -205,61 +205,57 @@ TRACY_API const char* GetThreadName( uint32_t id )
         }
         ptr = ptr->next;
     }
-#else
-#  if defined _WIN32
-#    ifdef TRACY_UWP
-    static auto _GetThreadDescription = &::GetThreadDescription;
-#    else
-    static auto _GetThreadDescription = (t_GetThreadDescription)GetProcAddress( GetModuleHandleA( "kernel32.dll" ), "GetThreadDescription" );
-#    endif
-    if( _GetThreadDescription )
-    {
-        auto hnd = OpenThread( THREAD_QUERY_LIMITED_INFORMATION, FALSE, (DWORD)id );
-        if( hnd != 0 )
-        {
-            PWSTR tmp;
-            _GetThreadDescription( hnd, &tmp );
-            auto ret = wcstombs( buf, tmp, 256 );
-            CloseHandle( hnd );
-            if( ret != 0 )
-            {
-                return buf;
-            }
-        }
-    }
-#  elif defined __linux__
-    int cs, fd;
-    char path[32];
-#   ifdef __ANDROID__
-    int tid = gettid();
-#   else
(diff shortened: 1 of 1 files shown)


wolfpld:
I will need some sort of explanation here.

ReplayCoding:
Previously, if `TRACY_ENABLE` was defined, `GetThreadName` would return the thread id as a string unless it was set using `SetThreadName`. While this is fine in cases where we can set the thread names ourselves, this becomes an issue if we can't set them.

In my case, thread names are set before my code is loaded, and I don't have access to the code that sets the names.

wolfpld:
I still don't understand what this is trying to solve. Please provide a high-level description of the problem you are having, the required conditions (since you state that your case is non-trivial), the results you get, the results you wish for, how the code change will fix the issue, etc.

ReplayCoding:
I'm writing a game mod that (among other things) hooks RAD Telemetry to use Tracy. Unfortunately, the mod is loaded after most of the thread names are setup. The game does set the system thread names, but as of the current commit on master, Tracy won't use them:
The thread names are falling back to showing the TID

Ideally it should use the OS thread name if Tracy isn't given a name using `SetThreadName`, which this PR implements:
The thread names are using the OS thread name

### Implementation

Currently (on master), `GetThreadName` functions like this:
```cpp
TRACY_API const char* GetThreadName( uint32_t id )
{
#ifdef TRACY_ENABLE
    auto ptr = GetThreadNameData().load( std::memory_order_relaxed );
    while( ptr )
    {
        if( ptr->id == id )
        {
            return ptr->name;
        }
        ptr = ptr->next;
    }
#else
    // Get OS thread name.
#endif
    // Fallback if we don't get a name from GetThreadNameData or OS.
    sprintf( buf, "%" PRIu32, id );
    return buf;
}
```
This PR changes it to function like this:
```cpp
TRACY_API const char* GetThreadName( uint32_t id )
{
    static char buf[256];
#ifdef TRACY_ENABLE
    auto ptr = GetThreadNameData().load( std::memory_order_relaxed );
    while( ptr )
    {
        if( ptr->id == id )
        {
            return ptr->name;
        }
        ptr = ptr->next;
    }
#endif

  // There was no name stored in ThreadNameData. Get OS thread name.

  // We couldn't get an OS thread name, fallback to returning TID
  sprintf( buf, "%" PRIu32, id );
  return buf;
}
```

public/common/TracySystem.cpp
```diff
@@ -205,61 +205,57 @@ TRACY_API const char* GetThreadName( uint32_t id )
...
+          if( ret != 0 )
+          {
+              return buf;
+          }
+      }
+  }
+#elif defined __linux__
+  int cs, fd;
+  char path[32];
+  snprintf( path, sizeof( path ), "/proc/self/task/%d/comm", id );
```
wolfpld, line 234: I wonder if this was ever working before.

Issue #286, opened 2021-11-23
Poor scrolling behavior with many threads


adepke:
I have a trace that contains a very large amount of threads, and attempting to zoom onto zones far down in the thread list causes threads up above to leave the viewing space, shifting everything extremely far up. [Here's a trace](https://github.com/wolfpld/tracy/files/7585103/LotsOfZones.tracy.txt) producing this behavior.

https://user-images.githubusercontent.com/18013792/142952480-0a502491-d5b6-4f0b-807f-f2fc64fe66c8.m

wolfpld:
It is a known issue. Sorry, no easy fix for this right now (I tried).

wolfpld:
Fixed by #519.

PR #503, opened 2022-12-18
Timeline animation tweaks

tvoeroes:
* Changed the resizing logic such that when one `TimelineItem` is shrinking the `TimelineItem`s below it are following at the same speed.
* I also sneaked in a `TimelineController` height calculation fix.
* Reduced the number of lines of code.
* Tested @ 60FPS and 15FPS.
* Removed `TimelineItem::AdjustThreadPosition()` because it became obsolete.

Why:

* With this change, the `TimelineItem` shrinking is more consistent with `TimelineItem` expanding.
* Without the change each `TimelineItem` is moving at a different speed across the screen when resizing. This makes it much harder for the brain to keep track of what is happening.
* Without the change, the animations can take a very long time if a many `TimelineItem`s are resizing at the same time. To the point where in one of my use cases I had to wait multiple seconds for the animation to finish before I could reorient myself.
* With the change the timeline seems more usable to me.

### DIFF
--- a/server/TracyTimelineController.cpp
+++ b/server/TracyTimelineController.cpp
@@ -7,6 +7,7 @@ namespace tracy
 
 TimelineController::TimelineController( View& view, Worker& worker )
     : m_height( 0 )
+    , m_scroll( 0 )
     , m_firstFrame( true )
     , m_view( view )
     , m_worker( worker )
@@ -30,7 +31,12 @@ void TimelineController::End( double pxns, int offset, const ImVec2& wpos, bool 
         item->Draw( m_firstFrame, pxns, offset, wpos, hover, yMin, yMax );
     }
 
-    m_height = offset;
+    const auto scrollPos = ImGui::GetScrollY();
+    if( ( scrollPos == 0 && m_scroll != 0 ) || offset > m_height )
+    {
+        m_height = offset;
+    }
+    m_scroll = scrollPos;
 }
 
 }
--- a/server/TracyTimelineController.cpp
+++ b/server/TracyTimelineController.cpp
@@ -7,7 +7,6 @@ namespace tracy
 
 TimelineController::TimelineController( View& view, Worker& worker )
     : m_height( 0 )
-    , m_scroll( 0 )
     , m_firstFrame( true )
     , m_view( view )
     , m_worker( worker )
@@ -31,16 +30,7 @@ void TimelineController::End( double pxns, int offset, const ImVec2& wpos, bool 
         item->Draw( m_firstFrame, pxns, offset, wpos, hover, yMin, yMax );
     }
 
-    const auto scrollPos = ImGui::GetScrollY();
-    if( scrollPos == 0 && m_scroll != 0 )
-    {
-        m_height = 0;
-    }
-    else
-    {
-        if( offset > m_height ) m_height = offset;
-    }
-    m_scroll = scrollPos;
+    m_height = offset;
 }
 
 }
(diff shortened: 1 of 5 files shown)


server/TracyTimelineController.cpp
```diff
@@ -31,16 +30,7 @@ void TimelineController::End( double pxns, int offset, const ImVec2& wpos, bool
...
-    if( scrollPos == 0 && m_scroll != 0 )
-    {
-        m_height = 0;
-    }
-    else
-    {
-        if( offset > m_height ) m_height = offset;
-    }
-    m_scroll = scrollPos;
+    m_height = offset;
```
wolfpld, line 33: Why was this changed?

server/TracyTimelineItem.cpp
```diff
@@ -22,16 +21,16 @@ void TimelineItem::Draw( bool firstFrame, double pxns, int& offset, const ImVec2
 {
     if( !IsVisible() )
     {
-        m_height = 0;
-        m_offset = 0;
+        if( m_height != 0 )
+            AdjustThreadHeight( firstFrame, offset, offset );
```
wolfpld, line 25: Please make a one-liner here.

server/TracyTimelineController.cpp
tvoeroes: The line `m_height = 0;` was causing a one-frame long glitch of the vertical scroll bar. I see that this is controversal because the change removes the ability to scroll vertically past the last plot. Alternatively I propose:

```
    if( ImGui::GetScrollY() == 0 )
    {
        m_height = offset;
    }
    else
    {
        if( offset > m_height ) m_height = offset;
    }
```

Which fixes the glitch without removing the feature. Or even something like this (my favorite):

```
m_height = offset + int( ImGui::GetWindowHeight() * 0.75f );
```

Which would always allow scrolling past the end even when no `TimelineItem`s were shrinked.

tvoeroes: Simplified:

```
if( ImGui::GetScrollY() == 0 || offset > m_height ) m_height = offset;
```

tvoeroes: Two points that I just realized: My original change and the second alternative (`m_height = offset + int( ImGui::GetWindowHeight() * 0.75f );`) would also prevent vertically scrolling beyond where the plots end.

tvoeroes: I now went with
```
if( ImGui::GetScrollY() == 0 || offset > m_height ) m_height = offset;
```
for this pull request because the other options behave a little funky when scrolling horizontally while all the way on the bottom of the timeline.

wolfpld: The reason for this behavior is to prevent unwanted vertical view position changes, when inspecting bottom items on the timeline. In a live trace (or even when panning offline) it is possible there will be very little items to show, which would forcefully move the view to the top (as by default you can't "scroll vertically past last item").

Having all the extended vertical scroll space is not always wanted, so you can reset it by scrolling to the top, which produces the momentary glitch you have observed.

Please review that your changes do not change this behavior.

FWIW this is really a stop-gap solution, as ideally the position of the thread you are viewing should be fixed on the screen during panning. Currently it's not the case, as threads will jump up and down, depending on what is displayed above, as shown in #286. The solution here looks to be non-trivial as you need to at least include the mouse position into the calculation, to determine which thread should be the anchor point. Maybe this would be something you would want to try to tackle, considering you are already familiar with the code?

tvoeroes: I have now completely reverted the reset logic except for resetting to the correct height instead of to 0. So the only change in this part of the code is that there is no more one-frame glitch.

The sample save file from #286 is much easier to navigate with the resizing changes but the issue that when `TimelineItem`s collapse (e.g. when zooming in) you have to manually scroll up remains. Simply setting `m_height = offset;` every frame would resolve that issue as well but as I now understand this change is not desirable and there needs to be a better solution for this.

I also looked in to how much effort it would be to rework such that the center remains in view when `TimelineItem`s are resized vertically. I'll see if I get around to doing it. It is something that I would like to have.

PR #510, opened 2023-01-18
Miscellaneous emscripten fixes

topolarity:
Series of emscripten-specific fixes:
- Increase `STACK_SIZE` to 128 kB from 64 kB (was causing crashes, especially on startup)
- Re-scale horizontal scroll values so that scroll speed (approximately) matches desktop
- Fix NULL pointer dereference
- Increase `PTHREAD_POOL_SIZE` from 4 to 8 (may be unnecessary)

### DIFF
--- a/profiler/build/wasm/build.mk
+++ b/profiler/build/wasm/build.mk
@@ -5,7 +5,7 @@ CFLAGS += -sUSE_FREETYPE=1 -pthread
 CXXFLAGS := $(CFLAGS) -std=c++17
 DEFINES += -DIMGUI_ENABLE_FREETYPE -DIMGUI_IMPL_OPENGL_ES2
 INCLUDES := -I../../../imgui -I$(HOME)/.emscripten_cache/sysroot/include/capstone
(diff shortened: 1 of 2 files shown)


wolfpld:
If the pthread pool size change may be unnecessary, why is it made?

topolarity:
> If the pthread pool size change may be unnecessary, why is it made?

The thread pool is definitely being exhausted, but I don't know enough about Tracy's usage of threads to know when it depends on being able to execute all its threads concurrently.

Happy to remove the change if you'd prefer.

> `imgui_impl_glfw.cpp` is sourced from ImGui. Please submit your changes to ImGui first. I don't want to maintain yet another patch on foreign sources.

Fair enough - I've submitted https://github.com/ocornut/imgui/pull/6096

wolfpld:
> The thread pool is definitely being exhausted, but I don't know enough about Tracy's usage of threads to know when it depends on being able to execute all its threads concurrently.
> 
> Happy to remove the change if you'd prefer.

I have seen some weird problems related to threads, but these were mainly on Chrome-based browsers. Firefox seems to handle this much better.

There are already some ifdefs to minimize thread usage with emscripten, but maybe this is not enough?

https://github.com/wolfpld/tracy/blob/e4bd88c/server/TracyWorker.cpp#L1271-L1277

topolarity:
Good point. I'll try to investigate this weekend and see I can track down where the threads are being started

wolfpld:
I believe most of this PR is either already applied or obsolete.

PR #514, opened 2023-01-22
Fix color red and blue byte swap

jeslaspravin:
I noticed this strange bug. When sending a message with the color `Red` to the profiler(for example using `TracyMessageLC`) the color being sent is displayed in Profiler such that the R byte gets swapped with the G byte and will be displayed as `Blue`

Please note that I have tested these changes only on my windows machine. I do not have other platforms to test this on.

Below is the screenshot of `QueueMessageColor` that is being enqueued from `TracyClient`
image

Below is the screenshot of the `Profiler in master`
image

Below is the screenshot of the `Profiler in Fixed PR`
image

### Change in functions
* Worker::AddSourceLocation
* Worker::ProcessZoneColor
* Worker::ProcessMessageColor
* Worker::ProcessMessageLiteralColor

### DIFF
--- a/server/TracyWorker.cpp
+++ b/server/TracyWorker.cpp
@@ -3725,7 +3725,7 @@ void Worker::AddSourceLocation( const QueueSourceLocation& srcloc )
         }
     }
     CheckString( srcloc.function );
-    const uint32_t color = ( srcloc.r << 16 ) | ( srcloc.g << 8 ) | srcloc.b;
+    const uint32_t color = srcloc.r | ( srcloc.g << 8 ) | ( srcloc.b << 16 );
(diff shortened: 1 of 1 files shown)


jeslaspravin:
Issue [https://github.com/wolfpld/tracy/issues/515](https://github.com/wolfpld/tracy/issues/515)

wolfpld:
Your solution is incorrect, as it assumes the protocol field names are the reference. They are not. The reference is in the manual, which states that it is required to enter the colors in `0xRRGGBB` format. What happens later has little importance, as long as the colors displayed in the profiler match what the user has entered.

Mislabeled protocol color field are fixed in eb7c13e, 6652999, d471225.

jeslaspravin:
> 

Yep you are right. I see that `RBGA` packing is used in ImGUI as well. Thanks for the fix!

PR #516, opened 2023-01-25
Fix connection check for capturing

pzread:
The tracy worker `Exec` resets the `IsConnected` to false when finishing the capture. If it finishes too quick, the waiting loop in `capture.cpp` might not ever see `IsConnected = true`.

This changes the waiting condition to use `HasData`, which stays as true once the worker receives the data.


wolfpld:
Hi, please rewrite the commit message to include the reasoning you have included in this PR. Thanks!

pzread:
Done.

PR #521, opened 2023-02-08
Fix MSVC compiler warning

john-plate:
The existing way to disable MSVC warning 4554 is not a good solution as the warning will still be triggered whenever the template is instantiated. It is better to change the code in a way that doesn't trigger the (false) warning at all.

### DIFF
--- a/public/client/tracy_concurrentqueue.h
+++ b/public/client/tracy_concurrentqueue.h
@@ -210,19 +210,13 @@ namespace details
 		}
 	};
 
-#ifdef _MSC_VER
-#pragma warning(push)
-#pragma warning(disable: 4554)
-#endif
 	template<typename T>
 	static inline bool circular_less_than(T a, T b)
 	{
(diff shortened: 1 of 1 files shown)


wolfpld:
Please submit this fix to the upstream repo (and link it here). Thanks.

john-plate:
I'm not aware of an upstream repo and can't find it. Where is it?

john-plate:
It's already fixed upstream in a different way:
https://github.com/cameron314/concurrentqueue/commit/9dc1b2cfcad03b4ee22ea57ddb5c453c41c19ac9
Would be great to see this fix in Tracy!

wolfpld:
Can you cherry-pick the fix and submit a PR?

john-plate:
Sure:
https://github.com/wolfpld/tracy/pull/522

wolfpld:
Thanks!

PR #523, opened 2023-02-10
Add missing include in TracyD3D11.hpp

palver123:
This header used many D3D11 interfaces but didn't include the header defining them. This caused compiler errors on Windows.

### DIFF
--- a/public/tracy/TracyD3D11.hpp
+++ b/public/tracy/TracyD3D11.hpp
@@ -42,6 +42,8 @@ using TracyD3D11Ctx = void*;
 #include "../common/TracyAlign.hpp"
 #include "../common/TracyAlloc.hpp"
 
+#include <d3d11.h>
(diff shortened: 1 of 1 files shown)


wolfpld:
Generally it is expected to include the required gfx API headers before including Tracy headers. OpenGL and in lesser part Vulkan typically use various loaders which may require custom approach to header inclusion. However, in case of D3D11 the environment is homogenous, so it should be fine to just include the header in this place.

palver123:
> Generally it is expected to include the required gfx API headers before including Tracy headers. 

Hmm I was suspecting this, but then I looked at TracyD3D12.hpp where d3d12.h is included. Plus including d3d11.h before Tracy headers, will pull in windows.h too early causing the usual NOMINMAX hassle, while including it the way I did avoids that

PR #530, opened 2023-02-27
Fix unstable zone grouping

Lectem:
The zone grouping was only taking into account the first zone for both its duration and width, which was wrong and could cause zones to "disappear" if the first zone of the group was really small. 

image
image

Grouping now works by taking into account the size of the zone and not the clipped to screen size for grouping decisions, to avoid having zone groups popping at the beginning and end of the window. In order to have consistent results and avoid popping, we're using the start of the last zone in a group as reference instead of its end. This means that it now includes the zone width, so that we make the same decisions wether we are in considering new zones in the loop or to start grouping zones.

Altogether, this seems to have fixed any zone popping when panning, at the cost of not grouping the first/last zone if it is clipped, which seems to be less of an issue to me.

### DIFF
--- a/server/TracyView_ZoneTimeline.cpp
+++ b/server/TracyView_ZoneTimeline.cpp
@@ -636,7 +636,7 @@ int View::DrawZoneLevel( const V& vec, bool hover, double pxns, int64_t nspx, co
     {
         auto& ev = a(*it);
         const auto end = m_worker.GetZoneEnd( ev );
-        const auto zsz = std::max( (end - ev.Start()) * pxns, pxns * 0.5 );
+        const auto zsz = std::max( ( end - ev.Start() ) * pxns, pxns * 0.5 );
         if( zsz < MinVisSize )
         {
             const auto MinVisNs = MinVisSize * nspx;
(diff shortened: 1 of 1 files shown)


wolfpld:
I made some changes to prevent popup at the start of timeline viewport in f428a5b5. This does not yet fix the core problem here.

wolfpld:
The logic for determining what should be drawn has been decoupled from the drawing process itself. The code now lives here:

https://github.com/wolfpld/tracy/blob/1e1833edc2d0e852f538246fbe8f4b10af6bd680/server/TracyTimelineItemThread.cpp#L299-L432.

wolfpld:
I have basically rewritten the grouping code. Please see if the following code makes sense to you and if it solves your problem. It should be much easier to follow now that all calculations are in the time domain.

```c++
const auto MinVisNs = int64_t( round( GetScale() * MinVisSize * nspx ) );

...

while( it < zitend )
{
    auto& ev = a(*it);
    const auto end = m_worker.GetZoneEnd( ev );
    const auto zsz = end - ev.Start();
    if( zsz < MinVisNs )
    {
        auto nextTime = end + MinVisNs;
        auto next = it + 1;
        for(;;)
        {
            next = std::lower_bound( next, zitend, nextTime, [] ( const auto& l, const auto& r ) { Adapter a; return (uint64_t)a(l).End() < (uint64_t)r; } );
            if( next == zitend ) break;
            auto prev = next - 1;
            if( prev == it ) break;
            const auto pt = m_worker.GetZoneEnd( a(*prev) );
            const auto nt = m_worker.GetZoneEnd( a(*next) );
            if( nt - pt >= MinVisNs ) break;
            nextTime = nt + MinVisNs;
        }
        m_draw.emplace_back( TimelineDraw { TimelineDrawType::Folded, uint16_t( depth ), (void**)&ev, m_worker.GetZoneEnd( a(*(next-1)) ), uint32_t( next - it ) } );
        it = next;
    }
    else
    {
        ...
    }
}
```

Lectem:
The issue that I had seems to be fixed with your modifications:
image

wolfpld:
Thanks. If you can, please also review the new code to see if what it does makes sense to you.

PR #532, opened 2023-03-01
Add: Alternative Vulkan context constructor

spnda:
This adds a new constructor to the Vulkan context class that does not require to have a VkCommandBuffer and VkCommandPool created for it. It instead fully relies on host operations that were added with newer versions of Vulkan to completely avoid the use of command buffers. However, there are cases where the physical device does not support host timestamps with `vkGetCalibratedTimestampsEXT` where the command buffer based constructor would need to be used (I've placed an assertion to detect this). The user would need to check for this. (Most modern GPU drivers support this though)

I've tried to stick with the style of the rest of the header, I hope that that looks fine. However, I'm not sure wether I should be moving common code out of the constructors so that their differences are easily spottable and are simpler to read.


wolfpld:
Please adapt your changes to the code style.

PR #541, opened 2023-03-23
add version to meson.build 'project()'

kklobe:
Meson projects which use tracy as a subproject and specify a version for the dependency will fail due to an undefined version number.

A new tag would be helpful as well so I can submit a new version to the Meson WrapDB, maybe '0.9.1.1'.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('tracy', ['cpp'])
+project('tracy', ['cpp'], version: '0.9.1')
 
 if get_option('tracy_enable')
   add_project_arguments('-DTRACY_ENABLE', language : 'cpp')


kklobe:
without it, it breaks our [project](https://github.com/dosbox-staging/dosbox-staging), since we specify the version:
```meson
# Tracy
tracy_dep = optional_dep
if get_option('tracy')
    tracy_dep = dependency(
        'tracy',
        version: ['>= 0.9', '< 1'],
        ...
```
I'm willing to help with any solution that you'd be happy with, even something simple like `scripts/release.sh`

wolfpld:
There is a sort-of release script at `extra/make-build.sh`. Maybe you can extend it to parse `public/common/TracyVersion.hpp` and update `meson.build`?

kklobe:
Let me know what you think of the proposed changes.

kklobe:
If you could cut a '0.9.1.1' tag, I can get the Meson WrapDB updated - thanks.

wolfpld:
There are some pretty big changes going on in the repository right now, and things may not be stable. I'd rather not tag anything at the moment, unless it's for client-only use.

kklobe:
Sounds good, I'll add a manual patch on the WrapDB side for now.

PR #545, opened 2023-04-04
linux: respect `TRACY_NO_SAMPLING` for sys-tracing

topolarity:
This compile-time flag was being ignored on Linux.

This change adds gating for software-sampled stack trace sampling following the same pattern as the other "TRACY_NO_SAMPLE_*" options. In particular, if `TRACY_NO_SAMPLING=1` is provided as an environment variable, software stack sampling is also disabled.

### DIFF
--- a/public/client/TracySysTrace.cpp
+++ b/public/client/TracySysTrace.cpp
@@ -770,6 +770,13 @@ bool SysTraceStart( int64_t& samplingPeriod )
     TracyDebug( "sched_wakeup id: %i\n", wakeupId );
     TracyDebug( "drm_vblank_event id: %i\n", vsyncId );
 
+#ifdef TRACY_NO_SAMPLING
+    const bool noSoftwareSampling = true;
+#else
+    const char* noSoftwareSamplingEnv = GetEnvVar( "TRACY_NO_SAMPLING" );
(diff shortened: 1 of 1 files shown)


public/client/TracySysTrace.cpp
```diff
@@ -839,28 +846,31 @@ bool SysTraceStart( int64_t& samplingPeriod )
     pe.clockid = CLOCK_MONOTONIC_RAW;
 #endif
 
-    TracyDebug( "Setup software sampling\n" );
-    ProbePreciseIp( pe, currentPid );
-    for( int i=0; i<s_numCpus; i++ )
+    if ( !noSoftwareSampling )
```
wolfpld, line 849: No space between `if` and `(`.

topolarity:
What would you like updated? The manual already says:
> Call stack sampling may be disabled by using the TRACY_NO_SAMPLING define.

wolfpld:
> What would you like updated? The manual already says:

I was thinking about the following wording:

> If you want to restrict it to only listening on the localhost interface, define the TRACY_ONLY_LOCALHOST macro at compile-time, or set the TRACY_ONLY_LOCALHOST environment variable to 1 at runtime.

But I see it is also missing for the other macros that already support reading env vars.

PR #555, opened 2023-04-19
Added an option to print out "messages" while exporting a .trace file to CSV using "csvexport" utility

nehasharan:


csvexport/src/csvexport.cpp
```diff
@@ -23,12 +23,13 @@ void print_usage_exit(int e)
...
-    fprintf(stderr, "  -c, --case        Case sensitive filtering\n");
-    fprintf(stderr, "  -e, --self        Get self times\n");
-    fprintf(stderr, "  -u, --unwrap      Report each zone event\n");
+    fprintf(stderr, "  -h, --help                       Print usage\n");
+    fprintf(stderr, "  -f, --filter arg                 Filter zone names (default: "")\n");
+    fprintf(stderr, "  -s, --sep arg                    CSV separator (default: ,)\n");
+    fprintf(stderr, "  -c, --case                       Case sensitive filtering\n");
+    fprintf(stderr, "  -e, --self                       Get self times\n");
+    fprintf(stderr, "  -u, --unwrap                     Report each zone event\n");
+    fprintf(stderr, "  -m, --unwrap only messages       Report only messages\n");
```
wolfpld, line 32: This does not correspond with what the program expects.

nehasharan: Hi @wolfpld there is currently no way to print messages which we add to zones.

fosterbrereton: @wolfpld would this be more acceptable?
```suggestion
    fprintf(stderr, "  -m, --messages                   Report only messages\n");
```

wolfpld: Yes, but the additional padding added between arguments and their descriptions is no longer needed.

csvexport/src/csvexport.cpp
```diff
@@ -198,6 +204,38 @@ int main(int argc, char** argv)
...
+                std::string row = join(values, args.separator);
+                printf("%s\n", row.data());
+            }
+        }
+        else
+        {
+            printf("There are currently no messages!\n");
+        }
+    
+    return 0;
```
wolfpld, line 236: Align return with indentation.

nehasharan: @wolfpld done.

PR #557, opened 2023-04-28
allow filtering zones in Zone finder based on the user text

KristofferC:
In the Julia language we integrate with Tracy and record when functions are compiled. As the user text for that zone we put the signature of the function getting compiled.

It would thus then be useful to be able to filter on the user text when viewing zones in e.g. the Find Zone window to see e.h. the compile time of a certain class of functions.

This PR adds a filter box that can be used to do such filtering. A possible use case is shown below:

Screenshot 2023-04-28 at 11 23 42

A possible question is if one should be able to filter on the other things you can group by (Thread, Parent, etc) and in that case how the UI should look for that.


server/TracyView_FindZone.cpp
```diff
@@ -1362,6 +1362,23 @@ void View::DrawFindZone()
...
+        ImGui::SameLine();
+        if( ImGui::Button( ICON_FA_DELETE_LEFT " Clear" ) )
+        {
+            m_userTextFilter.Clear();
+            filterChanged = true;
+        }
+        ImGui::Separator();
+        if( filterChanged )
+        {
+            m_findZone.ResetGroups();
```
KristofferC, line 1379: I'm not sure if this is the correct way to "dirty" the zone list to cause it to redraw but this seems to work at least.

server/TracyView_FindZone.cpp
```diff
@@ -1362,6 +1362,23 @@ void View::DrawFindZone()
         SmallCheckbox( "Show zone time in frames", &m_findZone.showZoneInFrames );
         ImGui::Separator();
 
+        ImGui::AlignTextToFramePadding();
+        TextDisabledUnformatted( "Filter user text:" );
+        ImGui::SameLine();
+        bool filterChanged = m_userTextFilter.Draw( ICON_FA_FILTER "###resultFilter", 200 );
+
+        ImGui::SameLine();
+        if( ImGui::Button( ICON_FA_DELETE_LEFT " Clear" ) )
```
wolfpld, line 1371: This button ID clashes with one that is already present. Can be fixed by appending e.g. `##userText` after the button text.

wolfpld:
Here I have selected zones with user text "16". The blue area on the histogram represents this selection.

obraz

If I now use the functionality you added to filter out any zones that don't have user text "16", I get a proper number of zones in other grouping modes, such as by thread. However, if I select such a filtered group, the filtering does not apply to the histogram.

obraz

KristofferC:
I tried to fix the histogram by  keeping some state of what got filtered to the next frame where the histogram reads it and uses it to not consider filtered zones as selected:

Screenshot 2023-05-01 at 13 12 20

I'm a bit uncertain if the use of a `set` here is the best way of going about this.

wolfpld:
Please adapt spacing in function calls to the coding standard.

> I'm a bit uncertain if the use of a `set` here is the best way of going about this.

It doesn't feel nice, but I don't have a better idea.

KristofferC:
> Please adapt spacing in function calls to the coding standard.

Should be fixed. I didn't rebase the commits since I figured it is easy enough to just squash everything together on an eventual merge.

PR #569, opened 2023-06-03
Allow for dynamic symbol loading for vulkan profiling

robertblaketaylor:
Code assumes that application will static link the loader, which is not true for all cases (ex. vulkan apps that load the symbols into a symbol table to skip the underlying trampoline).

Added TRACY_VK_USE_SYMBOL_TABLE to have Tracy populate a symbol table directly.


robertblaketaylor:
Merge error, was on latest release, will re-submit the pull request

PR #570, opened 2023-06-03
Support use of dynamic vulkan symbols

robertblaketaylor:
Code assumes static linkage of the loader, while this is common, many apps opt to build their own symbol table to improve call performance (avoid the trampoline). 

Added TRACY_VK_USE_SYMBOL_TABLE to direct Tracy to load the symbols dynamically.

### DIFF
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -1519,8 +1519,6 @@ To mark a GPU zone use the \texttt{TracyVkZone(ctx, cmdbuf, name)} macro, where 
 
 You also need to periodically collect the GPU events using the \texttt{TracyVkCollect(ctx, cmdbuf)} macro\footnote{It is considerably faster than the OpenGL's \texttt{TracyGpuCollect}.}. The provided command buffer must be in the recording state and outside a render pass instance.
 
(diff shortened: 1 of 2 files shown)


wolfpld:
Please update the manual with instructions on how to use this.

manual/tracy.tex
```diff
@@ -1519,6 +1519,8 @@ \subsubsection{Vulkan}
 
 You also need to periodically collect the GPU events using the \texttt{TracyVkCollect(ctx, cmdbuf)} macro\footnote{It is considerably faster than the OpenGL's \texttt{TracyGpuCollect}.}. The provided command buffer must be in the recording state and outside a render pass instance.
 
+For cases where you are not static linking the Vulkan loader, you can use TRACY_VK_USE_SYMBOL_TABLE to have the underlying implementation dynamically load the Vulkan symbols needed for sampling.
```
wolfpld, line 1522: You are modifying existing interfaces. Describe how the new approach changes what you need to pass to macros, preferably with an example that shows how things look in both approaches.

Use proper formatting for fixed-width text. The macro name must be properly escaped to work as intended.

PR #576, opened 2023-06-22
fix: add missing #endif in 'TracyVulkan.hpp'

Light7734:


robertblaketaylor:
@Light7734 thanks for fixing this, looks like an error while iterating and copying over code.

Light7734:
Happy to help ^^
Thanks for the great tool you've given us :heart:

PR #581, opened 2023-07-11
Added cmake versioning and removed hardcoded share dir path

Christian-Prather:
### Changes and Reason
Added a version string to the cmake to allow the shared library to be set up with correct version number and symlink. This is particularity important when incorporating Tracy into a [Yocto](https://www.yoctoproject.org/) recipe. The outcome of this change is a .so file and symlink as follows 
```
lrwxrwxrwx  1 christian christian     19 Jul 11 16:29 libTracyClient.so -> libTracyClient.so.0
lrwxrwxrwx  1 christian christian     23 Jul 11 16:29 libTracyClient.so.0 -> libTracyClient.so.0.9.1
-rwxrwxr-x  1 christian christian 829344 Jul 11 16:29 libTracyClient.so.0.9.1
```
Also replaced the hard coded share dir with the built in [cmake variable](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) for the share dir to match the rest of the cmake.

Hopefully you find these acceptable.

### Testing 
Able to build and use the library just as before and if used in Yocto image successfully adds library to image.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,14 @@
 cmake_minimum_required(VERSION 3.10)
 
-project(Tracy LANGUAGES CXX)
+# Run version helper script
+include(version.cmake)
+
+set(TRACY_VERSION_MAJOR ${major})
+set(TRACY_VERSION_MINOR ${minor})
+set(TRACY_VERSION_PATCH ${patch})
+set(TRACY_VERSION_STRING "${TRACY_VERSION_MAJOR}.${TRACY_VERSION_MINOR}.${TRACY_VERSION_PATCH}")
+
+project(Tracy LANGUAGES CXX VERSION ${TRACY_VERSION_STRING})
 
 if(${BUILD_SHARED_LIBS})
 	set(DEFAULT_STATIC OFF)
@@ -84,6 +92,8 @@ endif()
 include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
 
+set_target_properties(TracyClient PROPERTIES VERSION ${PROJECT_VERSION})
+
 set(tracy_includes
     ${TRACY_PUBLIC_DIR}/tracy/TracyC.h
     ${TRACY_PUBLIC_DIR}/tracy/Tracy.hpp
@@ -146,10 +156,10 @@ install(FILES ${common_includes}
 install(EXPORT TracyConfig
         NAMESPACE Tracy::
         FILE TracyTargets.cmake
-        DESTINATION share/Tracy)
+        DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/Tracy)
 include(CMakePackageConfigHelpers)
 configure_package_config_file(${CMAKE_CURRENT_SOURCE_DIR}/Config.cmake.in
         "${CMAKE_CURRENT_BINARY_DIR}/TracyConfig.cmake"
-        INSTALL_DESTINATION share/Tracy)
+        INSTALL_DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/Tracy)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/TracyConfig.cmake
-        DESTINATION share/Tracy)
+        DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/Tracy)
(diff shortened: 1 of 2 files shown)


Christian-Prather:
Hey so sorry for the late response, I attached the libraries that are generated in my Yocto dev system and should of shown the default creation of them on a native system I have since updated it. You can see in the updated PR comment that this is handled by default with the additional symlink that is created. I hope this answers your question properly!

wolfpld:
Please read about soname: https://en.wikipedia.org/wiki/Soname

There is no interchangeability between versions of the shared library produced by Tracy. The symlinks you want to create say otherwise.

I can merge the patch as is if you want, but I think you are shooting yourself in the foot, which will only become apparent when a new version is released.

Christian-Prather:
Okay thanks for your patience on that! I believe I am with you now, since the MINOR version bumps are essentially being treated as MAJOR due to breaking changes can I assume the PATCH version would be API and ABI compatible i.e 0.9.1 and 0.9.2 are interchangeable? If so I believe I can adjust the soname to align with the MINOR version instead of the MINOR.
```
lrwxrwxrwx  1 christian christian     19 Jul 14 14:14 libTracyClient.so -> libTracyClient.so.9
-rwxrwxr-x  1 christian christian 829344 Jul 14 14:14 libTracyClient.so.0.9.1
lrwxrwxrwx  1 christian christian     23 Jul 14 14:14 libTracyClient.so.9 -> libTracyClient.so.0.9.1
```

Also I have to say I absolutely love this project, open source for the win!

wolfpld:
Unfortunately, there is no guarantee of ABI compatibility between patch levels. The correct solution would be to link `libTracyClient.so` directly to `libTracyClient.so.0.9.1` and then, for example, to `libTracyClient.so.0.9.2` when a new version is released. This way, the linker can see the `.so` library name, but the executable will point to the exact `.so.0.9.1` version.

Christian-Prather:
Okay awesome I can totally do that, I want to confirm its behavior in Yocto before though, since that was a driving force of this, if thats alright so ill push and update later tonight or tomorrow.

Christian-Prather:
Updated PR, the new produced libraries are as follows
```
lrwxrwxrwx  1 christian christian     23 Jul 14 14:31 libTracyClient.so -> libTracyClient.so.0.9.1
-rwxrwxr-x  1 christian christian 829344 Jul 14 14:31 libTracyClient.so.0.9.1

```

wolfpld:
There's another issue related to how the version number is specified. Please see #541 for the explanation and possible solution approach.

Christian-Prather:
Okay, what is your process for calling ```make-build.sh```? Do you update the ```TracyVersion.hpp``` file then call it?
Just trying to get an idea of your release flow. I'm guessing I could at a .sh script to that which would edit the ```CMakelists.txt``` but I'm also wondering if you would just benefit from a GitHub actions pipeline to auto release on button press? If you would I am happy to do that for you, I've done quite a few for projects at my work that we have open sourced.

wolfpld:
> Okay, what is your process for calling `make-build.sh`? Do you update the `TracyVersion.hpp` file then call it?

Basically yes. The header file is the primary source of version information, as it governs what is displayed in the UI and compatibility when loading traces from disk.

> I'm also wondering if you would just benefit from a GitHub actions pipeline to auto release on button press? If you would I am happy to do that for you, I've done quite a few for projects at my work that we have open sourced.

It would be nice. There are some pipelines set up, but they are only used to check if the build passes on the primary platforms. The release process is a bit manual at the moment.

Christian-Prather:
Okay I'm happy to take a stab at a pipeline flow to see if you like it. It will probably take me a week doing it between lunch breaks and after work to fully test. My only question is do you want it as part of this PR, or have this on in then the pipeline, or pipeline then this one?

wolfpld:
It would be best to finish up this PR and then make another one, just for the pipelines work.

Christian-Prather:
Okay sounds good, I can add a parser in the cmake for the TracyVersion header to at least get the stuff synced manually and add a pipeline later this week

Christian-Prather:
I added a cmake script that gets called at the start of the cmakelists to process the header file. Tested and its producing correct versioned library. Let me know what you think.

PR #590, opened 2023-07-31
Added binding links to the README.md

clibequilibrium:
This PR adds third party binding links so the user can easily navigate them via readme.


wolfpld:
Thanks for including the other bindings.

PR #591, opened 2023-08-01
C++20 co_await support prototype

n0lavar:
Hi!
I haven't seen coroutine support in any profiler so far, however it would be an incredibly handy feature. Now to profile coroutines I have to use zone macros only for synchronous sections. I created wrappers for awaitable objects in which I overloaded `await_suspend`, `await_resume` and `operator co_await`, and spesial macro for async zone.

Here is a brief example of what it looks like:
```cpp
tracy::Result<concurrencpp::result<void>> LoadChunk()
{
    ZoneAsyncC(tracy::Color::Azure);

    co_await tracy::Awaitable(concurrencpp::resume_on(g_pThreadPoolExecutor));
    co_await PrepareChunk();

    SimulateWork(50, 250);

    co_await LoadChunkResources();

    {
        ZoneScopedN("Sync work");
        SimulateWork(50, 250);
    }

    co_await ProceedChunkResources();
    SimulateWork(50, 250);
}
```

The limitations are:
* awaitable and result (aka task) classes must have move constructor
* user must always use tracy wrappers
* zone macro must be at the very beginning of a coroutine

Here's what it looks like at the moment. I added messages for the first "fiber" to make it easier to track.
image
image

I used concurrencpp for this example, but I had to modify it and make its move constructors available. You can see the example in this project:
https://mega.nz/file/h9liTSgJ#-93vlP01T77oE7cKOG35VqnspuW7c3bay-u37YwgxhY

What I would like to add on the tracy side to support this feature:
* "CPU data"-like lines between those scopes would be really helpful
* Special `QueueType` for async zones
* Inner "scopes" have same name. I wasn't able to fix it without macros, because co_await operator can't have any arguments, even if it's `std::source_location`. However, something like CO_AWAIT(...) should help

I was also unable to fix one bug that I believe is on the tracy side. There is a lot of logging in the code now, which I had planned to turn off before opening the request. However, after removing it, the code started crashing somewhere inside tracy. After some research it turned out that the TracyAsyncEvents.hpp::84 and TracyAsyncEvents.hpp::97 zones are to blame. Since there are no state changes there, I assume that something does not have time to happen inside tracy. I request you to help me with this problem as I don't have enough knowledge in this library.

### DIFF
--- a/public/client/TracyAsyncEvents.hpp
+++ b/public/client/TracyAsyncEvents.hpp
@@ -2,126 +2,78 @@
 
 #include "TracyProfiler.hpp"
 
-#define DEBUG_TEMP 1
-#if DEBUG_TEMP
-#include <syncstream>
-#include <iostream>
-#include <string>
+#if 0
+    #define ASSERT(x)
+#else
+    #define ASSERT(x) if(!(x)) __debugbreak();
 #endif
 
 namespace tracy
 {
     class AsyncScopedZone;
     struct SourceLocationData;
-    inline thread_local tracy::AsyncScopedZone* g_pCurrentZone = nullptr;
+    inline thread_local AsyncScopedZone* g_pCurrentZone = nullptr;
 
-    void StartAsyncEvent(const AsyncScopedZone* pAsyncScopedZone
-#if DEBUG_TEMP
-        , size_t nFiber
-#endif
-    );
-    void StopAsyncEvent();
+    void StartAsyncEvent(AsyncScopedZone* pAsyncScopedZone);
+    void StopAsyncEvent(AsyncScopedZone* pAsyncScopedZone);
 
     class AsyncScopedZone
     {
     public:
-        AsyncScopedZone(const SourceLocationData* pSourceLocation
-#if DEBUG_TEMP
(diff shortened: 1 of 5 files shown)


wolfpld:
Tracy has support for fibers/coroutines/greenthreads/whatever. All work should be done in the context of what is already available.

n0lavar:
Are you saying that we should drop this feature and use the existing functionality, or that this feature should be implemented with it?

n0lavar:
I deleted the logs and cleaned up the code, but still no solution to the problem described above. Now if you switch off ASSERT, the problem returns. It looks like there are some optimisations happening that shouldn't happen and __debugbreak() makes them disappear, but I couldn't figure out what's wrong. 

In addition, I realised that to support this functionality you need to wrap EVERY awaitable object, even every internal async library object. This is acceptable for me, but clearly not enough for a production solution.  So I will close this requester, although locally my results look very good.

Also I didn't understand how you propose to use fibers here, I couldn't adapt existing solutions to my needs. C++20 coroutines don't have unique names, neither do they have a clear beginning and end, in my opinion your solution is not enough.

PR #595, opened 2023-08-17
Wrap std::numeric_limits<T>::max() in parenthesis

bjornblissing:
The windows.h header file defines the macro max. If the max macro is include it will lead to name collisions with the std::numeric_limits<T>::max() function.

One solution is to define NOMINMAX before the inclusion of windows.h. However, that might be a demanding task for a large codebase. Defining the NOMINMAX as global define may also break previous code.

Another to solution to the problem is to wrap the numeric_limits function in parenthesis to instruct the compiler to treat it as a function and not a macro.

This commit wraps the std::numeric_limits<T>::max() calls in the public interfacing header files, with parenthesis.

### DIFF
--- a/public/client/TracyLock.hpp
+++ b/public/client/TracyLock.hpp
@@ -21,7 +21,7 @@ public:
         , m_active( false )
 #endif
     {
-        assert( m_id != std::numeric_limits<uint32_t>::max() );
+        assert( m_id != (std::numeric_limits<uint32_t>::max)() );
 
         auto item = Profiler::QueueSerial();
         MemWrite( &item->hdr.type, QueueType::LockAnnounce );
@@ -154,7 +154,7 @@ public:
 
     tracy_force_inline void CustomName( const char* name, size_t size )
     {
-        assert( size < std::numeric_limits<uint16_t>::max() );
+        assert( size < (std::numeric_limits<uint16_t>::max)() );
         auto ptr = (char*)tracy_malloc( size );
         memcpy( ptr, name, size );
         auto item = Profiler::QueueSerial();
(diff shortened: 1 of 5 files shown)


wolfpld:
> Another to solution to the problem is to wrap the numeric_limits function in parenthesis to instruct the compiler to treat it as a function and not a macro.

Can you point me to a relevant section in the language standard that has an explanation for this behavior?

bjornblissing:
I am not that well-versed in how to interpret C++-standard language. But I guess that by wrapping the library indentifier protects it from macro substitution in during phase 4 of the lexer.

The C99 standard is a bit more clear:

> 7.1.4 Use of library functions
Any function declared in a header may be additionally implemented as a function-like macro deﬁned in the header, so if a library function is declared explicitly when its header is included, one of the techniques shown below can be used to ensure the declaration is not affected by such a macro. Any macro deﬁnition of a function can be suppressed locally by enclosing the name of the function in parentheses, because the name is then not followed by the left parenthesis that indicates expansion of a macro function name. For the same syntactic reason, it is permitted to take the address of a library function even if it is also deﬁned as a macro.

This is the official recommendation from Microsoft on how to handle name conflicts between library code and macros:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/recommendations-for-choosing-between-functions-and-macros

The solution works for all compilers I have checked:
https://godbolt.org/z/1P73o8TvK

The pattern can already be found in Tracy repo:
Once here: https://github.com/wolfpld/tracy/blob/84eff4b05d0cd7aede6fe4bb32ca0b9e9c8389e1/server/tracy_robin_hood.h#L2131

And extensive use in this file:
https://github.com/wolfpld/tracy/blob/84eff4b05d0cd7aede6fe4bb32ca0b9e9c8389e1/import-chrome/src/json.hpp#L4018

wolfpld:
Thanks!

PR #597, opened 2023-08-20
Use patchable rdtsc sequence to avoid slowdowns under rr

Keno:
We (Julia) ship both support for using tracy to trace julia applications, as well as using `rr` (https://github.com/rr-debugger/rr) for record-replay debugging. After our most recent rebuild of tracy, users have been reporting signfificant performance slowdowns when `rr` recording a session that happens to also load the tracy library (even if tracing is not enabled). Upon further examination, the recompile happened to trigger a protective heuristic that disabled rr's patching of tracy's use of `rdtsc` because an earlier part of the same function happened to look like a conditional branch into the patch region. See https://github.com/rr-debugger/rr/pull/3580 for details. To avoid this issue occurring again in future rebuilds of tracy, adjust tracy's `rdtsc` sequence to be `nopl; rdtsc`, which (as of of the linked PR) is a sequence that is guaranteed to bypass this heuristic and not incur the additional overhead when run under rr.


wolfpld:
Please put the `nop` insertion under a compile-time flag. Otherwise, everyone would pay the 10-byte-per-function I$ cost even if they do not use `rr`.

public/client/TracyProfiler.hpp
```diff
@@ -209,7 +209,22 @@ class Profiler
         if( HardwareSupportsInvariantTSC() )
         {
             uint64_t rax, rdx;
+#ifdef TRACY_NO_EXTRA_NOPSLEDS
             asm volatile ( "rdtsc" : "=a" (rax), "=d" (rdx) );
+#else
```
wolfpld, line 214: Currently you need to opt-out. The condition should be reversed.

Keno:
(I handed this off to @topolarity, just to avoid confusing as to why he's committing here ;) )

PR #609, opened 2023-09-05
adapt libbacktrace to musl

xhebox:
__WORDSIZE is not defined on MUSL by including limits.h only

### DIFF
--- a/public/libbacktrace/config.h
+++ b/public/libbacktrace/config.h
@@ -1,4 +1,8 @@
 #include <limits.h>
+#if defined(__linux__) && !defined(__GLIBC__) && !defined(__WORDSIZE)
+// include __WORDSIZE headers for musl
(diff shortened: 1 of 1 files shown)


xhebox:
Any review?

wolfpld:
I would expect some explanation for the logic change you introduced in the second commit. You seem to equate the word size with the long size, which is not true in some data models, i.e. LLP64.

xhebox:
> which is not true in some data models, i.e. LLP64.

Oh, makes sense.

> I would expect some explanation for the logic change you introduced in the second commit

__WORD_SIZE is not defined in `limits.h` on musl libc headers. It is defined in `/include/reg.h` though, that won't work on macos. Maybe

xhebox:
I've rebased to the latest master branch.

PR #618, opened 2023-09-11
Fixed Vulkan query id generation not being thread safe

gan74:
Vulkan query id generation isn't thread safe, which can lead to a variety of problems, including validation errors and even crashes, if GPU zones are emitted from several threads. 

This fixes it, by using an atomic counter to generate the id.

### DIFF
--- a/public/tracy/TracyD3D12.hpp
+++ b/public/tracy/TracyD3D12.hpp
@@ -25,7 +25,7 @@
 
 namespace tracy
 {
-	class D3D12ZoneScope {};
+    class D3D12ZoneScope {};
 }
 
 using TracyD3D12Ctx = void*;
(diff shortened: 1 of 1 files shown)


wolfpld:
Shouldn't you use a more relaxed memory ordering than seq_cst?

gan74:
The query generation should only care about the increment being atomic. From what I understand, we can indeed relax the ordering to `memory_order_relaxed`

public/tracy/TracyVulkan.hpp
```diff
@@ -255,6 +257,7 @@ class VkCtx
             return;
         }
 #endif
+        assert(m_head >= m_tail);
```
wolfpld, line 260: There's also this access to `m_head` which should have memory order specfied. Moreover, value should be read before checking the assert, as the `m_head` value may change between the assert and the usage in line 270 below.

gan74:
This assert shouldn't even be there, tail could be being modified by another thread.

I took a second look at the whole PR to check for any other problems, and caught one. It should be good now, and I squashed all my commits into a single one. (For some reason this closed the PR?)

To recapitulate: 

- `m_head` is now atomic, ever increasing, and only written inside NextQueryId. (`m_tail` trails `m_head` and thus is also ever increasing)
- We do not care about the sequencing of value outputted by NextQueryId, so increments on `m_head` can use a relaxed memory order. 
- `m_tail` and `m_oldCnt` are only ever used inside Collect, thus do not require additional care

wolfpld:
Yeah, just make a new PR.

PR #624, opened 2023-09-21
Fixes around UDP broadcast

slomp:


wolfpld:
I will need some arguments for the port change.

slomp:
Ah, sorry, the explanation was in the commit messages.

Basically, Apple/macOS is using this port for something, which prevents the profiler to `bind()` the udp broadcast socket.

I noticed that clients were not showing in the profiler window and decided to investigate.
I fixed [a couple of bugs](https://github.com/wolfpld/tracy/pull/624/commits/83b52d9a2042e8bff70f89cffedd6d6a04a740c7) but still got "error 48 address in use" during socket `bind()`ing. I then found out that port 8086 was being used by the system for something, and that I could not even kill the related process.

Precise information about it is a bit scarce, but this is what I managed to find:

It has something to do with "Web service, iTunes Radio streams"
https://support.apple.com/en-us/HT202944

It's also an official IANA port (for "Distributed SCADA Networking Rendezvous Port")
https://www.adminsub.net/tcp-udp-port-finder/8086
https://www.speedguide.net/port.php?port=8086
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=69&search=&page=110

To be fair, I think even 8087 can become problematic given the port assignments above.
If we want to stick with the Intel homage, we could use port 4004, which does not seem to be used for anything "mainstream".

slomp:
Ok, I just ran some experiments here.

It's unclear what's prompting the use of port 8086.
I did reboot (and shutdown) the M1 MacBook yesterday and the port was still assigned...

I tried again today, and the port was free...

Here's what I'm gonna do: update this PR to only account for the bug fixes, and leave the port discussion open as an issue for now.

slomp:
Moved the port issue to https://github.com/wolfpld/tracy/issues/625

PR #629, opened 2023-10-01
Add TracyConstExpr override option to fix some compiler error.

qiutang98:
Hi, wolfpld san.
When /ZI option enable in MSVC compiler, the macro "__LINE__" can't treat as an constexpr, which will cause error C2131.
Seems can override "TracyLine" as const value to fix the compiler error problem, but may cause same scope object repeat name.
It will very convenient If can use an optional macro to override "constexpr" to "const".

### DIFF
--- a/public/tracy/Tracy.hpp
+++ b/public/tracy/Tracy.hpp
@@ -16,6 +16,10 @@
 #  define TracyLine __LINE__
 #endif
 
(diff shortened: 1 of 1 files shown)


wolfpld:
"Edit & continue" breaks the C++ standard and is not supported.

PR #639, opened 2023-10-17
Correct load method for Vk instance extension symbol

martty:
`vkGetPhysicalDeviceCalibrateableTimeDomainsEXT` is an instance function, and it should be loaded via `vkGetInstanceProcAddr`.

### DIFF
--- a/public/tracy/TracyVulkan.hpp
+++ b/public/tracy/TracyVulkan.hpp
@@ -61,7 +61,9 @@ namespace tracy
     Operation(vkResetQueryPool)
 
 #define LoadVkDeviceExtensionSymbols(Operation) \
-    Operation(vkGetCalibratedTimestampsEXT) \
+    Operation(vkGetCalibratedTimestampsEXT)
+
+#define LoadVkInstanceExtensionSymbols(Operation) \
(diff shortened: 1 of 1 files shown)


wolfpld:
Can you point me to documentation that would back this up? `vkGetPhysicalDeviceCalibrateableTimeDomainsEXT` is provided by `VK_EXT_calibrated_timestamps`, which is a device extension.

martty:
I too have only now found about this, so apologies for being slightly incorrect in the wording.
The relevant passages are: 
https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap4.html#initialization-phys-dev-extensions

And:
https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap4.html#vkGetDeviceProcAddr
> The returned function pointer is of type [PFN_vkVoidFunction](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap4.html#PFN_vkVoidFunction), and must be cast to the type of the command being queried before use. The function pointer must only be called with a dispatchable object (the first parameter) that is device or a child of device.
> In this function, device-level excludes all physical-device-level commands.

PR #641, opened 2023-10-19
Meson improvements

YaLTeR:
A number of meson improvements from my work to instrument Mutter and its dependencies. See individual commits.

The header install dir fix makes it work without errors, but it's still a bit ugly because it installs stuff into `/usr/include/client/` and `/usr/include/common/`. Ideally everything should go under `/usr/include/tracy/`, but that breaks some include paths relying on `../`. I might try to figure it out later.

### DIFF
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -19,10 +19,10 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install linux libraries
       if: ${{ matrix.os == 'ubuntu-22.04' }}
-      run: sudo apt-get update && sudo apt-get -y install libdbus-1-dev libcapstone-dev libtbb-dev libdebuginfod-dev libxkbcommon-dev libegl-dev libwayland-dev
+      run: sudo apt-get update && sudo apt-get -y install libdbus-1-dev libcapstone-dev libtbb-dev libdebuginfod-dev libxkbcommon-dev libegl-dev libwayland-dev meson
     - name: Install macos libraries
       if: ${{ matrix.os == 'macOS-latest' }}
-      run: brew install capstone tbb pkg-config glfw
+      run: brew install capstone tbb pkg-config glfw meson
     - name: Profiler GUI
       run: make -j`nproc` -C profiler/build/unix debug release
     - name: Update utility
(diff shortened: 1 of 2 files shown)


wolfpld:
Can you add a github action that would run the meson build path? It would make testing the changes easier.

YaLTeR:
Yeah, let me try.

.github/workflows/gcc.yml
```diff
@@ -35,6 +35,8 @@ jobs:
       run: make -j`nproc` -C import-chrome/build/unix debug release
     - name: Library
       run: make -j`nproc` -C library/unix debug release
+    - name: Library (meson)
+      run: meson setup build && meson compile -C build
```
wolfpld, line 39: This probably needs to be excluded on macos.

YaLTeR: Why so? Appears to have worked: https://github.com/YaLTeR/tracy/actions/runs/6572890389

wolfpld: Never mind then. Maybe we should also do `ninja install` and then publish the artifacts?

YaLTeR: I thought about this a bit, especially considering that I do package this .so for my instrumented mutter repo. However, I don't think it makes much sense to do generally, because in practice you do want to set different defines based on the use-case (i.e. I set `TRACY_ONLY_LOCALHOST` for anything I send to others, yet I leave it off for my own machines), and a built library supports only a single set of defines. If it was configured at runtime instead, then it would make more sense to publish it IMO.

YaLTeR: I guess `TRACY_ONLY_LOCALHOST` is not a good example, `TRACY_MANUAL_LIFETIME` is a better one that needs explicit support from the profiled application.

wolfpld: > If it was configured at runtime instead, then it would make more sense to publish it IMO.

My intention was more to see what the install job does, if all the files are where they should be, etc. It may not be apparent just from looking at the build script.

YaLTeR: Hm, then could do an install with a prefix set to local dir. Meson prints all files it installs.

wolfpld: Sure.

YaLTeR: https://github.com/YaLTeR/tracy/actions/runs/6573262386

PR #642, opened 2023-10-19
Defer GPU context creation from the C API

YaLTeR:
Same as how the C++ API does it. Otherwise with on demand mode the profiler never receives the GPU context info.

This makes things not crash, but there seems to be a different issue where GPU zones always start close to timestamp 0, rather than after the on-demand connection.

image

And if I connect again later, the GPU zones will still appear at the very beginning.

I'm not sure after glancing through the code what is at fault here. Moreover, even without on-demand, the GPU zones appear earlier than they should, so there might be two separate issues at play. Some of them are also overlapping weirdly. I am on AMD RX 6700M on Linux 6.5.6 and using OpenGL timestamp queries.

### DIFF
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -4258,6 +4258,11 @@ TRACY_API void ___tracy_emit_gpu_new_context( ___tracy_gpu_new_context_data data
     tracy::MemWrite( &item->gpuNewContext.context, data.context );
     tracy::MemWrite( &item->gpuNewContext.flags, data.flags );
     tracy::MemWrite( &item->gpuNewContext.type, data.type );
+
+#ifdef TRACY_ON_DEMAND
+    tracy::GetProfiler().DeferItem( *item );
+#endif
+
     TracyLfqCommitC;
 }
 
@@ -4270,6 +4275,11 @@ TRACY_API void ___tracy_emit_gpu_context_name( const struct ___tracy_gpu_context
     tracy::MemWrite( &item->gpuContextNameFat.context, data.context );
     tracy::MemWrite( &item->gpuContextNameFat.ptr, (uint64_t)ptr );
     tracy::MemWrite( &item->gpuContextNameFat.size, data.len );
+
+#ifdef TRACY_ON_DEMAND
+    tracy::GetProfiler().DeferItem( *item );
+#endif
+
     TracyLfqCommitC;
 }
 


wolfpld:
There is some logic in Tracy to detect overflows and treat the readings as continuous in such cases, but this requires the timestamp stream to be active at a reasonable granularity. That's not something that happens in on-demand mode. However, as noted in the comment linked above, the drivers do not guarantee that the overflow will occur on a power-of-two boundary.

YaLTeR:
I tried dumping the log as noted in that comment and it looks mostly tame with the occasional big negative value every few seconds. First every ~two seconds, then several times a second (???). So in on-demand mode there's no way for Tracy to keep track of this overflow and tell very early timestamps from later ones.

I should try to create the GPU context only upon profiler connection maybe to help at least a tiny bit with this offset.

wolfpld:
2 seconds sounds like 2^31 nanoseconds.

YaLTeR:
Mhm, this also means that if there isn't a frame for a little bit (common in compositors since they go idling), Tracy can't tell how many overflows have occurred. Is there any event to sync GPU timestamp to CPU timestamp again? I could send that every frame since we already measure that in Mutter.

wolfpld:
What you are asking for is provided by calibrated contexts, which are only available in Vulkan and I think D3D12.

There is no resynchronization for OpenGL because it would require stopping the CPU until the GPU catches up, and you don't want to do that if you are profiling a game.

It shouldn't be hard to add something like this, though.

YaLTeR:
According to one of the graphics developers from mutter, these frequent timestamp resets are caused by AMD GPU hardware graphics blocks shutting down for power saving. Something that requires firmware changes to fix, and is unlikely to happen.

It's possible to force a high performance profile by echoing `manual` into `/sys/devices/pci*/*/*/power_dpm_force_performance_level` and echoing the digit corresponding to the COMPUTE profile into `pp_power_profile_mode` from the same directory, which prevents the graphics blocks from shutting down. When I did this on my system, the GPU timestamps pretty much stopped overflowing, even after several minutes I was able to connect with on-demand and see correct GPU zones, save for a few ms of drift.

I think I'll still try to figure out the CPU/GPU timestamp sync message to avoid the drift in this case, and possibly even make it work without forcing the power profile.

wolfpld:
Thanks, I documented this in 5994629c.

PR #644, opened 2023-10-22
Add GpuTimeSync event

YaLTeR:
Allows to resynchronise GPU and CPU timestamps during profiling. I tested this with my Mutter setup, and it seems to work very well, even with the default power profile.

Do I need to bump the file format version and/or meson version, and to what number?


wolfpld:
> Do I need to bump the file format version and/or meson version, and to what number?

No, these are fine as they are. But you have altered the network protocol and thus you must bump its revision. See `public/common/TracyProtocol.hpp`.

wolfpld:
Oh, and you should also document this in the manual.

YaLTeR:
Did both.

Issue #637, opened 2023-10-15
TRACY_MANUAL_LIFETIME and delayed or optional startup


YaLTeR:
I am using Tracy to profile Mutter and GNOME Shell. I would also like to instrument some libraries like GLib or GJS.

These libraries are loaded into many different processes. However, only the particular process I'm interested in (GNOME Shell) should do any profiling, or start any Tracy threads—otherwise there's a lot of unnecessary overhead and spam in the connection list.

TRACY_MANUAL_LIFETIME is perfect for this, except there's a problem: calling most Tracy functions will crash with an assertion before the profiler is started. What I want instead is for them to do nothing before the profiler is started, kind of like TRACY_ON_DEMAND. I will call `___tracy_startup_profiler()` in the one binary that I need, making the instrumentation live, and instrumentation in GLib loaded into all other processes should do nothing.

Right now in my Tracy fork I added a workaround function and check it manually in GLib before calling other Tracy functions.

```cpp
TRACY_API int ___tracy_is_profiler_started( void )
{
    return tracy::s_profilerData != nullptr;
}
```

I think it would be nicer if it worked more like TRACY_ON_DEMAND, as in if it also added the check automatically into the relevant Tracy functions. It could be a new define, or it could just do it when TRACY_MANUAL_LIFETIME is set.

Is this something that would be accepted upstream, and if so, how exactly should I go about this?

wolfpld:
For your use case it would probably be best to just compile a separate copy of glib with tracing enabled and LD_LIBRARY_PATH (or compatible) it for the application you want to profile.

If you want to manually manage the lifetime of the profiler, you will also need to manually manage when you can and cannot call the profiler. Creating a set of wrapper functions that mimic the available API might be a solution that works for you.

> Is this something that would be accepted upstream, and if so, how exactly should I go about this?

I don't think that the cost of maintenance would make this viable, but if you think that you can come up with something that would be manageable, then give it a try.

PR #652, opened 2023-10-25
Add TracyIsStarted

YaLTeR:
When using TRACY_MANUAL_LIFETIME, calling most Tracy functions before starting the profiler results in an assertion. Notably, even TracyIsConnected is affected. There is, however, no function to check if the profiler had already started. This PR adds such a function.

Related: #637. This PR doesn't add any automatic checks, just allows application code to do them manually.


wolfpld:
How will this work in a multithreaded environment? For proper use, `s_profilerData` should be an atomic, but I don't want to introduce an atomic here.

wolfpld:
Probably a good solution.

wolfpld:
It may be worth noting that updating `s_isProfilerStarted` in `ShutdownProfiler()` will not prevent access to `s_profilerData` in functions that are already after the `IsProfilerStarted()` check.

PR #656, opened 2023-11-09
respect mesons default_library option

apache-hb:
Meson provides a build flag called `default_library` which is the standard way to specify the type of library to build. I left `tracy_shared_libs` in to prevent a breaking change.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -95,8 +95,8 @@ if get_option('tracy_no_crash_handler')
 endif
 
 tracy_shared_libs = get_option('default_library') == 'shared'
-if tracy_shared_libs and not get_option('tracy_shared_libs')
-  warning('tracy_shared_libs is set to false, but default_library is set to shared. Building shared library.')
+if not tracy_shared_libs and get_option('tracy_shared_libs')
+  warning('tracy_shared_libs is set to true, but default_library is set to static. Building static library.')
 endif
 
 if tracy_shared_libs
--- a/meson.build
+++ b/meson.build
@@ -1,106 +1,108 @@
-project('tracy', ['cpp'], version: '0.10.0')
+project('tracy', ['cpp'], version: '0.10.0', meson_version: '>=0.60.0')
 
+# internal compiler flags
 tracy_compile_args = []
 
+# compiler flags shared between client and server
+tracy_common_args = []
+
 if get_option('tracy_enable')
-  tracy_compile_args += ['-DTRACY_ENABLE']
+  tracy_common_args += ['-DTRACY_ENABLE']
 endif
 
 if get_option('tracy_on_demand')
-  tracy_compile_args += ['-DTRACY_ON_DEMAND']
+  tracy_common_args += ['-DTRACY_ON_DEMAND']
 endif
 
 if get_option('tracy_callstack')
-  tracy_compile_args += ['-DTRACY_CALLSTACK']
+  tracy_common_args += ['-DTRACY_CALLSTACK']
 endif
 
 if get_option('tracy_no_callstack')
-  tracy_compile_args += ['-DTRACY_NO_CALLSTACK']
+  tracy_common_args += ['-DTRACY_NO_CALLSTACK']
 endif
 
 if get_option('tracy_no_callstack_inlines')
-  tracy_compile_args += ['-DTRACY_NO_CALLSTACK_INLINES']
+  tracy_common_args += ['-DTRACY_NO_CALLSTACK_INLINES']
 endif
 
 if get_option('tracy_only_localhost')
-  tracy_compile_args += ['-DTRACY_ONLY_LOCALHOST']
+  tracy_common_args += ['-DTRACY_ONLY_LOCALHOST']
 endif
 
 if get_option('tracy_no_broadcast')
-  tracy_compile_args += ['-DTRACY_NO_BROADCAST']
+  tracy_common_args += ['-DTRACY_NO_BROADCAST']
 endif
 
 if get_option('tracy_only_ipv4')
-  tracy_compile_args += ['-DTRACY_ONLY_IPV4']
+  tracy_common_args += ['-DTRACY_ONLY_IPV4']
 endif
 
 if get_option('tracy_no_code_transfer')
-  tracy_compile_args += ['-DTRACY_NO_CODE_TRANSFER']
+  tracy_common_args += ['-DTRACY_NO_CODE_TRANSFER']
 endif
 
 if get_option('tracy_no_context_switch')
-  tracy_compile_args += ['-DTRACY_NO_CONTEXT_SWITCH']
+  tracy_common_args += ['-DTRACY_NO_CONTEXT_SWITCH']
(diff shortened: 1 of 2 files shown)


wolfpld:
@YaLTeR Can you look at this?

meson.build
```diff
@@ -1,106 +1,108 @@
-project('tracy', ['cpp'], version: '0.10.0')
+project('tracy', ['cpp'], version: '0.10.0', meson_version: '>=0.54.0')
 
+# internal compiler flags
 tracy_compile_args = []
 
+# compiler flags shared between client and server
```
YaLTeR, line 6: Something like this is clearer I think
```suggestion
# compiler flags shared between the capture library itself and the code using it
```

meson.build
```diff
@@ -167,25 +169,20 @@ if compiler.get_id() != 'msvc'
...
-  tracy = shared_library('tracy', tracy_src, tracy_header_files,
-    dependencies        : [ threads_dep ],
-    include_directories : tracy_public_include_dirs,
-    override_options    : override_options,
-    install             : true)
-else
-  tracy = static_library('tracy', tracy_src, tracy_header_files,
+tracy_compile_args += tracy_common_args
+
+tracy = library('tracy', tracy_src, tracy_header_files,
```
YaLTeR, line 174: As far as I can tell, if `default_library` is `static`, then this will build a static library, even if `tracy_shared_libs` is enabled. I would consider this a breaking change, as using a static Tracy capture library instead of shared can cause assertions with certain multi-DLL application setups. So, it might be better to just drop the `tracy_shared_libs` option in this case.

meson.build
```diff
@@ -1,106 +1,108 @@
...
+  tracy_common_args += ['-DTRACY_TIMER_FALLBACK']
 endif
 
 if get_option('tracy_no_crash_handler')
-  tracy_compile_args += ['-DTRACY_NO_CRASH_HANDLER']
+  tracy_common_args += ['-DTRACY_NO_CRASH_HANDLER']
 endif
 
-add_project_arguments(tracy_compile_args, language : 'cpp')
+tracy_shared_libs = get_option('default_library') == 'shared' or get_option('tracy_shared_libs')
```
YaLTeR, line 97: For future reference: there's also `default_library` = `both`, which will build both libraries at once. However, there is currently no way to set a define only for the shared library in that case. This is an issue about it: https://github.com/mesonbuild/meson/issues/3304

apache-hb: since 1.3.0 `library` supports `<lang>_shared_args` and `<lang>_static_args`, as its still in release candidate phase i decided to not use them. Im unsure as to what the recommended way to currently handle this is

meson.build
apache-hb: Perhaps an `error` if `tracy_shared_libs=true` and `default_library=static` would be a good intermediate step?, `option` inside the `meson_options.txt` has `deprecated` since 0.60 but I opted to avoid it as to not increase the minimum required meson version. If bumping the min version to 0.60 is acceptable then that could be an option.

YaLTeR: But then there's the inverse problem that `default_library=shared` and `tracy_shared_libs=false` will build a shared lib, right? Kinda weird still. I guess it's a call for @wolfpld on the option dropping policy here.

meson.build
YaLTeR: From what I gathered, the current way everyone is doing it is with an option like `tracy_shared_libs` but yeah idk. I think `default_library` should work just as well? As long as you can override it for subprojects.

meson.build
wolfpld: Please add the deprecation tag. The Debian people will find ways to complain that something doesn't work in their museum either way.

meson.build
apache-hb: default_library can be specified per subproject via `-Dsubproject:default_library=xxx` or in the toplevel projects `default_options` field in `project`

meson.build
apache-hb: added

meson.build
```diff
@@ -94,7 +94,11 @@ if get_option('tracy_no_crash_handler')
   tracy_common_args += ['-DTRACY_NO_CRASH_HANDLER']
 endif
 
-tracy_shared_libs = get_option('default_library') == 'shared' or get_option('tracy_shared_libs')
+tracy_shared_libs = get_option('default_library') == 'shared'
+if tracy_shared_libs and not get_option('tracy_shared_libs')
+  warning('tracy_shared_libs is set to false, but default_library is set to shared. Building shared library.')
```
YaLTeR, line 99: You can hit surprising assertions when `tracy_shared_libs` is enabled, but `default_library=static`. I think this check and warning should be inverted. Especially since `tracy_shared_libs` is deprecated, i.e. expected to be left at its default false value.

meson_options.txt
```diff
@@ -19,5 +19,5 @@ option('tracy_timer_fallback', type : 'boolean', value : false, description : 'U
 option('tracy_delayed_init', type : 'boolean', value : false, description : 'Enable delayed initialization of the library (init on first call)')
 option('tracy_manual_lifetime', type : 'boolean', value : false, description : 'Enable the manual lifetime management of the profile')
 option('tracy_fibers', type : 'boolean', value : false, description : 'Enable fibers support')
-option('tracy_shared_libs', type : 'boolean', value : false, description : 'Builds Tracy as a shared object')
+option('tracy_shared_libs', type : 'boolean', value : false, description : 'Builds Tracy as a shared object', deprecated : 'use default_library=shared instead')
```
YaLTeR, line 22: This argument to `deprecated` actually says which option meson should set instead. So currently it errors out:

```
tracy| DEPRECATION: Option 'tracy_shared_libs' is replaced by 'use default_library=shared instead'

subprojects/tracy/meson.build:1:0: ERROR: Tried to set unknown builtin option tracy:use default_library=shared instead
```

apache-hb: the updated version with `deprecation: 'default_library'` has an issue where specifying `-Dtracy_shared_libs=true` is now an error as `bool` cannot match into `Default library type`. is this acceptable or should i use the simple bool deprecation flag and put the deprecation notice in the description?

YaLTeR: Hm, I think simple bool with deprecation and notice in description would be better.

YaLTeR (approved):
Seems to work well now, thanks :+1:

wolfpld:
Sigh, it's always Debian.

https://github.com/wolfpld/tracy/actions/runs/6914526369/job/18812452985#step:11:14

YaLTeR:
We could try a `pip3 install --user meson` to get the latest version on the CI. Alternatively I guess roll back whatever needs the newer version.

wolfpld:
The only reason I used the ubuntu image on CI is because I wasn't aware of how the GitHub runner images relate to Docker images. I will change this to use the Arch container.

wolfpld:
Fixed with 906f73cab31ee53824420819dfeeb3e6040dbf95.

PR #659, opened 2023-11-14
Explicitly set CS_OPT_MEM to cover capstone edge case

simplyWiri:


wolfpld:
Closing, as we talked about this issue on discord and it manifested when a client was run in a specific environment, which makes little sense, as capstone is only used on the server side.

PR #665, opened 2023-11-17
Add support for offline callstack symbol resolving

tiago-rodrigues:
Add support for offline symbol resolving by setting the "TRACY_SYMBOL_OFFLINE_RESOLVE=1" env var.
Add symbol resolving functionality to the "update" tool, with
offline symbol resolvers for linux (using addr2line) and windows (using dbghelper).

There are a few reasons why resolving symbols offline might be useful:

- runtime symbol resolving can be slow and use quite a bit memory - specially on windows (which is counter productive when we are doing tracy captures to investigate memory usage and we have a lot of backtrace traffic as we instrument every memory allocation with a an interposer and inject backtraces for each alloc/free).

-  libbacktrace does not work with dlopenned shared libraries after the first symbol resolve has happened, this can be quite problematic on applications that rely heavily on dynamically loaded libs like Omniverse (we can have ~200 shared libs loaded after starting our app due to the way the extension system works. Still I'll investigate adding a potential solution for this in another PR where we refresh libbacktrace internal state.

- doing offline symbol resolving can allow a bit more flexibility on retargeting the resolving to a different set shared libraries (if symbols have been stripped for example) etc.


wolfpld:
> I haven't added documentation about this in the PR already as I wanted to float the idea of doing the offline symbol resolving first and see if there were any major objections to it first.

Looks like a very nice addition, judging by the description. Go ahead and write the documentation. I'm not sure about the name "tracy-edit" though, but nothing comes to mind that has a nice ring to it.

I'll do a proper review later.

YaLTeR:
Oh, this would be quite useful for GNOME Shell profiling. I currently have to disable symbol resolution there because GJS seems to very heavily contend on dlsym() with Tracy's symbol workers, causing slowdowns. And callstacks are quite useful.

manual/tracy.tex
```diff
@@ -1782,6 +1782,32 @@ \subsubsection{Debugging symbols}
...
+
+By default, tracy client resolves callstack symbols in a background thread at runtime.
+This process requires that tracy client load symbols for the shared libraries 
+involved, which requires additial memory allocations, and potential impact runtime performance if a lot of symbol queries are involved.
+As an alternative to to runtime symbol resolution, we can set the environment variable
+ \texttt{TRACY\_SYMBOL\_OFFLINE\_RESOLVE} to 1 and instead have tracy client only resolve
+ the minimal set of info required for offline resolution (a shared library path and an offset into that shared library).
+ 
+The generated tracy capture will have callstack frames symbols showing \texttt{[unresolved]}.
+The \texttt{tracy-edit} tool can be used to load that capture, perforce symbol resolution offline
```
wolfpld, line 1795: perforce -> perform

public/client/TracyCallstack.cpp
```diff
@@ -93,6 +93,14 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
 namespace tracy
 {
 
+// when "TRACY_SYMBOL_OFFLINE_RESOLVE" is set to "1", instead of fully resolving symbols at runtime,
+// simply resolve the offset and image name (which will be enough the resolving to be done offline)
+bool getDoOfflineSymbolResolve()
```
wolfpld, line 98: It would probably be a good idea to add a `#ifdef TRACY_SYMBOL_OFFLINE_RESOLVE` to enable hardcode enabling this at build time, like you can do with some other options. Also, rename this function to `ShouldResolveSymbolsOffline`.

public/client/TracyCallstack.cpp
```diff
@@ -157,6 +170,65 @@ void InitCallstack()
...
     SymSetOptions( SYMOPT_LOAD_LINES );
 
+#ifdef TRACY_DBGHELP_LOCK
+    DBGHELP_UNLOCK;
+#endif
+}
+
+DWORD64 dbgHelpLoadSymbolsForModule(PCSTR imageName, DWORD64 baseOfDll, DWORD bllSize)
+{
+    if( !s_doOfflineSymbolResolve )
```
wolfpld, line 180: `if( s_doOfflineSymbolResolve ) return 0;`

public/client/TracyCallstack.cpp
```diff
@@ -270,6 +329,11 @@ void EndCallstack()
 
 const char* DecodeCallstackPtrFast( uint64_t ptr )
 {
+    if( s_doOfflineSymbolResolve )
+    {
+        return "[unresolved]";
+    }
```
wolfpld, line 335: Write as a one-liner.

public/client/TracyCallstack.cpp
```diff
@@ -135,19 +143,24 @@ struct KernelDriver
...
-
+bool s_doOfflineSymbolResolve = false;
 
 void InitCallstackCritical()
 {
     ___tracy_RtlWalkFrameChain = (___tracy_t_RtlWalkFrameChain)GetProcAddress( GetModuleHandleA( "ntdll.dll" ), "RtlWalkFrameChain" );
 }
 
-void InitCallstack()
+void dbgHelpInit()
```
wolfpld, line 153: Keep function names capitalized.

public/client/TracyCallstack.cpp
```diff
@@ -157,6 +170,65 @@ void InitCallstack()
     SymInitialize( GetCurrentProcess(), nullptr, true );
     SymSetOptions( SYMOPT_LOAD_LINES );
 
+#ifdef TRACY_DBGHELP_LOCK
+    DBGHELP_UNLOCK;
+#endif
+}
+
+DWORD64 dbgHelpLoadSymbolsForModule(PCSTR imageName, DWORD64 baseOfDll, DWORD bllSize)
```
wolfpld, line 178: Follow the convention wrt to how spaces are used, etc.

public/client/TracyCallstack.cpp
```diff
@@ -157,6 +170,65 @@ void InitCallstack()
...
+    dbgHelpLoadSymbolsForModule( imageName, baseOfDll, dllSize );
+
+    ModuleCache* cachedModule = s_modCache->push_next();
+    cachedModule->start = baseOfDll;
+    cachedModule->end = baseOfDll + dllSize;
+
+    // when doing offline symbol resolution, we must store the full path of the dll for the resolving to work 
+    if( s_doOfflineSymbolResolve )
+    {
+        cachedModule->name = (char*)tracy_malloc_fast(imageNameLength + 2);
```
wolfpld, line 198: Why is there need for two additional characters?

server/TracyWorker.cpp
```diff
@@ -2411,6 +2426,19 @@ const char* Worker::GetString( const StringIdx& idx ) const
     return m_data.stringData[idx.Idx()];
 }
 
+uint32_t Worker::AddNewString(const char* newString)
```
wolfpld, line 2429: Why is there a need for a new function to add strings?

tracy-edit/src/tracy-edit.cpp
```diff
@@ -0,0 +1,220 @@
...
+                << "' (" << e.what() << ")" << std::endl;
+            continue;
+        }
+    }
+
+    SymbolResolver* resolver = CreateResolver();
+    if (resolver)
+    {
+        PatchSymbols(resolver, worker, pathSubstitutionList, args.verbose);
+
```
wolfpld, line 213: Remove newline.

tracy-edit/src/OfflineSymbolResolverDbgHelper.cpp
```diff
@@ -0,0 +1,142 @@
...
+                    SymbolEntryList& resolvedEntries)
+{
+    if( resolver )
+    {
+        return resolver->ResolveSymbolsForModule( imagePath, inputEntryList, resolvedEntries );
+    }
+    return false;
+}
+
+#endif // #ifdef _WIN32
```
wolfpld, line 142: Add EOF newlines.

tracy-edit/src/OfflineSymbolResolverDbgHelper.cpp
```diff
@@ -0,0 +1,142 @@
...
+void DestroySymbolResolver(SymbolResolver* resolver)
+{
+    delete resolver;
+}
+
+bool ResolveSymbols(SymbolResolver* resolver, const std::string& imagePath,
+                    const FrameEntryList& inputEntryList,
+                    SymbolEntryList& resolvedEntries)
+{
+    if( resolver )
```
wolfpld, line 135: `if( !resolver ) return false;`

tracy-edit/src/OfflineSymbolResolverDbgHelper.cpp
```diff
@@ -0,0 +1,142 @@
...
+SymbolResolver* CreateResolver()
+{
+    SymbolResolver* resolver = new SymbolResolver();
+    return resolver;
+}
+
+void DestroySymbolResolver(SymbolResolver* resolver)
+{
+    delete resolver;
+}
```
wolfpld, line 129: Why is `new` and `delete` relegated to functions here?

public/client/TracyCallstack.cpp
tiago-rodrigues: makes sense, I'll try to add support for the compile time ifdef in another commit later today/tomorrow as it will require a bit of #ifdefing in that file to make sure the libbacktrace/dbghelper is ifdefed out

public/client/TracyCallstack.cpp
tiago-rodrigues: you are right, it was a copy paste from the code below - fixed

server/TracyWorker.cpp
tiago-rodrigues: There is no need really, the only thing diff was this one was public and had that extra assert.
I've removed it and made StoreString public instead.

tracy-edit/src/OfflineSymbolResolverDbgHelper.cpp
tiago-rodrigues: The code was a bit different before when I had more state in SymbolResolver which is a private type in the CPP. There is no need for this right, removed it.

public/client/TracyCallstack.cpp
wolfpld: Add the ifdef as an option to the user, but also keep the current env variable based solution.

wolfpld: Ah, I see you already committed the changes. I'm not sure it's worth having ifdefs scattered all over the code. Sure, it's dead code, but it also increases the mental effort required to understand what's going on.

wolfpld:
The conflicting `gcc.yml` has been split into `linux.yml` and `macos.yml`, you should be able to just copy and paste your change to resolve the conflict. See 906f73cab31ee53824420819dfeeb3e6040dbf95.

tiago-rodrigues:
> Oh, this would be quite useful for GNOME Shell profiling. I currently have to disable symbol resolution there because GJS seems to very heavily contend on dlsym() with Tracy's symbol workers, causing slowdowns. And callstacks are very useful.

@YaLTeR: although it won't call into libbacktrace anymore to resolve symbols, not sure if it will still have some contention as it still uses  `dladdr() ` (to get the .so path and base address)? If so, maybe maintaining our own .so range mapping and updating with dl_iterate_phdr when we we get a request that doesn't land in any known ranges, would be a way to avoid it.

YaLTeR:
Hmm, if it still uses dladdr() then I suspect it will still contend :/

tiago-rodrigues:
> Hmm, if it still uses dladdr() then I suspect it will still contend :/
It's probably best to tackle this in another PR? However, I added a possible solition here https://github.com/tiago-rodrigues/tracy/commit/08db7d394bf7486d7b85d53b43ec781e6a02bb84, adding an image range cache to avoid calling dladdr() when doing offline symbol resolution (and dl_iterate_phdr is supported). It should only refresh when an address is encountered that's outside of the known ranges. It is also used for the runtime symbol resolution case to re-create libbacktrace state and allow symbol resolution for dynamically loaded shared libs (at the cost of mem leaking when it refreshes as I can't see any libbacktrace state freeing/recreation.)

CMakeLists.txt
```diff
@@ -85,6 +85,7 @@ set_option(TRACY_FIBERS "Enable fibers support" OFF)
 set_option(TRACY_NO_CRASH_HANDLER "Disable crash handling" OFF)
 set_option(TRACY_TIMER_FALLBACK "Use lower resolution timers" OFF)
 set_option(TRACE_CLIENT_LIBUNWIND_BACKTRACE "Use libunwind backtracing where supported" OFF)
+set_option(TRACY_SYMBOL_OFFLINE_RESOLVE "Instead of runtime symbol resolving, sore offset and sharedlib path for offline symbol resolution" OFF)
```
wolfpld, line 88: Is "sore" the right word here?

public/client/TracyCallstack.cpp
```diff
@@ -401,16 +468,29 @@ CallstackSymbolData DecodeSymbolAddress( uint64_t ptr )
 
 CallstackEntryData DecodeCallstackPtr( uint64_t ptr )
 {
+    InitRpmalloc();
+
+    const ModuleNameAndBaseAddress moduleNameAndAddress = GetModuleNameAndPrepareSymbols( ptr );
```
wolfpld, line 473: This is outside dbghelp lock.

public/client/TracyCallstack.cpp
```diff
@@ -938,20 +1039,42 @@ void SymInfoError( void* /*data*/, const char* /*msg*/, int /*errnum*/ )
     cb_data[cb_num-1].symAddr = 0;
 }
 
+void getSymbolForOfflineResolve(void* address, Dl_info& dlinfo, CallstackEntry& cbEntry)
```
wolfpld, line 1042: Capitalize function names.

public/client/TracyCallstack.cpp
```diff
@@ -938,20 +1039,42 @@ void SymInfoError( void* /*data*/, const char* /*msg*/, int /*errnum*/ )
...
-
         const char* symloc = nullptr;
         Dl_info dlinfo;
-        if( dladdr( (void*)ptr, &dlinfo ) ) symloc = dlinfo.dli_fname;
+        if( dladdr( (void*)ptr, &dlinfo ) )
+        {
+            symloc = dlinfo.dli_fname;
+        }
+
+        if(cb_bts && !s_shouldResolveSymbolsOffline)
```
wolfpld, line 1065: Is there a case where both these checks are necessary?

tracy-edit/src/tracy-edit.cpp
```diff
@@ -0,0 +1,213 @@
...
+    struct option long_opts[] = 
+    {
+        { "help", no_argument, NULL, 'h' },
+        { "verbose", no_argument, NULL, 'v' },
+        { "resolveSymbols", no_argument, NULL, 'r' },
+        { "substitutePath", required_argument, NULL, 's' },
+        { "compression", required_argument, NULL, 'c' },
+        { "compressesionLevel", required_argument, NULL, 'l' },
+        { NULL, 0, NULL, 0 }
+    };
```
wolfpld, line 84: Can you unify processing of compression parameters with the `update` utility?

tracy-edit/src/OfflineSymbolResolver.cpp
```diff
@@ -0,0 +1,136 @@
...
+    tracy::StringLocation location = worker.StoreString( str.c_str(), str.length() );
+    return tracy::StringIdx( location.idx );
+}
+
+bool PatchSymbols( tracy::Worker& worker, const PathSubstitutionList& pathSubstituionlist, bool verbose )
+{
+    uint64_t callstackFrameCount = worker.GetCallstackFrameCount();
+    std::string relativeSoNameMatch = "[unresolved]";
+
+    std::cout << "Found '" << callstackFrameCount << "' callstack frames. Batching into image groups..." << std::endl;
```
wolfpld, line 39: Is the count supposed to be quoted here?

CMakeLists.txt
tiago-rodrigues: sorry it was a typo, fixed

public/client/TracyCallstack.cpp
tiago-rodrigues: I've moved the locking up to include it.
Maybe it would be better to have a scoped wrapper for this lock though.

public/client/TracyCallstack.cpp
tiago-rodrigues: no there shouldn't be, fixed it

tracy-edit/src/tracy-edit.cpp
tiago-rodrigues: oh, I have to admit I didn't realize the update tool already did that. Maybe I should actually move the symbol resolution to it instead of creating the tracy-edit tool then?

wolfpld: It may make sense to do that. Try to keep the new functionality in a separate source file.

tiago-rodrigues: done

tracy-edit/src/OfflineSymbolResolver.cpp
tiago-rodrigues: I've remove all the numeric quoting in there

wolfpld:
I did a quick test and the resolved symbols don't show up in the symbol statistics list. Is this something you will be working on? If not, I think the documentation should state that there are functional differences.

tiago-rodrigues:
> I did a quick test and the resolved symbols don't show up in the symbol statistics list. Is this something you will be working on? If not, I think the documentation should state that there are functional differences.

Right, I forgot to tackle that one. I was mainly using the memory view for the workflow I'm working right now.
I've updated the documentation to reflect this limitation. 

I'll try to do another PR after where I'll attempt to fixup the symbol map for the newly resolved symbols. However it might be a bit more involved if I ended up avoiding calling dladdr (which I'll have another PR as well in order to reduce the contention mentioned by YalTeR : https://github.com/wolfpld/tracy/pull/665#issuecomment-1819795848).

PR #666, opened 2023-11-19
Don't error on frame end without start

YaLTeR:
With on-demand profiling we're very likely to connect in the middle of a discontinuous frame and thus receive a frame end without any preceding frame start. So don't error out in this case.

This makes on-demand profiling usable with discontinuous frames.

If this is the right approach here, I can also remove the `FrameEnd` failure stuff altogether as it seems otherwise unused.

### DIFF
--- a/server/TracyWorker.cpp
+++ b/server/TracyWorker.cpp
@@ -5077,13 +5077,12 @@ void Worker::ProcessFrameMarkEnd( const QueueFrameMark& ev )
     } );
 
     assert( fd->continuous == 0 );
-    const auto time = TscTime( ev.time );
     if( fd->frames.empty() )
     {
-        if ( !m_ignoreFrameEndFaults )
-            FrameEndFailure();
+        if( !m_ignoreFrameEndFaults ) FrameEndFailure();
         return;
     }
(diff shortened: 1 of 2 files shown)


wolfpld:
This would remove validation for non-on-demand users.

A proper solution would be to keep track of which frame starts were sent and filter out mismatched ends on the client, similar to what zone markup does. But with zones, you have an object to store state in, and frame markers are just function calls. You would have to have some sort of map for this, since users can give you any number of pointers to track. This is problematic because you can't use the existing data structures, as all profiler allocations should go through rpmalloc, and you need to solve multithreading somehow, and that gets ugly pretty quickly.

Another solution would be to add a flag, similar to `m_ignoreMemFreeFaults`, which you can use to control whether the failure check should be performed, as in `Worker::ProcessMemFreeImpl()`.

YaLTeR:
> Another solution would be to add a flag, similar to `m_ignoreMemFreeFaults`, which you can use to control whether the failure check should be performed, as in `Worker::ProcessMemFreeImpl()`.

Did this. The flag is set when it's on-demand mode.

PR #671, opened 2023-11-24
(Do not submit) - draft for adapting to capstone arm64 renaming

bjacob:
FYI @wolfpld: Capstone has just done a breaking renaming: ARM64 --> AArch64 (https://github.com/capstone-engine/capstone/pull/2026)

This PR is a good-enough local patch for anyone trying to build against latest Capstone. It uses Capstone's own provided compatibility macros to resolve to either the old or the new identifiers.

The problem which IIUC makes this PR not good to merge is that by relying on these Capstone macros... it probably doesn't actually support older Capstone versions, which presumably don't have these macros.

You probably need to resolve that by forking these macros on the Tracy side...

### DIFF
--- a/server/TracySourceView.cpp
+++ b/server/TracySourceView.cpp
@@ -711,7 +711,7 @@ bool SourceView::Disassemble( uint64_t symAddr, const Worker& worker )
         rval = cs_open( CS_ARCH_ARM, CS_MODE_ARM, &handle );
         break;
     case CpuArchArm64:
-        rval = cs_open( CS_ARCH_ARM64, CS_MODE_ARM, &handle );
+        rval = cs_open( CS_AARCH64pre(CS_ARCH_), CS_MODE_ARM, &handle );
         break;
     default:
         assert( false );
@@ -776,9 +776,9 @@ bool SourceView::Disassemble( uint64_t symAddr, const Worker& worker )
                     }
                     break;
                 case CpuArchArm64:
-                    if( detail.arm64.op_count == 1 && detail.arm64.operands[0].type == ARM64_OP_IMM )
+                    if( detail.CS_aarch64().op_count == 1 && detail.CS_aarch64().operands[0].type == CS_AARCH64(_OP_IMM) )
                     {
-                        jumpAddr = (uint64_t)detail.arm64.operands[0].imm;
+                        jumpAddr = (uint64_t)detail.CS_aarch64().operands[0].imm;
                     }
                     break;
                 default:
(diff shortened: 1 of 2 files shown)


hanhanW:
oooh! I also hit the same issue! It seems that I should maybe just build capstone v4, but not `next`? Let me give it a shot..

hanhanW:
`v5` works for me, I will use `v5` for now

Rot127:
@hanhanW Mind that `v5` is pretty out of date already. It is based on LLVM 7, misses some Aarch64 extensions and has some bugs. We have a [guide](https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md) regarding the upcoming v6 if you want to take a look.

hanhanW:
> @hanhanW Mind that `v5` is pretty out of date already. It is based on LLVM 7, misses some Aarch64 extensions and has some bugs. We have a [guide](https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md) regarding the upcoming v6 if you want to take a look.

I'm good with `v5` for now because I don't need to jump into asm level at this moment. I mainly wanted to address compilation issues when building tracy with IREE. Thanks for the heads-up!

bjacob:
@wolfpld, it seems that multiple people have hit this. Would you suggest merging some flavor of this (do you want to suggest changes to this draft) ?

bjacob:
> @bjacob Just stumbled on this. Capstone v5 and v4 have those macros now. See capstone.h in each branch.

I'm not sure that we can count on all users' package managers to have received the backport.

bjacob:
The guaranteed thing that we can do here is, as suggested in the PR description, fork those CS compatibility macros on the Tracy side.

wolfpld:
#707 should solve this.

bjacob:
> https://github.com/wolfpld/tracy/pull/707 should solve this.

From a cursory look --- it looks like this hardcodes Capstone 5.0.1 ? I actually need to use Capstone 6 to use recent ISA extensions. Doing so would require Tracy-side source changes as in the present PR, right ?

bjacob:
Looks like a good step though, from where the capstone 5->6 move could be made as a second step ?  Given the headache that capstone has been giving tracy users, I definitely welcome Tracy internalizing it.

wolfpld:
The key thing is that it hardcodes *a* version, so you have one target, instead of the crapshoot that is currently the case (the :hankey: prize goes to ubububu, which ships 3.x :exploding_head:).

bjacob:
Yup, exactly. Once https://github.com/wolfpld/tracy/pull/707 is merged, and we've rebased our downstream to work with that, i'll rebase the present PR into a capstone 5->6 bump (which you may or may not merge, but it's useful for me even if unmerged).

wolfpld:
FYI, CMake has been "merged".

ScottTodd:
> Yup, exactly. Once #707 is merged, and we've rebased our downstream to work with that, i'll rebase the present PR into a capstone 5->6 bump (which you may or may not merge, but it's useful for me even if unmerged).

I'm currently updating our (IREE's) downstream to use the CMake files now present here in Tracy upstream (we'll just use the upstream build and drop our custom CMake + custom instructions, unless there is some reason to keep them that I'm not seeing). @bjacob did you want to continue with a capstone 5 -> 6 update?

bjacob:
Yup, that's the plan laid out at https://github.com/iree-org/iree/issues/16777. The above https://github.com/wolfpld/tracy/pull/671#issuecomment-2016830130 implies that the first step in the tasklist there is already completed, and what you describe doing is the second step in that tasklist, so after that we can bump from Capstone 5 to 6, the third step in that tasklist.

ScottTodd:
I'm not sure why those steps are dependent on one another?

bjacob:
not a strict dependence at every step, just my suggestion how in what order to tackle things.

Rot127:
I am sorry, forgot to drop it here. But we have now a compatibility header now. It is still a PR until the other maintainers manage to look at it: https://github.com/capstone-engine/capstone/pull/2321

wolfpld:
Closing, as v6 is now on master.

PR #674, opened 2023-11-29
Add image cache to avoid calling dladdr() and add libbacktrace elf image list refresh

tiago-rodrigues:
On glibc platforms, add an image cache to avoid calling dladdr() when doing offline symbol resolution, as dladdr() can cause contention with other dl* calls (see: https://github.com/wolfpld/tracy/pull/665#issuecomment-1817363122).

Add support for libbacktrace to detect new elfs have been dynamically loaded after backtrace_initialize() has been called, and consider them for further symbol resolution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,7 @@ set_option(TRACY_NO_CRASH_HANDLER "Disable crash handling" OFF)
 set_option(TRACY_TIMER_FALLBACK "Use lower resolution timers" OFF)
 set_option(TRACE_CLIENT_LIBUNWIND_BACKTRACE "Use libunwind backtracing where supported" OFF)
 set_option(TRACY_SYMBOL_OFFLINE_RESOLVE "Instead of full runtime symbol resolution, only resolve the image path and offset to enable offline symbol resolution" OFF)
+set_option(TRACY_LIBBACKTRACE_ELF_DYNLOAD_SUPPORT "Enable libbacktrace to support dynamically loaded elfs in symbol resolution resolution after the first symbol resolve operation" OFF)
 
 if(NOT TRACY_STATIC)
     target_compile_definitions(TracyClient PRIVATE TRACY_EXPORTS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,7 +86,6 @@ set_option(TRACY_NO_CRASH_HANDLER "Disable crash handling" OFF)
 set_option(TRACY_TIMER_FALLBACK "Use lower resolution timers" OFF)
 set_option(TRACE_CLIENT_LIBUNWIND_BACKTRACE "Use libunwind backtracing where supported" OFF)
 set_option(TRACY_SYMBOL_OFFLINE_RESOLVE "Instead of full runtime symbol resolution, only resolve the image path and offset to enable offline symbol resolution" OFF)
-set_option(TRACY_ENABLE_IMAGE_CACHE "On glibc platforms, when doing offline symbol resolution, use an cache to determine the image path and offset instead of dladdr()" OFF)
 
 if(NOT TRACY_STATIC)
     target_compile_definitions(TracyClient PRIVATE TRACY_EXPORTS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,7 @@ set_option(TRACY_NO_CRASH_HANDLER "Disable crash handling" OFF)
 set_option(TRACY_TIMER_FALLBACK "Use lower resolution timers" OFF)
 set_option(TRACE_CLIENT_LIBUNWIND_BACKTRACE "Use libunwind backtracing where supported" OFF)
 set_option(TRACY_SYMBOL_OFFLINE_RESOLVE "Instead of full runtime symbol resolution, only resolve the image path and offset to enable offline symbol resolution" OFF)
+set_option(TRACY_ENABLE_IMAGE_CACHE "On glibc platforms, when doing offline symbol resolution, use an cache to determine the image path and offset instead of dladdr()" OFF)
 
 if(NOT TRACY_STATIC)
     target_compile_definitions(TracyClient PRIVATE TRACY_EXPORTS)
--- a/public/client/TracyCallstack.cpp
+++ b/public/client/TracyCallstack.cpp
@@ -90,6 +90,124 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
 #endif
 #endif
 
+#if defined( TRACY_ENABLE_IMAGE_CACHE) && (TRACY_HAS_CALLSTACK == 3)
+#   define TRACY_USE_IMAGE_CACHE
+#   include <link.h>
+#endif
+
+#ifdef TRACY_USE_IMAGE_CACHE
+// when we have access to dl_iterate_phdr(), we can build a cache of address ranges to image paths
+// so we can quickly determine which image an address falls into.
+// We refresh this cache only when we hit an address that doesn't fall into any known range.
+class ImageCache
+{
+public:
+    struct ImageEntry
+    {
+        ImageEntry( void* startAddress, void* endAddress, const char* name )
+        : m_startAddress( startAddress ), m_endAddress( endAddress ), m_name( name ) {}
+
+        void* m_startAddress;
+        void* m_endAddress;
+        const char* m_name;
+    };
+
+    ImageCache()
+    {
+        m_images = (tracy::FastVector<ImageEntry>*)tracy::tracy_malloc( sizeof( tracy::FastVector<ImageEntry> ) );
+        new(m_images) tracy::FastVector<ImageEntry>( 512 );
+
+        Refresh();
+    }
+    const ImageEntry* GetImageForAddress( void* address ) 
+    {
+        const ImageEntry* entry = GetImageEntryForAddress( address );
+        if( !entry )
+        {
+            //printf("* addr not found: %p (%d entries) refreshing (m_numberOfRefreshes: %d)\n",
+            //       address, m_images->size(), m_numberOfRefreshes);
+            Refresh();
+            return GetImageEntryForAddress( address );
+        }
+        return entry;
+    }
+
+    typedef void (*RefreshCallback)();
+    void SetOnRefreshCallback( RefreshCallback callback )
+    {
+        m_onRefreshCallback = callback;
+    }
+
+private:
+    tracy::FastVector<ImageEntry>* m_images;
+    const char* m_imageName = nullptr;
+    RefreshCallback m_onRefreshCallback;
+    uint32_t m_numberOfRefreshes = 0;
+
+    static int Callback( struct dl_phdr_info* info, size_t size, void* data ) 
+    {
+        ImageCache* cache = reinterpret_cast<ImageCache*>( data );
+
+        ImageEntry* image = cache->m_images->push_next();
+        image->m_startAddress = reinterpret_cast<void*>( info->dlpi_addr );
+        const uint32_t headerCount = info->dlpi_phnum;
+        assert( headerCount > 0);
+        image->m_endAddress = reinterpret_cast<void*>( info->dlpi_addr + 
+            info->dlpi_phdr[info->dlpi_phnum - 1].p_vaddr + info->dlpi_phdr[info->dlpi_phnum - 1].p_memsz);
+
+        // the base executable name isn't provided when iterating with dl_iterate_phdr,
+        // so we must get it in an alternative way and cache it
+        if( info->dlpi_name && info->dlpi_name[0] != '\0' )
+        {
+            image->m_name = info->dlpi_name;
+        }
+        else
+        {
+            if( !cache->m_imageName )
+            {
+                Dl_info dlInfo;
+                if( dladdr( (void *)info->dlpi_addr, &dlInfo ) )
+                {
+                    cache->m_imageName = dlInfo.dli_fname;
+                }
+            }
+            image->m_name = cache->m_imageName;
+        }
+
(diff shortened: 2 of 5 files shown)


YaLTeR:
Gave this a try and it seems to work very well! I am able to record GNOME Shell with offline symbol resolving without much extra lag seemingly.

public/client/TracyCallstack.cpp
```diff
@@ -90,6 +90,124 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
...
+        if( it != m_images->end() && address < it->m_endAddress )
+        {
+            return &(*it);
+        }
+        return nullptr;
+    }
+};
+#endif //#ifdef TRACY_USE_IMAGE_CACHE
+
 namespace tracy
```
wolfpld, line 211: Is there a reason for keeping the above code outside the `tracy` namespace?

public/client/TracyCallstack.cpp
```diff
@@ -90,6 +90,124 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
...
+
+        Refresh();
+    }
+    const ImageEntry* GetImageForAddress( void* address ) 
+    {
+        const ImageEntry* entry = GetImageEntryForAddress( address );
+        if( !entry )
+        {
+            //printf("* addr not found: %p (%d entries) refreshing (m_numberOfRefreshes: %d)\n",
+            //       address, m_images->size(), m_numberOfRefreshes);
```
wolfpld, line 128: Remove debug code or convert to `TracyDebug()`.

public/client/TracyCallstack.cpp
```diff
@@ -90,6 +90,124 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
...
+    }
+
+    const ImageEntry* GetImageEntryForAddress( void* address ) const 
+    {
+        auto it = std::lower_bound( m_images->begin(), m_images->end(), address, 
+            []( const ImageEntry& lhs, const void* rhs ) { return lhs.m_startAddress > rhs; } );
+
+        if( it != m_images->end() && address < it->m_endAddress )
+        {
+            return &(*it);
```
wolfpld, line 204: Why not `return it`?

CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ set_option(TRACY_NO_CRASH_HANDLER "Disable crash handling" OFF)
 set_option(TRACY_TIMER_FALLBACK "Use lower resolution timers" OFF)
 set_option(TRACE_CLIENT_LIBUNWIND_BACKTRACE "Use libunwind backtracing where supported" OFF)
 set_option(TRACY_SYMBOL_OFFLINE_RESOLVE "Instead of full runtime symbol resolution, only resolve the image path and offset to enable offline symbol resolution" OFF)
+set_option(TRACY_ENABLE_IMAGE_CACHE "On glibc platforms, when doing offline symbol resolution, use an cache to determine the image path and offset instead of dladdr()" OFF)
```
wolfpld, line 89: What is the reason for keeping this as an option?

public/client/TracyCallstack.cpp
tiago-rodrigues: not really, moved it under

CMakeLists.txt
tiago-rodrigues: Hum, that is strange. I'll unoptimize the code and do some stepping again  in my case to make sure what I was seeing was actually correct (I'll hopefully get to it later today). But after the first symbol resolve was hit, I was missing symbol resolving for shared libraries loaded after. Looking through the libbacktrace code I could see it was using backtrace_state::fileline_fn as the control value to already called backtrace_initialize() (which was what was responsible for the shared library indexing), and that was set to something non-null after the call. That doesn't appear to be the case for you though, maybe I misinterpreted the results in the debugger with optimized code and something else was causing the issues.

tiago-rodrigues: right, confirmed it's the same with HEAD client code. 
So I really only get elf_add() called a bunch of times, the first time it calls backtrace_initialize() and it indexes the .so with dl_iterate_phdr(). backtrace_initialize() get's called exactly once for me, on the first call to backtrace_pcinfo(), and I get many dynamic loaded of .so after that which are not included in the libbacktrace internal state as I nothing else called elf_add() for me after that.
In your case, is it not setting backtrace_state::fileline_fn or maybe something is resetting it?

wolfpld: This is what I get with printf debugging:

```
backtrace_initialize: /proc/self/exe
elf_add: /proc/self/exe
elf_add: /proc/self/exe
elf_add: /usr/lib/libdebuginfod.so.1
elf_add:
elf_add: /usr/lib/libstdc++.so.6
elf_add: /usr/lib/libm.so.6
elf_add:
elf_add: /usr/lib/libgcc_s.so.1
elf_add: /usr/lib/libc.so.6
elf_add:
elf_add: /usr/lib/libcurl.so.4
elf_add:
elf_add: /usr/lib/libelf.so.1
elf_add:
elf_add: /lib64/ld-linux-x86-64.so.2
elf_add:
elf_add: /usr/lib/libnghttp2.so.14
elf_add:
elf_add: /usr/lib/libidn2.so.0
elf_add: /usr/lib/libssh2.so.1
elf_add:
elf_add: /usr/lib/libpsl.so.5
elf_add: /usr/lib/libssl.so.3
elf_add:
elf_add: /usr/lib/libcrypto.so.3
elf_add:
elf_add: /usr/lib/libgssapi_krb5.so.2
elf_add:
elf_add: /usr/lib/libzstd.so.1
elf_add:
elf_add: /usr/lib/libbrotlidec.so.1
elf_add:
elf_add: /usr/lib/libz.so.1
elf_add:
elf_add: /usr/lib/libunistring.so.5
elf_add: /usr/lib/libkrb5.so.3
elf_add:
elf_add: /usr/lib/libk5crypto.so.3
elf_add:
elf_add: /usr/lib/libcom_err.so.2
elf_add:
elf_add: /usr/lib/libkrb5support.so.0
elf_add:
elf_add: /usr/lib/libkeyutils.so.1
elf_add:
elf_add: /usr/lib/libresolv.so.2
elf_add:
elf_add: /usr/lib/libbrotlicommon.so.1
elf_add:
```

So as I understand it, the problem you have is that shared objects that are dlopened at a later time are not visible because libbacktrace has already collected a list of shared objects that the executable is linking to and will not look for more.

This feels like something libbacktrace should handle itself (in an ideal world). You already know when you want to reset libbacktrace state. Have you tried running `dl_iterate_phdr` to call `phdr_callback` instead?

tiago-rodrigues: > So as I understand it, the problem you have is that shared objects that are dlopened at a later time are not visible because libbacktrace has already collected a list of shared objects that the executable is linking to and will not look for more

exactly.

> This feels like something libbacktrace should handle itself (in an ideal world).

I guess before I started looking more closely at the code I didn't realize you had already modified that libbacktrace code quite a bit and it wasn't just a dump from upstream. 

> you already know when you want to reset libbacktrace state. Have you tried running dl_iterate_phdr to call phdr_callback instead?

Yes, libacktrace code is already doing pretty much what I was doing once, so I can try to expose an update method I can call from outside - I'll try to do this this week and update this PR.

wolfpld: > I guess before I started looking more closely at the code I didn't realize you had already modified that libbacktrace code quite a bit and it wasn't just a dump from upstream.

There have been some changes that add some sorely missing features, but there have been no substantial changes. The changes to `dl_iterate_phdr` are technical: 7e8961d2fc3

wolfpld: > Yes, libacktrace code is already doing pretty much what I was doing once, so I can try to expose an update method I can call from outside - I'll try to do this this week and update this PR.

Please check if it would be feasible to implement this update functionality within libbacktrace itself.

wolfpld: > The changes to dl_iterate_phdr are technical: https://github.com/wolfpld/tracy/commit/7e8961d2fc3a8883ba61738b201d04cddb15b795

I think this was needed because `elf_add` can now call debuginfod, which might issue network requests, and this didn't play well with the requirements of `dl_iterate_phdr`.

tiago-rodrigues: >Please check if it would be feasible to implement this update functionality within libbacktrace itself.

I made an attempt at it in the latest commit

public/libbacktrace/elf.cpp
wolfpld, line 1: This whole file is marked as changed.

tiago-rodrigues: sorry line endings shenanigans - should be fixed now.
The indentation in those libbacktrace files is also somewhat "special"

public/client/TracyCallstack.cpp
```diff
@@ -90,9 +90,111 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
...
+            if( !cache->m_imageName )
+            {
+                Dl_info dlInfo;
+                if( dladdr( (void *)info->dlpi_addr, &dlInfo ) )
+                {
+                    cache->m_imageName = dlInfo.dli_fname;
+                }
+            }
+            image->m_name = cache->m_imageName;
+        }
```
wolfpld, line 169: Is `image->m_name` valid after this function returns? My guess would be that the `info` argument is only valid inside this function, and the `dlInfo` provided filename will be invalidated after the next call to `dladdr`.

public/client/TracyCallstack.cpp
```diff
@@ -90,9 +90,111 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
...
+    {
+        m_images->clear();
+
+        dl_iterate_phdr( Callback, this );
+        
+        std::sort( m_images->begin(), m_images->end(), 
+            []( const ImageEntry& lhs, const ImageEntry& rhs ) { return lhs.m_startAddress > rhs.m_startAddress; } );
+    }
+
+    const ImageEntry* GetImageEntryForAddress( void* address ) const
```
wolfpld, line 184: Rename to `GetImageForAddressImpl`.

public/client/TracyCallstack.cpp
```diff
@@ -90,9 +90,111 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
...
+        new(m_images) tracy::FastVector<ImageEntry>( 512 );
+
+        Refresh();
+    }
+    const ImageEntry* GetImageForAddress( void* address ) 
+    {
+        const ImageEntry* entry = GetImageEntryForAddress( address );
+        if( !entry )
+        {
+            Refresh();
```
wolfpld, line 130: Is there a mechanism that will prevent constant reloads when a large set of unmapped addresses is queried?

public/libbacktrace/elf.cpp
```diff
@@ -7366,6 +7393,12 @@ phdr_callback_mock (struct dl_phdr_info *info, size_t size ATTRIBUTE_UNUSED,
   }
   else ptr->dlpi_name = nullptr;
   ptr->dlpi_addr = info->dlpi_addr;
+
+  // calculate the address range so we can quickly determine is a PC is within the range of this image
```
wolfpld, line 7397: Typo.

public/client/TracyCallstack.cpp
tiago-rodrigues: good point - fixed.

public/client/TracyCallstack.cpp
tiago-rodrigues: There isn't one atm. The expectation would be that it wouldn't actually be called that often in real scenarios. In my case with a few hundred dynamic loaded shared libs, and instrumenting all memory allocations with a callstack captures I only get it called ~5 times until it actually has the full map as symbol resolution is relatively slow in itself.
What would be your suggestion, a "min ms between refreshes" and either skip (and return unknown symbols for that period) or "stall" the resolution thread for a period or time?

wolfpld: Stalling to allow batch processing of a larger set may be a solution. But I was more interested in whether you have considered this case. I don't think this is immediately actionable.

public/client/TracyCallstack.cpp
wolfpld: Where is this memory released?

public/client/TracyCallstack.cpp
```diff
@@ -90,9 +90,122 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
...
+        const uint32_t headerCount = info->dlpi_phnum;
+        assert( headerCount > 0);
+        image->m_endAddress = reinterpret_cast<void*>( info->dlpi_addr + 
+            info->dlpi_phdr[info->dlpi_phnum - 1].p_vaddr + info->dlpi_phdr[info->dlpi_phnum - 1].p_memsz);
+
+        // the base executable name isn't provided when iterating with dl_iterate_phdr,
+        // so we must get it in an alternative way and cache it
+        if( info->dlpi_name && info->dlpi_name[0] != '\0' )
+        {
+            image->m_name = info->dlpi_name;
```
wolfpld, line 165: What about this assignment?

tiago-rodrigues: fair point - copied it there as well.
I now free and rebuild images names with every refresh for simplicity - would probably be better if I just recreated the new image entries, although it would require searching and comparing the paths to be sure so not sure it's worth it.

YaLTeR:
I'm testing libunwind bt + dynload and I'm getting funny strings in the backtrace occasionally, could be memory corruption?

image

YaLTeR:
Okay, I get this without dynload too, so I'll make an issue instead.

YaLTeR:
Made an issue at #684; I somewhat suspect the latter changes in this PR though, because when I tested it in the beginning it seemed to work fine.

PR #682, opened 2023-12-11
Add tracy client address complile flag

amschnorr:
Support passing tracy client address via compile flag. Adds convenience for compiling tracy on multiple targets that have different network requirements. (In this case using QNX so can't use the default 255.255.255.255 as @michaeldleslie helped point out). 

Can add documentation in the appropriate spot.

### DIFF
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -503,7 +503,7 @@ The program name that is sent out in the broadcast messages can be customized by
 
 By default, the Tracy client will listen on all network interfaces. If you want to restrict it to only listening on the localhost interface, define the \texttt{TRACY\_ONLY\_LOCALHOST} macro at compile-time, or set the \texttt{TRACY\_ONLY\_LOCALHOST} environment variable to $1$ at runtime.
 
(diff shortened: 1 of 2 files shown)


wolfpld:
Please add docs and maybe update the QNX error right below?

manual/tracy.tex
```diff
@@ -503,6 +503,8 @@ \subsubsection{Client network interface}
 
 By default, the Tracy client will listen on all network interfaces. If you want to restrict it to only listening on the localhost interface, define the \texttt{TRACY\_ONLY\_LOCALHOST} macro at compile-time, or set the \texttt{TRACY\_ONLY\_LOCALHOST} environment variable to $1$ at runtime.
 
+If you need to use a specific Tracy client address, such as QNX requires, define the \texttt{TRACY\_CLIENT\_ADDRESS} macro at compile-time as the desired string address, or set the \texttt{TRACY\_ONLY\_LOCALHOST} environment variable to the desired string address at runtime.
```
wolfpld, line 506: Should the env variable part be here?

amschnorr: Removed, thanks.

PR #690, opened 2023-12-17
Protect against crashes when MANUAL_LIFETIME is used

plekakis:
Most public API functions are now guarded against TracyIsStarted. This is always true when MANUAL_LIFETIME is not defined and true after StartupProfiler is called.
This does mean that no data recording is taking place until *after* Tracy has been initialised. It doesn't break existing behaviour, this is only a path taken when MANUAL_LIFETIME is defined.
In addition to the Scoped classes, the LockCtx is also guarded. This means that the mutexes won't be instrumented if at the time of their context allocation Tracy hasn't been initialised (this is expected in this case).

### DIFF
--- a/public/client/TracyLock.hpp
+++ b/public/client/TracyLock.hpp
@@ -16,7 +16,7 @@ class LockableCtx
 public:
     tracy_force_inline LockableCtx( const SourceLocationData* srcloc ) :
 #ifdef TRACY_ON_DEMAND
-        , m_lockCount( 0 )
+        m_lockCount( 0 )
         , m_active( false )
 #else
         m_active(TracyIsStarted)
@@ -260,7 +260,7 @@ public:
     tracy_force_inline SharedLockableCtx( const SourceLocationData* srcloc )
         :
 #ifdef TRACY_ON_DEMAND
-        , m_lockCount( 0 )
+        m_lockCount( 0 )
         , m_active( false )
 #else
         m_active(TracyIsStarted)
--- a/public/client/TracyLock.hpp
+++ b/public/client/TracyLock.hpp
@@ -22,7 +22,7 @@ public:
         m_active(TracyIsStarted)
 #endif
     {
-#if !TRACY_ON_DEMAND
+#ifndef TRACY_ON_DEMAND
         if (!m_active) return;
 #endif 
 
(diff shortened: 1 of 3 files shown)


public/client/TracyLock.hpp
```diff
@@ -14,13 +14,20 @@ namespace tracy
...
-        : m_id( GetLockCounter().fetch_add( 1, std::memory_order_relaxed ) )
+    tracy_force_inline LockableCtx( const SourceLocationData* srcloc ) :
 #ifdef TRACY_ON_DEMAND
         , m_lockCount( 0 )
         , m_active( false )
+#else
+        m_active(TracyIsStarted)
 #endif
     {
+#if !TRACY_ON_DEMAND
```
wolfpld, line 25: This is not a valid check if `TRACY_ON_DEMAND` is not defined.

plekakis: You're right, sorry I thought it was defaulting to 0 instead.
I will fix and re-commit.

wolfpld: You omitted this one.

wolfpld:
The issues I see with the current solution:
- It mixes two different changes.
- What is the expected usage pattern where the internal `TracyIsStarted` check is needed? If you manage the profiler lifetime yourself, you already know where you can and can't make the calls. Adding internal checks just makes things slower.
- A blanket check in `QueueSerialFinish` shows me that this is not a complete solution, but a stop gap measure. There are no use cases where you would call this function directly from your code.
- You are making wrong coding style changes for no reason.
- You are dealing with the `MANUAL_LIFETIME` case, but the logic changes you are introducing depend on `TRACY_ON_DEMAND`.
- You add `m_active` and related handling to a code path that doesn't currently need it, for no good reason.

plekakis:
Yes, I agree, let's close this PR. It needs more thought, for sure, however as discussed previously:
_"If you manage the profiler lifetime yourself, you already know where you can and can't make the calls"_
This is not correct when it comes to mutex instrumentation. For now I can simply use ON_DEMAND which suits my use-case well actually.

wolfpld:
> This is not correct when it comes to mutex instrumentation.

That's true. I have looked at how to solve this myself and it is already non-trivial in the on-demand case. I'm not sure a manual lifetime management solution would be viable here.

PR #706, opened 2024-01-01
fix import-fuchsia: handle string references properly

c-cube:
some sample traces using that:
- https://magic-trace.org/v31.0-3d83af17c/assets/c-demo.fxt.gz
- https://magic-trace.org/v31.0-3d83af17c/assets/c-hello-world.fxt.gz
- https://magic-trace.org/v31.0-3d83af17c/assets/ocaml-hello-world.fxt.gz

### DIFF
--- a/import-fuchsia/src/import-fuchsia.cpp
+++ b/import-fuchsia/src/import-fuchsia.cpp
@@ -1,6 +1,7 @@
 #include <cinttypes>
 #include <cstddef>
 #include <cstdint>
+#include <exception>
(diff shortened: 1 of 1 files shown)


c-cube:
I talked to an expert and got some more sample fuchsia traces (generated by another tool this time, namely, https://magic-trace.org/). They are now correctly parsed by the import tool, names and all. I also added a basic sanity check by checking the magic number at the beginning instead of 
just skipping it.

PR #716, opened 2024-01-24
This change makes the crash handler only install when tracy is connected.

gedalia:
This avoids the issue that an end user has their own crash handler which gets ignored when tracy is compiled in, even if tracy isn't connected to capture the crash.

### DIFF
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -1440,6 +1440,52 @@ Profiler::Profiler()
 #endif
 }
 
+void Profiler::InstallCrashHandler()
+{
+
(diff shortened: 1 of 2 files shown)


public/client/TracyProfiler.cpp
```diff
@@ -1480,26 +1526,6 @@ void Profiler::SpawnWorkerThreads()
     m_exceptionHandler = AddVectoredExceptionHandler( 1, CrashFilter );
```
wolfpld, line 1526: You missed this.

gedalia: oops, sorry just testing on linux, hopefully next commit fixes.

PR #744, opened 2024-03-07
Changes logic for only (in)active allocations frame tree:

citron0xa9:
Swaps logic to:
 - skip active allocations if only inactive should be displayed
  - skip inactive allocations if only active should be displayed


citron0xa9:
Sorry I did forget to merge the latest commits and did not see that this was already resolved.

PR #746, opened 2024-03-11
Expose lock API to TracyC.h

Mulej:
This exposes the lock API to the C headers, allowing it to be used in non-C++ languages.


wolfpld:
I will need some documentation on how this can be used.

Mulej:
I will add a sub section on marking locks with it under the Marking locks section.

PR #747, opened 2024-03-12
Adding a copy to CSV button to the statistics view.

benvanik:
This copies the visible rows of either the instrumentation or GPU statistics view to a CSV string matching a subset of the csvexport format:
image

I noticed us sharing screenshots of the statistics view since being able the use the filter/limits of the statistics view to narrow down the zones of interest in the UI is really speedy - but screenshots aren't super usable :)

### DIFF
--- a/server/TracyView_Statistics.cpp
+++ b/server/TracyView_Statistics.cpp
@@ -1,3 +1,5 @@
+#include <sstream>
+
 #include "TracyFilesystem.hpp"
 #include "TracyImGui.hpp"
 #include "TracyPrint.hpp"
@@ -85,6 +87,7 @@ void View::DrawStatistics()
 
     Vector<SrcLocZonesSlim> srcloc;
 
+    bool copySrclocsToClipboard = false;
     if( m_statMode == 0 )
     {
         if( !m_worker.AreSourceLocationZonesReady() )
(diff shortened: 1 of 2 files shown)


wolfpld:
Can you move the button to the left, to match the callstack window?

obraz

server/TracyView_Statistics.cpp
```diff
@@ -440,6 +445,8 @@ void View::DrawStatistics()
         ImGui::Spacing();
         ImGui::SameLine();
         TextFocused( "Visible zones:", RealToString( srcloc.size() ) );
+        ImGui::SameLine();
+        copySrclocsToClipboard = ClipboardButton();
```
wolfpld, line 449: Try adding some spacing or a vertical separator. This will help separate things more clearly.

benvanik:
To the left of which element? It doesn't work with sampling and if all the way on the left toggling between Instrumentation/GPU/Sampling would cause everything to shift when the button was hidden in Sampling mode.

It only copies the visible zones so I put it adjacent to "visible zones" ala the copy buttons in other dialogs: 
image

Happy to move it wherever, though, just let me know!

wolfpld:
Ok, makes sense.

PR #748, opened 2024-03-12
Adding ctrl-r as a hotkey for reconnecting from the main profiler view.

benvanik:
This is effectively the same as clicking on the connection status button, clicking discard, and then clicking reconnect.


wolfpld:
Please make this ctrl+shift+alt+r, if there is to be no confirmation about this potentially destructive action.

benvanik:
I'm not familiar with latex - getting this:
```
./tracy.tex:2394: Undefined control sequence.
\CurrentMenuElement ->\alt 
```
Is there someplace I need to add that?

Issue #471, opened 2022-09-15
Better support for cluster environments


alexanderbock:
Totally understand if this is out of scope and a pretty niche usecase.

In our institution we are having a planetarium that is running the same instance 6 times in a networked environment.  In the past I have used Tracy in this environment by starting the GUI 7 times and connecting remotely to all instances manually.  It would be really neat to be able to connect to all of the clients from a single GUI and possibly also align the timelines from all of the instances and show the places where one of the instances takes longer to execute a function, for example.

Just to be clear, this would be N instances of the _same_ executable and they should always go through the same function calls and where they disagree is where the interesting stuff happens.

GCCFeli:
I'm in a slightly different situation. I'm running several game services in cluster and each service has different functionality. It would be great if all services(clients) could connect to a single GUI and the timelines are aligned.

PeterTh:
Cluster tooling is an entirely different can of worms, but FWIW we'd also be very interested in (even just basic) support for this use case.

GCCFeli:
> Cluster tooling is an entirely different can of worms, but FWIW we'd also be very interested in (even just basic) support for this use case. (Where "basic support" would probably mean ingesting data from several processes and aligning the times)

Ingesting data from several processes and aligning the times are enough for my case. For now I'm hacking this by a simple proxy which connected by cluster processes, act as the only client to tracy.

wolfpld:
Making a proxy that would mux multiple clients would be a preferred solution here. To properly handle thread identifiers, which may be duplicated across different processes, you may use the already existing encoding:

https://github.com/wolfpld/tracy/blob/e1395f5a5370eff8fbf24bf450b6a934e6597197/import-chrome/src/import-chrome.cpp#L143-L183

You can see how this works in https://github.com/wolfpld/tracy/issues/213#issuecomment-841641885.

In 0.9 there were many changes in how the timeline items are handled, which is not really visible to users right now. Each track displayed on the timeline is now an instance of https://github.com/wolfpld/tracy/blob/master/server/TracyTimelineItem.hpp and the management of these items is now well defined in https://github.com/wolfpld/tracy/blob/master/server/TracyTimelineController.hpp, instead of the mess it was before. The takeaway here is that it should be now relatively easy to rearrange the threads, so that threads originating from the same process are next to each other, or to add different colorings to thread backgrounds, etc.

jamesfmilne:
[...]

PR #766, opened 2024-04-04
Implements experimental tracy client multiplexer

cipharius:
This is a proof of concept implementation of tracy multiplexer which allows tracing multiple Tracy clients in realtime. It intends to resolve issue #471.

The multiplexer works by listening for client broadcast messages and automatically establishing connection with Tracy clients. Once connection is established between multiplexer and a Tracy server, all client events are fused into one event stream and passed on to the server.

Usage:
* Launch the multiplexer
* Launch at least one client
* With Tracy profiler connect to multiplexer over TCP port 8085
* During profiling session, new Tracy clients will be automatically adopted and included in the timeline

Some implementation details:
* The multiplexer steals first client's welcome message and uses it to establish connection with the Tracy server;
* Different client events are combined at ThreadContext boundaries;
* Server requests are replicated to all connected clients, once all clients have responded to the request the most appropriate response is picked and forwarded to the Tracy server. This part is lacking in this proof of concept and works best with 3+ clients, because in that case it's possible to find which response differs from rest and use that;
* Time is normalised and adjusted within multiplexer in order to have a coherent timeline in the Tracy profiler;

This proof of concept has some fragile parts that would be challenging to keep up to date as the profiler is updated. To address this, multiplexer, TracyWorker and TracyProfiler should be refactored to increase procedure reuse. For example I had to reimplement procedure that computes event packet size and had to manually sift through all events that contain time.

### DIFF
--- a/multiplex/CMakeLists.txt
+++ b/multiplex/CMakeLists.txt
@@ -1,0 +1,24 @@
+cmake_minimum_required(VERSION 3.16)
+
+option(NO_ISA_EXTENSIONS "Disable ISA extensions (don't pass -march=native or -mcpu=native to the compiler)" OFF)
+
+include(${CMAKE_CURRENT_LIST_DIR}/../cmake/version.cmake)
+
+set(CMAKE_CXX_STANDARD 20)
+
+project(
+    tracy-multiplex
+    LANGUAGES ASM C CXX
+    VERSION ${TRACY_VERSION_STRING}
+)
+
+include(${CMAKE_CURRENT_LIST_DIR}/../cmake/config.cmake)
+include(${CMAKE_CURRENT_LIST_DIR}/../cmake/vendor.cmake)
+include(${CMAKE_CURRENT_LIST_DIR}/../cmake/server.cmake)
+
+set(PROGRAM_FILES
+    src/multiplex.cpp
+)
+
+add_executable(${PROJECT_NAME} ${PROGRAM_FILES} ${COMMON_FILES} ${SERVER_FILES})
+target_link_libraries(${PROJECT_NAME} PRIVATE TracyServer TracyGetOpt)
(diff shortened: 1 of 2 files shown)


Arpafaucon:
Hi ! I set up the env today but I cannot reproduce your example :/ When spawning the different processes in the order you specify, I get the following error

```
❯ ./multiplex/build/tracy-multiplex
Starting Tracy multiplexer on port 8085
Listening for client UDP broadcast messages...
Error: Multiplexer has no welcome message, launch the first client
[... repeated multiple times]
Error: Multiplexer has no welcome message, launch the first client
Connecting to a client:
	Name: tracy-test
	ID: 34729646139584
	Address: 192.168.56.32
	Port: 8086
Welcome message:
	timerMul: 0.434029
	initBegin: 13107256857726
	initEnd: 13107820545094
	delay: 191
	resolution: 56
	epoch: 1719930553
	exectime: 1719930161
	pid: 89027
	samplingPeriod: 100000
	flags: 4
	cpuArch: 2
	cpuId: 34540
Connected to client successfuly

Got supposedly weird type 59
tracy-multiplex: /home/gregoire.roussel/dev/perf/tracy/multiplex/src/multiplex.cpp:736: bool handle_client_response(ClientConnection &, const tracy::QueueItem *, uint32_t): Assertion `ev->hdr.idx >= (int)tracy::QueueType::StringData' failed.
[1]    88859 IOT instruction (core dumped)  ./multiplex/build/tracy-multiplex
```

the assertion is in the `default` case of your big `switch` handling client responses.

Notes 
- the `Got supposedly weird type 59` is from me, to understand what's the `hdr.type` that's causing the failed assertion.
- `59` is a `tracy::QueueType::ThreadContext`;  I guess the `assert` documents that we should not expect `ThreadContext` messages there ?

Arpafaucon:
> @Arpafaucon Are you testing with tracy GUI and client compiled from this commit's checkout?
> 
> It will most likely fail if you try to use it with too new tracy version, due to how coupled the code is with the protocol.

yes, I compiled `capture`, `multiplex` and my test program from your commit 57fe7556ec68587acb62aca40b7865647eaec9f3

PR #772, opened 2024-04-13
Server cmake fix : No need for ASM detection

Lectem:
Since we do not have asm sources in the project directly, there is no need to list it in `project`. This otherwise causes CMake to look for an assembler which may not be available on Windows with a default VS2022 install.

> -- Building for: Visual Studio 17 2022
> Parsing public/common/TracyVersion.hpp file
> VERSION 0.10.0
> -- Selecting Windows SDK version 10.0.22621.0 to target Windows 10.0.19045.
> -- The ASM compiler identification is unknown
> -- Didn't find assembler
> -- The C compiler identification is MSVC 19.39.33521.0
> -- The CXX compiler identification is MSVC 19.39.33521.0
> CMake Error at CMakeLists.txt:13 (project):
>   No CMAKE_ASM_COMPILER could be found.

Instead, dependencies should handle such things themselves if needed.

Also took this opportunity to set the default startup project under VS.


Lectem:
I stand corrected, I just found that the ZSTD sources in `vendor.cmake` do have assembly files.
```cmake
    # Assembly
    decompress/huf_decompress_amd64.S
```

And this builds fine on windows due to `ZSTD_ASM_SUPPORTED` being false on anything non-GNUC.

I'll try to draft something so that this is required only on platforms/compilers that can actually compile the GAS assembly.

wolfpld:
> I stand corrected, I just found that the ZSTD sources in `vendor.cmake` do have assembly files.

Yup.

> This otherwise causes CMake to look for an assembler which may not be available on Windows with a default VS2022 install.

Maybe a good question is what is a "default install"? I checked to see if things were working on a clean Windows + VS2022 install and everything was fine.

Lectem:
> Maybe a good question is what is a "default install"? I checked to see if things were working on a clean Windows + VS2022 install and everything was fine.

This is surprising, do you have NASM installed on the side perhaps ? what ASM executable does it pick ? (can be found in the cache as with variable `CMAKE_ASM_COMPILER`)

wolfpld:
> This is surprising, do you have NASM installed on the side perhaps ?

No, I don't really use that machine, so it's unlikely something got installed by accident.

> Or maybe there is a certain workload that needs to be selected in the VS installer.

I have the "classic C++ applications" selected, or whatever it is named in English.

> What ASM executable does it pick ? (can be found in the cache as with variable `CMAKE_ASM_COMPILER`)

`cl.exe`

Lectem:
Ok so it seems there's this note in the `project`/`enable_language` documentation...

> If enabling ASM, list it last so that CMake can check whether compilers for other languages like C work for assembly too.

I tried it on a sample project, and `project(sandbox LANGUAGES ASM C CXX)` gives me `No CMAKE_ASM_COMPILER could be found.` while `project(sandbox LANGUAGES C CXX ASM)` works.

Should I simply modify ASM to be last everywhere instead of enabling it only where needed like in https://github.com/wolfpld/tracy/pull/772/commits/7c3acf28bfc796c8a1d50b05917335235bb92a7d ?

wolfpld:
Reordering languages seems to be a simpler solution.

Lectem:
Should be better now !

wolfpld:
https://mastodon.gamedev.place/@wolfpld/112360012237378366

clement-siliceum:
Oh well, it won't annoy anyone anymore for sure now!

PR #786, opened 2024-05-07
Support adjusting plot height.

mcourteaux:
Plotting is useful, but once you add many, it starts taking up a lot of space rapidly. So I added a slider to shrink the plot heights.

image

### DIFF
--- a/profiler/src/profiler/TracyView_Plots.cpp
+++ b/profiler/src/profiler/TracyView_Plots.cpp
@@ -11,9 +11,6 @@
 namespace tracy
 {
 
-constexpr int PlotHeightPx = 100;
-
-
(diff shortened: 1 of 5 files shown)


wolfpld:
Please add save/load functionality in `TracyUserData.cpp`.

profiler/src/profiler/TracyView_Plots.cpp
```diff
@@ -24,7 +21,7 @@ bool View::DrawPlot( const TimelineContext& ctx, PlotData& plot, const std::vect
     const auto hover = ctx.hover;
     const auto ty = ctx.ty;
 
-    const auto PlotHeight = PlotHeightPx * GetScale();
+    const auto PlotHeight = GetViewData().plotHeight * GetScale();
```
wolfpld, line 24: You should be able to use `m_vd` here.

mcourteaux:
Taken care of :)

PR #793, opened 2024-05-17
Metal (Apple) GPU back-end for Tracy

slomp:
(I still need to update the manual, but I'm putting the code here for review to save some time).

The Metal back-end in Tracy operates differently than other GPU back-ends like Vulkan, Direct3D and OpenGL. Specifically, `TracyMetalZone()` must be placed around the site where a command encoder is created.

This is because not all hardware supports timestamps at command granularity, and can only provide timestamps around an entire command encoder. This accommodates for all tiers of hardware; in the future, variants of `TracyMetalZone()` will be added to support the habitual command-level granularity of Tracy GPU back-ends.

Metal also imposes a few restrictions that make the process of requesting and collecting queries more complicated in Tracy:
- timestamp query buffers are limited to 4096 queries (32KB, where each query is 8 bytes)
- when a timestamp query buffer is created, Metal initializes all timestamps with zeroes, and there's no way to reset them back to zero after timestamps get resolved; the only way to clear the timestamps is by allocating a new timestamp query buffer
- if a command encoder records no commands and its corresponding command buffer ends up committed to the command queue, Metal will "optimize-away" the encoder along with any timestamp queries associated with it (the timestamp will remain as zero and will never get resolved)

Because of the limitations above, two timestamp buffers are managed internally. Once one of the buffers fills up with requests, the second buffer can start serving new requests.

Once all requests in a buffer get resolved and collected, the entire buffer is discarded and a new one allocated for future requests. (Proper cycling through a ring buffer would require bookkeeping and completion handlers to collect only the known complete queries.)

In the current implementation, there is potential for a race condition when the buffer is discarded and reallocated. In practice, the race condition will never materialize so long as TracyMetalCollect() is called frequently to keep the amount of unresolved queries low.

Finally, there's a timeout mechanism during timestamp collection to detect "empty" command encoders and ensure progress.

### DIFF
--- a/public/common/TracyQueue.hpp
+++ b/public/common/TracyQueue.hpp
@@ -399,7 +399,8 @@ enum class GpuContextType : uint8_t
     Vulkan,
     OpenCL,
     Direct3D12,
-    Direct3D11
+    Direct3D11,
+    Metal
 };
 
 enum GpuContextFlags : uint8_t
--- a/public/tracy/TracyMetal.hmm
+++ b/public/tracy/TracyMetal.hmm
@@ -1,0 +1,364 @@
+#ifndef __TRACYMETAL_HMM__
+#define __TRACYMETAL_HMM__
+
+#ifndef TRACY_ENABLE
+
+#define TracyMetalContext(device,queue) nullptr
+#define TracyMetalDestroy(ctx)
+#define TracyMetalContextName(ctx, name, size)
+
+#define TracyMetalZone(ctx, name)
+#define TracyMetalZoneC(ctx, name, color)
+#define TracyMetalNamedZone(ctx, varname, name, active)
+#define TracyMetalNamedZoneC(ctx, varname, name, color, active)
+#define TracyMetalZoneTransient(ctx, varname, name, active)
+
+#define TracyMetalZoneS(ctx, name, depth)
+#define TracyMetalZoneCS(ctx, name, color, depth)
+#define TracyMetalNamedZoneS(ctx, varname, name, depth, active)
+#define TracyMetalNamedZoneCS(ctx, varname, name, color, depth, active)
+#define TracyMetalZoneTransientS(ctx, varname, name, depth, active)
+
+#define TracyMetalCollect(ctx)
+
+namespace tracy
+{
+class MetalZoneScope {};
+}
+
+using TracyMetalCtx = void*;
+
+#else
+
+#include <atomic>
+#include <assert.h>
+#include <stdlib.h>
+
+#include "Tracy.hpp"
+#include "../client/TracyProfiler.hpp"
+#include "../client/TracyCallstack.hpp"
+#include "../common/TracyAlign.hpp"
+#include "../common/TracyAlloc.hpp"
+
+// ok to import if in obj-c code
+#import <Metal/Metal.h>
+
+#define TracyMetalPanic(msg, ...) do { assert(false && "TracyMetal: " msg); TracyMessageLC("TracyMetal: " msg, tracy::Color::Red4); fprintf(stderr, "TracyMetal: %s\n", msg); __VA_ARGS__; } while(false);
+
+
+namespace tracy
+{
+
+class MetalCtx
+{
+    friend class MetalZoneScope;
+
+    enum { MaxQueries = 4 * 1024 };    // Metal: between 8 and 32768 _BYTES_...
+
+public:
+    MetalCtx(id<MTLDevice> device)
+        : m_device(device)
+    {
+        if (m_device == nil)
+        {
+            TracyMetalPanic("device is nil.", return);
+        }
+        if (![m_device supportsCounterSampling:MTLCounterSamplingPointAtDispatchBoundary])
+        {
+            TracyMetalPanic("timestamp sampling at compute dispatch boundary is not supported.", return);
+        }
+        if (![m_device supportsCounterSampling:MTLCounterSamplingPointAtDrawBoundary])
+        {
+            TracyMetalPanic("timestamp sampling at draw boundary is not supported.", return);
+        }
+        id<MTLCounterSet> timestampCounterSet = nil;
+        for (id<MTLCounterSet> counterSet in m_device.counterSets)
+        {
+            if ([counterSet.name isEqualToString:MTLCommonCounterSetTimestamp])
+            {
+                timestampCounterSet = counterSet;
+                break;
+            }
+        }
+        if (timestampCounterSet == nil)
+        {
+            TracyMetalPanic("timestamp counters are not supported on the platform.", return);
+        }
+        
+        MTLCounterSampleBufferDescriptor* sampleDescriptor = [[MTLCounterSampleBufferDescriptor alloc] init];
+        sampleDescriptor.counterSet = timestampCounterSet;
+        sampleDescriptor.sampleCount = MaxQueries;
+        sampleDescriptor.storageMode = MTLStorageModeShared;
+        sampleDescriptor.label = @"TracyMetalTimestampPool";
+        
+        NSError* error = nil;
+        id<MTLCounterSampleBuffer> counterSampleBuffer = [m_device newCounterSampleBufferWithDescriptor:sampleDescriptor error:&error];
+        if (error != nil)
+        {
+            NSLog(error.localizedDescription);
+            NSLog(error.localizedFailureReason);
+            TracyMetalPanic("unable to create sample buffer for timestamp counters.", return);
+        }
+        m_counterSampleBuffer = counterSampleBuffer;
+
+        MTLTimestamp cpuTimestamp = 0;
+        MTLTimestamp gpuTimestamp = 0;
+        float period = 1.0f;
+        [m_device sampleTimestamps:&cpuTimestamp gpuTimestamp:&gpuTimestamp];
+        
+        m_contextId = GetGpuCtxCounter().fetch_add(1);
+        
+        auto* item = Profiler::QueueSerial();
+        MemWrite(&item->hdr.type, QueueType::GpuNewContext);
+        MemWrite(&item->gpuNewContext.cpuTime, int64_t(cpuTimestamp));
+        MemWrite(&item->gpuNewContext.gpuTime, int64_t(gpuTimestamp));
+        MemWrite(&item->gpuNewContext.thread, uint32_t(0)); // #TODO: why not GetThreadHandle()?
+        MemWrite(&item->gpuNewContext.period, period);
+        MemWrite(&item->gpuNewContext.context, m_contextId);
+        MemWrite(&item->gpuNewContext.flags, GpuContextCalibration);
+        MemWrite(&item->gpuNewContext.type, GpuContextType::Metal);
+        Profiler::QueueSerialFinish();  // TODO: DeferItem() for TRACY_ON_DEMAND
+    }
+
+    ~MetalCtx()
+    {
+    }
+
+    void Name( const char* name, uint16_t len )
+    {
+        auto ptr = (char*)tracy_malloc( len );
+        memcpy( ptr, name, len );
+
+        auto item = Profiler::QueueSerial();
+        MemWrite( &item->hdr.type, QueueType::GpuContextName );
+        MemWrite( &item->gpuContextNameFat.context, m_contextId );
+        MemWrite( &item->gpuContextNameFat.ptr, (uint64_t)ptr );
+        MemWrite( &item->gpuContextNameFat.size, len );
+#ifdef TRACY_ON_DEMAND
+        GetProfiler().DeferItem( *item );
+#endif
+        Profiler::QueueSerialFinish();
+    }
+
+    bool Collect()
+    {
+        ZoneScopedC(Color::Red4);
+
+#ifdef TRACY_ON_DEMAND
+        if (!GetProfiler().IsConnected())
+        {
+            return true;
+        }
+#endif
+
+        // Only one thread is allowed to collect timestamps at any given time
+        // but there's no need to block contending threads
+        if (!m_collectionMutex.try_lock())
+        {
+            return true;
+        }
+
+        std::unique_lock lock (m_collectionMutex, std::adopt_lock);
+
+        uintptr_t begin = m_previousCheckpoint.load();
+        uintptr_t latestCheckpoint = m_queryCounter.load(); // TODO: MTLEvent? MTLFence?;
+        uint32_t count = RingCount(begin, latestCheckpoint);
+
+        if (count == 0)   // no pending timestamp queries
+        {
+            uintptr_t nextCheckpoint = m_queryCounter.load();
+            if (nextCheckpoint != latestCheckpoint)
+            {
+                // TODO: signal event / fence now?
+            }
+            return true;
+        }
+
+        if (count >= MaxQueries)
+        {
+            TracyMetalPanic("too many pending timestamp queries.", return false;);
+        }
+
+        NSRange range = { };
+        NSData* data = [m_counterSampleBuffer resolveCounterRange:range];
+        NSUInteger numResolvedTimestamps = data.length / sizeof(MTLCounterResultTimestamp);
+        MTLCounterResultTimestamp* timestamps = (MTLCounterResultTimestamp *)(data.bytes);
+        if (timestamps == nil)
+        {
+            TracyMetalPanic("unable to resolve timestamps.", return false;);
+        }
+
+        for (auto i = begin; i != latestCheckpoint; ++i)
+        {
+            uint32_t k = RingIndex(i);
+            MTLTimestamp& timestamp = timestamps[k].timestamp;
+            // TODO: check the value of timestamp: MTLCounterErrorValue, zero, or valid
+            if (timestamp == MTLCounterErrorValue)
+            {
+                break;
+            }
+            m_previousCheckpoint += 1;
+            auto* item = Profiler::QueueSerial();
+            MemWrite(&item->hdr.type, QueueType::GpuTime);
+            MemWrite(&item->gpuTime.gpuTime, timestamp);
+            MemWrite(&item->gpuTime.queryId, static_cast<uint16_t>(k));
+            MemWrite(&item->gpuTime.context, m_contextId);
+            Profiler::QueueSerialFinish();
+            timestamp = MTLCounterErrorValue;  // "reset" timestamp
+        }
+
+        //RecalibrateClocks();    // to account for drift
+
+        return true;
+    }
+
+private:
+    tracy_force_inline uint32_t RingIndex(uintptr_t index)
+    {
+        index %= MaxQueries;
+        return static_cast<uint32_t>(index);
+    }
+
+    tracy_force_inline uint32_t RingCount(uintptr_t begin, uintptr_t end)
+    {
+        // wrap-around safe: all unsigned
+        uintptr_t count = end - begin;
+        return static_cast<uint32_t>(count);
+    }
+
+    tracy_force_inline unsigned int NextQueryId()
+    {
+        auto id = m_queryCounter.fetch_add(1);
+        if (RingCount(m_previousCheckpoint, id) >= MaxQueries)
+        {
+            TracyMetalPanic("too many pending timestamp queries.");
+            // #TODO: return some sentinel value; ideally a "hidden" query index
+        }
+        return RingIndex(id);
+    }
+
+    tracy_force_inline uint8_t GetContextId() const
+    {
+        return m_contextId;
+    }
+
+    uint8_t m_contextId = 255;
+
+    id<MTLDevice> m_device = nil;
+    id<MTLCounterSampleBuffer> m_counterSampleBuffer = nil;
+    
+    using atomic_counter = std::atomic<uintptr_t>;
+    static_assert(atomic_counter::is_always_lock_free);
+    atomic_counter m_queryCounter = 0;
+
+    atomic_counter m_previousCheckpoint = 0;
+    atomic_counter::value_type m_nextCheckpoint = 0;
+    
+    std::mutex m_collectionMutex;
+};
+
+class MetalZoneScope
+{
+public:
+    tracy_force_inline MetalZoneScope( MetalCtx* ctx, id<MTLComputeCommandEncoder> cmdEncoder, const SourceLocationData* srcloc, bool is_active )
+#ifdef TRACY_ON_DEMAND
+        : m_active( is_active && GetProfiler().IsConnected() )
+#else
+        : m_active( is_active )
+#endif
+    {
+        if( !m_active ) return;
+        m_ctx = ctx;
+        m_cmdEncoder = cmdEncoder;
+
+        const auto queryId = ctx->NextQueryId();
+        [m_cmdEncoder sampleCountersInBuffer:m_ctx->m_counterSampleBuffer atSampleIndex:queryId withBarrier:YES];
+
+        auto* item = Profiler::QueueSerial();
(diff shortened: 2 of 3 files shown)


slomp:
@wolfpld I'd like to request reviews from @nosferalatu and @JamesMcCarthy44, but I can't seem to be able to add reviewers.

wolfpld:
I don't know how assigning reviewers work on Github. Mentioning people should be enough to get their attention.

slomp:
Also pinging @theblackunknown for a code review.

nosferalatu (approved):

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+        m_timestampRequestTime.resize(MaxQueries);
+
+        MTLTimestamp cpuTimestamp = 0;
+        MTLTimestamp gpuTimestamp = 0;
+        [m_device sampleTimestamps:&cpuTimestamp gpuTimestamp:&gpuTimestamp];
+        m_mostRecentTimestamp = gpuTimestamp;
+        TracyMetalDebug(1<<0, TracyMetalPanic(, "Calibration: CPU timestamp (Metal): %llu", cpuTimestamp));
+        TracyMetalDebug(1<<0, TracyMetalPanic(, "Calibration: GPU timestamp (Metal): %llu", gpuTimestamp));
+
+        cpuTimestamp = Profiler::GetTime();
```
theblackunknown, line 153: Is this expected that you ditch the CPU timestamp returned by the MTLDevice ? Is this for consistency with other Tracy event messages ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+        MemWrite(&item->hdr.type, QueueType::GpuNewContext);
+        MemWrite(&item->gpuNewContext.cpuTime, int64_t(cpuTimestamp));
+        MemWrite(&item->gpuNewContext.gpuTime, int64_t(gpuTimestamp));
+        MemWrite(&item->gpuNewContext.thread, uint32_t(0)); // #TODO: why not GetThreadHandle()?
+        MemWrite(&item->gpuNewContext.period, period);
+        MemWrite(&item->gpuNewContext.context, m_contextId);
+        //MemWrite(&item->gpuNewContext.flags, GpuContextCalibration);
+        MemWrite(&item->gpuNewContext.flags, GpuContextFlags(0));
+        MemWrite(&item->gpuNewContext.type, GpuContextType::Metal);
+        Profiler::QueueSerialFinish();  // TODO: DeferItem() for TRACY_ON_DEMAND
```
theblackunknown, line 170: Please do address the `TRACY_ON_DEMAND` it is rather easy to add here.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+
+        float period = 1.0f;
+        
+        m_contextId = GetGpuCtxCounter().fetch_add(1);
+        
+        auto* item = Profiler::QueueSerial();
+        MemWrite(&item->hdr.type, QueueType::GpuNewContext);
+        MemWrite(&item->gpuNewContext.cpuTime, int64_t(cpuTimestamp));
+        MemWrite(&item->gpuNewContext.gpuTime, int64_t(gpuTimestamp));
+        MemWrite(&item->gpuNewContext.thread, uint32_t(0)); // #TODO: why not GetThreadHandle()?
```
theblackunknown, line 164: Good question, I had the same while browsing the Tracy source code.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+        MemWrite(&item->gpuNewContext.period, period);
+        MemWrite(&item->gpuNewContext.context, m_contextId);
+        //MemWrite(&item->gpuNewContext.flags, GpuContextCalibration);
```
theblackunknown, line 167: Why do this means precisely for a GPU context to be calibrated for Tracy ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+
+    ~MetalCtx()
+    {
+        ZoneScopedNC("~TracyMetalCtx", tracy::Color::Red4);
+        ZoneValue(m_previousCheckpoint.load());
+        ZoneValue(m_queryCounter.load());
+        // collect the last remnants of Metal GPU activity...
+        // TODO: add a timeout to this loop?
+        while (m_previousCheckpoint.load() != m_queryCounter.load())
+            Collect();
```
theblackunknown, line 181: There is no such things in the Vulkan context, why is this needed here ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+        MemWrite(&item->gpuNewContext.flags, GpuContextFlags(0));
+        MemWrite(&item->gpuNewContext.type, GpuContextType::Metal);
+        Profiler::QueueSerialFinish();  // TODO: DeferItem() for TRACY_ON_DEMAND
+    }
+
+    ~MetalCtx()
+    {
+        ZoneScopedNC("~TracyMetalCtx", tracy::Color::Red4);
+        ZoneValue(m_previousCheckpoint.load());
+        ZoneValue(m_queryCounter.load());
```
theblackunknown, line 177: Are those debug information ? From a client perspective I would not be able to infer what they refer to.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+        memcpy( ptr, name, len );
+
+        auto item = Profiler::QueueSerial();
+        MemWrite( &item->hdr.type, QueueType::GpuContextName );
+        MemWrite( &item->gpuContextNameFat.context, m_contextId );
+        MemWrite( &item->gpuContextNameFat.ptr, (uint64_t)ptr );
+        MemWrite( &item->gpuContextNameFat.size, len );
+#ifdef TRACY_ON_DEMAND
+        GetProfiler().DeferItem( *item );
+#endif
```
theblackunknown, line 214: 👍

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+        TracyMetalDebug(1<<0, TracyMetalPanic(, "Calibration: GPU timestamp (Metal): %llu", gpuTimestamp));
+
+        cpuTimestamp = Profiler::GetTime();
+        TracyMetalDebug(1<<0, TracyMetalPanic(, "Calibration: CPU timestamp (Tracy): %llu", cpuTimestamp));
+
+        float period = 1.0f;
+        
+        m_contextId = GetGpuCtxCounter().fetch_add(1);
+        
+        auto* item = Profiler::QueueSerial();
```
theblackunknown, line 160: Why is this `auto*` and not simply `auto` like everywhere else ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+        {
+            return true;
+        }
+
+        std::unique_lock lock (m_collectionMutex, std::adopt_lock);
+
+        uintptr_t begin = m_previousCheckpoint.load();
+        uintptr_t latestCheckpoint = m_queryCounter.load(); // TODO: MTLEvent? MTLFence?;
+        ZoneValue(begin);
+        ZoneValue(latestCheckpoint);
```
theblackunknown, line 241: Are thoe debug information ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+
+        uint32_t count = RingCount(begin, latestCheckpoint);
+        if (count == 0)   // no pending timestamp queries
+        {
+            //uintptr_t nextCheckpoint = m_queryCounter.load();
+            //if (nextCheckpoint != latestCheckpoint)
+            //{
+            //    // TODO: signal event / fence now?
+            //}
```
theblackunknown, line 250: What should we do with this comment ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+
+        // resolve up until the ring buffer boundary and let a subsequenty call
+        // to Collect handle the wrap-around
+        bool reallocateBuffer = false;
+        if (RingIndex(begin) + count >= RingSize())
+        {
+            count = RingSize() - RingIndex(begin);
+            reallocateBuffer = true;
+        }
+        ZoneValue(count);
```
theblackunknown, line 262: Is this a debug information ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+            if ((t_start == MTLCounterErrorValue)  || (t_end == MTLCounterErrorValue))
+            {
+                TracyMetalPanic(, "Collect: invalid timestamp (MTLCounterErrorValue) at %u.", k);
+                break;
+            }
+            // Metal will initialize timestamp buffer with zeroes; encountering a zero-value
+            // timestamp means that the timestamp has not been written and resolved yet
+            if ((t_start == 0) || (t_end == 0))
+            {
+                auto checkTime = std::chrono::high_resolution_clock::now();
```
theblackunknown, line 305: Why not using `Profiler::GetTime()` here ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+                auto requestTime = m_timestampRequestTime[k];
+                auto ms_in_flight = std::chrono::duration<float>(checkTime-requestTime).count()*1000.0f;
+                TracyMetalDebug(1<<4, TracyMetalPanic(, "Collect: invalid timestamp (zero) at %u [%.0fms in flight].", k, ms_in_flight));
+                const float timeout_ms = 200.0f;
+                if (ms_in_flight < timeout_ms)
+                    break;
```
theblackunknown, line 311: What will happen to those ? They silently get discarded because they are too small ?
Why the choice of 200ms ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+                ZoneScopedN("TracyMetal::Collect::Drop");
+                TracyMetalPanic(, "Collect: giving up on timestamp at %u [%.0fms in flight].", k, ms_in_flight);
+                t_start = m_mostRecentTimestamp + 5;
+                t_end = t_start + 5;
```
theblackunknown, line 315: Does this means you try tp "patch" unresolved or not yet resolved timestamps ? 
Can't we defer their resolution instead ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+#endif
+
+    tracy_force_inline ~MetalZoneScope()
+    {
+        if( !m_active ) return;
+
+        SubmitZoneEndGpu(m_ctx, m_query.idx + 1);
+    }
+    
+    TracyMetalZoneScopeWireTap;
```
theblackunknown, line 556: What is the use of this ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+
+private:
+    const bool m_active;
+
+    MetalCtx* m_ctx;
+    id<MTLComputeCommandEncoder> m_cmdEncoder;
```
theblackunknown, line 562: This is currently unused because the above code is within `#if 0 ... #endif`

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+    {
+        auto* item = Profiler::QueueSerial();
+        MemWrite( &item->hdr.type, QueueType::GpuZoneEndSerial );
+        MemWrite( &item->gpuZoneEnd.cpuTime, Profiler::GetTime() );
+        MemWrite( &item->gpuZoneEnd.thread, GetThreadHandle() );
+        MemWrite( &item->gpuZoneEnd.queryId, uint16_t( queryId ) );
+        MemWrite( &item->gpuZoneEnd.context, ctx->GetContextId() );
+        Profiler::QueueSerialFinish();
+        
+        TracyMetalDebug(1<<2, TracyAllocN((void*)(uintptr_t)queryId, 1, "TracyMetalGpuZone"));
```
theblackunknown, line 588: Are those not supposed to be pair of `TracyAllocN`/`TracyFreeN` ?

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+
+        SubmitZoneEndGpu(m_ctx, m_query.idx + 1);
+    }
+    
+    TracyMetalZoneScopeWireTap;
+
+private:
+    const bool m_active;
+
+    MetalCtx* m_ctx;
```
theblackunknown, line 561: Alternatively you could just store the context ID

theblackunknown (commented):
Very nice work Marcos.
I have made a first pass of code review without testing so far.
The PR description and analysis is greatly appreciated !

Can you also clarify if this code is supposed to be use with ObjC ARC or not ?
I know that codebase from codebase we can run into different expectations with this behavior.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+    char buffer [1024]; \
+    snprintf(buffer, sizeof(buffer), "TracyMetal: " msg VA_ARGS(__VA_ARGS__)); \
+    TracyMessageC(buffer, strlen(buffer), tracy::Color::OrangeRed); \
+    fprintf(stderr, "%s\n", buffer); \
+    assert(false && "TracyMetal: " msg); \
+    ret; \
+    } while(false);
+
+
+#define TRACY_METAL_DEBUG_MASK (0)
```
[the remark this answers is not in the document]
slomp: Yes, at this initial stage, I'd like to have the verbose debugging log available when needed.
Should someone report issues, I can ask them to set this define prior to including the header.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+    enum { MaxQueries = 4 * 1024 };    // Metal: between 8 and 32768 _BYTES_...
+
+public:
+    MetalCtx(id<MTLDevice> device)
+    : m_device(device)
+    {
+        ZoneScopedNC("TracyMetalCtx", tracy::Color::Red4);
+        
+        TracyMetalDebug(1<<0, TracyMetalPanic(, "MTLCounterErrorValue = 0x%llx", MTLCounterErrorValue));
+        TracyMetalDebug(1<<0, TracyMetalPanic(, "MTLCounterDontSample = 0x%llx", MTLCounterDontSample));
```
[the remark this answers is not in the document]
slomp: Yes, at least at this initial stage. Should people start reporting issues, this can help me triage the problem.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+            TracyMetalPanic(, "WARNING: timestamp sampling at blit boundary is not supported.");
+        }
+        if (![m_device supportsCounterSampling:MTLCounterSamplingPointAtDispatchBoundary])
+        {
+            TracyMetalPanic(, "WARNING: timestamp sampling at compute dispatch boundary is not supported.");
+        }
+        if (![m_device supportsCounterSampling:MTLCounterSamplingPointAtTileDispatchBoundary])
+        {
+            TracyMetalPanic(, "WARNING: timestamp sampling at tile dispatch boundary is not supported.");
+        }
```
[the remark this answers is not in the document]
slomp: Yeah, good point, I'll do that.
In the future, you'd request the granularity expected, and if that fails, then a panic is issued.

public/tracy/TracyMetal.hmm
slomp: That's what Tracy expects. The CPU timestamp reported when creating the GPU context must be the timestamp that Tracy understands. This is consistent with other backends as well.

public/tracy/TracyMetal.hmm
slomp: Calibration is just to get better zone alignment in the timeline.
There are a few ways that Apple recommends recalibrating CPU and GPU timestamps, but I'll leave it to a subsequent PR to experiment with that.

public/tracy/TracyMetal.hmm
slomp: Probably a fluke of the Vulkan back-end.
In fact, there are a few things in the Vulkan back-end that I think I can simplify and make it more efficient. I'll get to that eventually.

public/tracy/TracyMetal.hmm
slomp: Debug Information, I'll try to move them to TracyMetalDebug macro.

public/tracy/TracyMetal.hmm
slomp: No particular reason. (I suppose one could argue that the extra `*` makes it clear it's a pointer.)

public/tracy/TracyMetal.hmm
slomp: Yup; will try to move them to `TracyMetalDebug` macro.

public/tracy/TracyMetal.hmm
slomp: I have some ideas for signaling in Metal, but I'd rather experiment with that in a future PR.
I'm leaving the comment with TODO there as a reminder.

public/tracy/TracyMetal.hmm
slomp: Empirical choice based on the applications I tested.
This is reasonable for any interactive, game-like loop.
(I'll add a macro for that)

public/tracy/TracyMetal.hmm
slomp: It's a debugging technique. An application can define the macro prior to including TracyMetal.hmm and run application-specific code at that point. Will probably go away once I am confident the back-end is working well.

public/tracy/TracyMetal.hmm
slomp: Yes, and I'll be working on that in a subsequent PR. I don't have hardware that supports command granularity right now.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+                TracyMetalPanic(, "Collect: invalid timestamp (MTLCounterErrorValue) at %u.", k);
+                break;
+            }
+            // Metal will initialize timestamp buffer with zeroes; encountering a zero-value
+            // timestamp means that the timestamp has not been written and resolved yet
+            if ((t_start == 0) || (t_end == 0))
+            {
+                auto checkTime = std::chrono::high_resolution_clock::now();
+                auto requestTime = m_timestampRequestTime[k];
+                auto ms_in_flight = std::chrono::duration<float>(checkTime-requestTime).count()*1000.0f;
```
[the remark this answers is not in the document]
slomp: Discord: https://discord.com/channels/585214693895962624/585214693895962630/1242530042366394409

public/tracy/TracyMetal.hmm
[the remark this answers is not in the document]
theblackunknown: It is understood that you want people to possibly override the macro, but the right coding pattern to do so is to wrap the `#define` with `#ifndef ... #endif` which is missing as of writing to avoid compiler warnings (and thus maybe errors) in client codebase.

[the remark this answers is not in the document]
slomp: Ah, LOL, yeah, I did not realize this was a code change _suggestion_. You are absolutely right!

public/tracy/TracyMetal.hmm
slomp: The units in Profiler::GetTime are unknown, and there's no interface, public or private, to convert two Profiler timestamps to a time span in a time quantity.
As a matter of fact, I inquired about it a while back in the Discord server:
https://discord.com/channels/585214693895962624/585214693895962630/1241523462758203552

(I got green light to implement the interface, but I'll leave that to a future PR)

public/tracy/TracyMetal.hmm
slomp: We can't defer, because the main reason a timestamp can't be resolved is because it the command encoder with the associated timestamp(s) was never scheduled for execution.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
+#ifndef __TRACYMETAL_HMM__
+#define __TRACYMETAL_HMM__
+
+/* The Metal back-end in Tracy operates differently than other GPU back-ends like Vulkan,
```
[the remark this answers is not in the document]
slomp: Added some additional comments about non-Apple-Silicon expectations.

public/tracy/TracyMetal.hmm
```diff
@@ -0,0 +1,634 @@
...
+        auto count = RingCount(m_previousCheckpoint, id);
+        if (count >= MaxQueries)
+        {
+            // TODO: return a proper (hidden) "sentinel" query
+            Query sentinel = Query{ m_counterSampleBuffers[1], MaxQueries-2 };
+            TracyMetalPanic(
+                return sentinel,
+                "NextQueryId: FULL! too many pending timestamp queries. [%llu, %llu] (%u)",
+                m_previousCheckpoint.load(), id, count
+            );
```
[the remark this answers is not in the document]
slomp: The warning message nows hints the user to call TracyMetalCollect more frequently.

public/tracy/TracyMetal.hmm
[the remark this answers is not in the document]
slomp: done!

public/tracy/TracyMetal.hmm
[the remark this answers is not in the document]
slomp: done!

public/tracy/TracyMetal.hmm
slomp: done!

public/tracy/TracyMetal.hmm
slomp: (FWIW, my understanding is that it does not matter which thread is acknowledged for this type of `GpuNewContext` message)

public/tracy/TracyMetal.hmm
slomp: done!

public/tracy/TracyMetal.hmm
slomp: done!

public/tracy/TracyMetal.hmm
slomp: done!

public/tracy/TracyMetal.hmm
slomp: a new macro has been introduced to control the timestamp abandonment.

public/tracy/TracyMetal.hmm
slomp: `TracyFree` happens in `Collect`, when the two timestamps announced here (begin and end timestamps) are resolved.

public/tracy/TracyMetal.hmm
slomp: The other back-ends seem to all store the context pointer, so I'll stick to that.

slomp:
@wolfpld Please review the changes to the manual, and if it looks good to you, this PR is ready to merge!

public/common/TracyQueue.hpp
```diff
@@ -401,7 +401,8 @@ enum class GpuContextType : uint8_t
     Vulkan,
     OpenCL,
     Direct3D12,
-    Direct3D11
+    Direct3D11,
+    Metal
```
wolfpld, line 405: This requires a protocol version bump (I'll do it).

PR #796, opened 2024-05-21
Fix savings calculation.

mcourteaux:
The fraction printed represented how much of the time was left after the savings are deduced. However, as it's printed after the "Savings:" label, it makes more sense to report how much time of the original one was saved.

Going from 1s to 100ms now reports 90% savings, instead of 10%.

### DIFF
--- a/profiler/src/profiler/TracyView_Compare.cpp
+++ b/profiler/src/profiler/TracyView_Compare.cpp
@@ -185,6 +185,44 @@ static void PrintDiff( const std::string& diff )
     }
 }
 
+static void PrintSpeedupOrSlowdown( double time_this, double time_external )
+{
+    const char *label, *label2;
+    const char *time_diff = TimeToString( abs( time_external - time_this ) );
+    ImVec4 color;
+    double factor;
+    if( time_external >= time_this )
+    {
+        label = "Speedup:", label2 = "faster";
+        color = ImVec4( 0.1f, 0.6f, 0.1f, 1.0f);
+        factor = time_external / time_this;
+    } else {
+        label = "Slowdown:", label2 = "slower";
+        color = ImVec4( 0.8f, 0.1f, 0.1f, 1.0f );
+        factor = time_this / time_external;
+    }
+    TextColoredUnformatted( color, label );
+    ImGui::SameLine();
+    TextColoredUnformatted(color, time_diff );
+    ImGui::SameLine();
+    TextColoredUnformatted( color, label2 );
+    ImGui::SameLine();
+    ImGui::Spacing();
+    ImGui::SameLine();
+    char buf[64];
+    memcpy( buf, "( ", 2 );
+    char *ptr = &buf[2];
+    ptr = PrintFloat( ptr, buf + sizeof(buf), factor, 3 );
+    memcpy( ptr, "x ", 2 );
+    ptr += 2;
+    int ssz = strlen( label2 );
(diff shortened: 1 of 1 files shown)


wolfpld:
Can you show how the negative case will look like?

mcourteaux:
Reduction:
image

Slow down:
image

Not very intuitive indeed. In this case, this would mean that we have 3.59 times slower (x3.59 == +2.59 == +259%).
Perhaps we should make the displayed information conditional on whether it's a slowdown or a speedup?

```
if (faster) {
  // talk about savings: time saved + fraction of original reduced + speedup factor
} else {
  // talk about slowdown: time lost + slowdown factor
}

```

Opinions?

wolfpld:
It may be worth to check how it looks. I would additionally color-code the negative result (eg., with a red text), to make it obvious that this isn't a speedup.

mcourteaux:
Okay, I'll update PR soon.

mcourteaux:
New version, with speedup:
image

And a slowdown:
image

I do like this, however, now that both the total time and the mean time are displayed with speedup/slowdown, I would argue that the normalization feature should be turned off. Thoughts?

mcourteaux:
Hmm, now reporting the absolute value of the change, to get rid of the confusion - sign in the front when dealing with a slowdown:

image

wolfpld:
Both slowdown labels should include a reference to what they are describing (ie., total / mean).

Normalization is useful when you can't have two traces with exactly the same number of function calls. The user has to consciously enable this option and is well informed about what will happen if they do.

I'm not sure the faster/slower reading is clear. For example, does 1.2x faster mean that the code runs 120% or 20% faster compared to the baseline? The value that was originally displayed was pretty clear in this case, even if it was mislabeled. It should read "execution time of the new code, relative to the old" instead of "savings". But that's quite a mouthful.

Some AI-generated proposals for a shorter label, Bing:

1.    New vs Old Exec Time
2.    Code Time Comparison
3.    Relative Code Speed
4.    Exec Time Ratio
5.    New/Old Time
6.    Code Speed Delta
7.    Time Efficiency
8.    Performance Shift
9.    Speed Factor
10.    Time Contrast

Llama3:

1.    New/Old Time
2.    Time Ratio
3.    Rel. Exec. Time
4.    New vs Old Time
5.    Time Delta
6.    Exec. Time Rel.
7.    Time Diff
8.    New/Old Ratio
9.    Time Change
10.    Time Rel.

I can't say that I particularly like any of these suggestions.

> 120% or 20% faster compared to the baseline

Maybe that's the answer. "10% of the baseline", or "of the original time" for your 1 s vs 100 ms example?

mcourteaux:
What about "1.430x as fast as external" and "0.130x as fast as external"? That's pretty unambiguous I think.

mcourteaux:
What about this?
image

wolfpld:
"x% of external" would work for me. It's short and pretty clear: "10% of external" or "250% of external".

mcourteaux:
I actually think I like this less verbose, more symbolic test better:

image

mcourteaux:
Not gonna lie, simplicity wins in my opinion. Sorry for the spam. I'll push this:

image

mcourteaux:
image

I like it! If there is anything, it'll have to wait, as I have to go. So feel free to merge after my push if you like it!

wolfpld:
I did some final adjustments in cf234411.

obraz

PR #815, opened 2024-06-17
Add scroll modifier options

imkunet:
### Problem statement

On Apple laptops, using the trackpad in Tracy scrolls far more than it does in other applications. To remedy this, scroll modifier options have been added.

### Pitfalls

- Somewhat weird easing behavior
- Frame overview vertical scrolling is still unchanged due to difficulty of implementing a multiplier for the whole values

### DIFF
--- a/profiler/src/main.cpp
+++ b/profiler/src/main.cpp
@@ -221,9 +221,12 @@ static void LoadConfig()
     if( !ini ) return;
 
     int v;
+    double v1;
     if( ini_sget( ini, "core", "threadedRendering", "%d", &v ) ) s_config.threadedRendering = v;
     if( ini_sget( ini, "core", "focusLostLimit", "%d", &v ) ) s_config.focusLostLimit = v;
     if( ini_sget( ini, "timeline", "targetFps", "%d", &v ) && v >= 1 && v < 10000 ) s_config.targetFps = v;
+    if( ini_sget( ini, "timeline", "horizontalScrollMultiplier", "%lf", &v1 ) && v1 > 0.0 ) s_config.horizontalScrollMultiplier = v1;
+    if( ini_sget( ini, "timeline", "verticalScrollMultiplier", "%lf", &v1 ) && v1 > 0.0 ) s_config.verticalScrollMultiplier = v1;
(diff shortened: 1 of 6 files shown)


wolfpld:
This may be a good feature to have, but I'm not sure it's the right solution for the problem at hand. How will this setting affect users who want to use both touchpad and mouse? It seems to me that the correct solution is to somehow process the Apple touchpad values to be more in line with what is available with other hardware configurations.

See https://github.com/ocornut/imgui/pull/6096 and https://github.com/ocornut/imgui/issues/4019 for possibly related discussion.

imkunet:
Absolutely agreed, it is a patchwork solution especially considering that resolving the behavioral differences between trackpad and mouse wheel would solve the frame overview zooming because that relies on the behavior of the mouse wheel having individual "notches".

From what I was able to glean from the referenced issues, this is a non-trivial issue on the part of ImGui and seems to require whole new systems just to handle touch-like events to be built in order to better handle this.

Perhaps emulation of individual wheel notches with some ratio of a delta wheel to individual click is the way to go? That of course raising questions of distinguishing between trackpad and mouse on the scroll events which doesn't appear to exist given the linked issue's discussion[^1].

[^1]: https://github.com/ocornut/imgui/issues/6103#issuecomment-1398851694

wolfpld:
> Perhaps emulation of individual wheel notches with some ratio of a delta wheel to individual click is the way to go?

I don't think this is a good solution. Ideally, the amount of touchpad zoom should be related to the amount of finger movement. Currently, it doesn't. It's too sensitive to fine movements, and the response to fast movements feels sluggish.

I don't remember the exact details of what the code is doing and why. The zoom was implemented with mouse wheel notches in mind, and the touchpad zoom was fine tuned later to make it somewhat usable. It should use a separate code path with no easing and direct mapping of input values to zoom changes, but that requires knowledge of which input device is being used.

Mentioning @ocornut, as the discussion is relevant to him.

ocornut:
I am going to be mostly away this week. Interested in what you find.

The io.AddXXXX API can now be used to distinguish touchpad vs mouse so this be useful if you need to specialize something here.

I am not well versed in Mac so there may be something else, eg a high-level zoom event?

wolfpld:
Merged manually.

PR #833, opened 2024-07-17
[test] Trigger failures via USR1/USR2

Arpafaucon:
Add signal handlers to trigger special conditions in `tracy-test`. This proves useful when working on the `capture` utility to check edge conditions.

### USR1: abort the program

example of call & output

maim_2024-07-17T17-01-39+02-00

in the profiler: 

maim_2024-07-17T16-38-32+02-00

### USR2: send buggy `TracyFree` event, to trigger an instrumentation failure

```
❯ timeout --signal USR2 0.4 ./tracy-test
Triggering instrumentation failure
terminate called without an active exception
timeout: the monitored command dumped core
```

maim_2024-07-17T17-01-09+02-00

### DIFF
--- a/test/test.cpp
+++ b/test/test.cpp
@@ -1,4 +1,6 @@
+#include <atomic>
 #include <chrono>
+#include <csignal>
 #include <iostream>
 #include <mutex>
 #include <thread>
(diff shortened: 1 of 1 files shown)


test/test.cpp
```diff
@@ -1,4 +1,6 @@
+#include <atomic>
 #include <chrono>
+#include <csignal>
```
wolfpld, line 3: Use `signal.h`.

Arpafaucon: done. Out of curiosity, what's the difference ? (always though it was the same thing, but for C++)

wolfpld: `csignal` guarantees that functions are available in the `std` namespace, and *maybe* in the global one.
`signal.h` guarantees that functions are available in the global namespace, and *maybe* in `std`.

Arpafaucon: thanks for the explanation !

PR #834, opened 2024-07-17
Fix achievement mgr crashing on uninitialized data

Aldarrion:
Both the achievement manager and the achievement data were globals in separate compilation units. Order of initialization is not defined in that case which sometimes caused the mgr to load uninitialized data and crash when an achievement should have been achieved but was not found.


wolfpld:
What's the impact of the issue you are fixing here?

wolfpld:
> It does not happen in your pre-build binaries

Ok, so no new release is needed.

Aldarrion:
Yes, fortunately not :) But it will prevent potential issues in future releases.

PR #835, opened 2024-07-17
[suggestion] Add .clang-format to help fit the existing coding style

Arpafaucon:
While working on #825 I found useful to spin up this config file to be able to fit the existing coding style.

It's not perfect, but I already goes a long way; maybe you'd be interested ?

Known caveats:
- could not manage to keep small enum on a single line
- you put 2 spaces in ifdefs (instead of the 4 elsewhere), haven't managed to replicate it
- it's not possible to remove the spaces around binary operators https://github.com/llvm/llvm-project/issues/31431
 
I committed 2 files so you can quickly look at the diff and get a sense of how files are affected

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,18 @@
+# Empirical format config, based on observed style guide
+# Use this only as an help to fit the surrounding code style - don't reformat whole files at once
+---
+BasedOnStyle: LLVM
+AllowShortIfStatementsOnASingleLine: WithoutElse
+AllowShortLoopsOnASingleLine: true
+AlwaysBreakTemplateDeclarations: Yes
+BreakBeforeBraces: Allman
+BreakConstructorInitializers: BeforeComma
+BreakStringLiterals: false
+ColumnLimit: 120
+FixNamespaceComments: false
+IndentPPDirectives: AfterHash
+IndentWidth: 4
+PointerAlignment: Left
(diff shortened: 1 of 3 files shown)


wolfpld:
How does this work in practice, when you are editing files? Proper code formatting is always a problem and it's good to automate this, but I don't want to be fighting with a system that is "fixing" deliberate side-steps from the established convention, like in the examples you show.

Arpafaucon:
I use VSCode; I don't know how much this applies to other users.

Personnally (and in my company), I hate the "format on save" option of VSCode, because it's acting behind your back. 

- In python source code, we adopted `black` with its strict & deterministic strategy. I like not having to discuss style issues in PR anymore. If you need to deviate from the formatting (typically when entering big arrays of numbers in a matrix-like format) you can always insert comments to locally disable the reformatting. We enforce this in CI, works well in practice, I think
- for C++, we haven't yet found a default good enough everywhere. So I encourage to use the "format selection" feature to only reformat the lines thet were just edited. This way, diffs stay small, but we get a tool to fix all the minor style issues and have a overall coherent style. We don't enforce this in CI, and assume devs are mature enough (not always the case, though, but we manage). That's what I would suggest for this repo, at least to start

wolfpld:
I have cherry-picked the commit. Formatting is done only on demand, so it's fine.

PR #839, opened 2024-07-19
Expose `TRACY_VERBOSE` in cmake for debug

Arpafaucon:
Proved very useful to debug kernel permissions issues ! I did not know it existed, so I added a small entry in the doc to mention it. It seems that's a frequent topic of discussion in the discord, that might help newcomers debug by themselves

### DIFF
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -638,6 +638,10 @@ Although the basic features will work without them, you'll have to grant elevate
     \item \texttt{-{}-pid=host}
 \end{itemize}
 
+\subsubsection{Troubleshooting}
+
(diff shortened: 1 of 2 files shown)


manual/tracy.tex
```diff
@@ -638,6 +638,10 @@ \subsubsection{Problematic platforms}
     \item \texttt{-{}-pid=host}
 \end{itemize}
 
+\subsubsection{Troubleshooting}
+
+Setting the \texttt{TRACY\_VERBOSE} variable will make the client display advanced information about the detected features. By matching those debug prints to the source code, you might be able to uncover why some of the features missing on your platform.
```
wolfpld, line 643: why some of the features +are+ missing

Arpafaucon: nice catch ! fixed

PR #853, opened 2024-08-02
fix can`t caputure context swithces on rooted Android 11 devices

firefhy:


wolfpld:
I think it would be more robust to scan mount points for debugfs to support all other cases.

PR #859, opened 2024-08-09
Lua: ensure source location is not longer than 255 characters

brakhane:
When a Lua script was loaded through loadstring, dbg.source is the whole source code; this not only causes a lot of unnecessary data to be sent, it can even crash Tracy if the source string is larger than 64 KB.

Therefore, use the short version which is never more than 60 characters.


wolfpld:
This breaks file contents retrieval (`script.lua` becomes `[string "script.lua"]`).

brakhane:
@wolfpld I've modified the code so it uses the first 255 bytes of dbg.source without doing the mangling that short_src does.

Let me know if it should be solved differently.

PR #880, opened 2024-09-11
[profiler] Add install target

ManifoldFR:
This PR adds an `install` CMake target for the profiler, so that we can just do `make install` / `ninja install` to install the binary to the prefix 😃


ManifoldFR:
Thanks for merging !

PR #881, opened 2024-09-12
Add basic error handling

tygyh:
Prevents creating strings with negative length


wolfpld:
What's the failure mode that these changes are protecting against?

tygyh:
> What's the failure mode that these changes are protecting against?

`ftell` returns `-1L` anytime it fails to load a file and `sprintf` returns `-1` when it fails to print.

PR #882, opened 2024-09-13
Patching Apple/Metal back-end Debug issues

slomp:
No changes in behavior or protocol.
Just better debugging and code scoping.


slomp:
@wolfpld This is just some fixes I noticed while doing additional debugging. Nothing that changes the behavior of the code, just better debugging facilities and some code scope refactoring. Ready to merge, if it looks good to you.

Issue #884, opened 2024-09-13
Proposal: Color zones based on (grand)parent Zone color.


mcourteaux:
I thought this to be useful when you mark one zone with a color, and then you can easily spot that that whole sub-tree of zones would have the same color. For example, when doing a game: everything physics related could be under one "physics" zone with a specific manual color, and the whole subtree would inherit that color. Example:

![image](https://github.com/user-attachments/assets/b63d335b-b815-46ff-9ac8-5848b69b7987)

Here the whole `update_ui` subtree gets the zone color from `update_ui`.

With _Source location dynamic_ it also looks interesting:

![image](https://github.com/user-attachments/assets/504ec00e-832d-4989-a0b8-55c1514f883d)

Inefficient proof-of-concept code:

```cpp
uint32_t View::GetZoneColor( const ZoneEvent& ev, uint64_t thread, int depth )
{
    const ZoneEvent *evt = &ev;
    if( !m_vd.forceColors )
    {
        while( evt )
        {
            if( m_worker.HasZoneExtra( *evt ) )
            {
                const auto custom_color = m_worker.GetZoneExtra( *evt ).color.Val();
                if( custom_color != 0 ) return custom_color | 0xFF000000;
            }
            const auto sl = evt->SrcLoc();
            const auto& srcloc = m_worker.GetSourceLocation( sl );
            const auto color = srcloc.color;
            if( color != 0 ) return color | 0xFF000000;

            evt = GetZoneParent( *evt );
        }
    }
    const auto sl = ev.SrcLoc();
    const auto& srcloc = m_worker.GetSourceLocation( sl );
    switch( m_vd.dynamicColors )
    {
    case 0:
        return 0xFFCC5555;
    case 1:
        return GetHsvColor( thread, depth );
    case 2:
        return GetRawSrcLocColor( srcloc, depth );
    default:
        assert( false );
        return 0;
    }
}
```

Thoughts on this?

wolfpld:
Seems fine as an option. `GetZoneParent()` does a top-down walk, so calling it for each of the parents wastes work. Preparing a temporary vector of all parents would be more efficient, which with the information we have at this point (thread, depth) wouldn't even need any reallocs, as we know exactly how big the vector should be.

mcourteaux:
So we could do an `alloca(depth * sizeof(ZoneEvent))` with a `GetZoneParentUntilTopLevel()`. How does that sound? I could implement, add a checkbox in the options UI, test with it for a while and PR that.

[...]

PR #885, opened 2024-09-13
Implement Zones inheriting colors from parent Zones as an option.

mcourteaux:
Option for coloring zones without custom color by inheriting it from a parent that has a custom color. If inherited, the color is made slightly darker to distinguish the "sources" of the colors and the ones inheriting from each other.

Option saved to ini file.

Performance considerations:
 
 - parents only fetched when feature enabled _and_ zone-to-be-drawn does not have a custom color.
 - parent array is `alloca()`'d

Screenshot without:

image

With:

image

### DIFF
--- a/profiler/src/profiler/TracyView_Options.cpp
+++ b/profiler/src/profiler/TracyView_Options.cpp
@@ -225,6 +225,9 @@ void View::DrawOptions()
     ImGui::SameLine();
     bool forceColors = m_vd.forceColors;
     if( SmallCheckbox( "Ignore custom", &forceColors ) ) m_vd.forceColors = forceColors;
+    ImGui::SameLine();
+    bool inheritColors = m_vd.inheritParentColors;
+    if( SmallCheckbox( "Inherit parent colors", &inheritColors ) ) m_vd.inheritParentColors = inheritColors;
     ImGui::Indent();
     ImGui::PushStyleVar( ImGuiStyleVar_FramePadding, ImVec2( 0, 0 ) );
     ImGui::RadioButton( "Static", &ival, 0 );
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -3412,6 +3412,7 @@ In this window, you can set various trace-related options. For example, the time
 \item \emph{Source location dynamic} -- Zone color is determined by source location (function name) and depth level.
 \end{itemize}
 Enabling the \emph{Ignore custom} option will force usage of the selected zone coloring scheme, disregarding any colors set by the user in profiled code.
+Enabling the \emph{Inherit parent colors} option will cause zones that have a color set by the user in the profiled code to be propagated down to the child zones, although slightly darker.
 \item \emph{\faRulerHorizontal{} Zone name shortening} -- controls display behavior of long zone names, which don't fit inside a zone box:
 \begin{itemize}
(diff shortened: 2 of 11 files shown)


mcourteaux:
Closes #884

mcourteaux:
Fixed the capitalization in the screenshot: `Inherit Colors` -> `Inherit colors`.
Perhaps I should even change it to `Inherit custom`?

mcourteaux:
Another example of this in my jobsystem. I used this to make every job get a different color: 

image

wolfpld:
> Fixed the capitalization in the screenshot: `Inherit Colors` -> `Inherit colors`.
> Perhaps I should even change it to `Inherit custom`?

I would name it "Inherit parent colors".

I would also rename `GetZoneParentsUntilTopLevel` to `GetAllZoneParents`, with an assert that the final depth is the depth passed as an parameter.

One more thought: the zone timeline drawing routines are already passing through all the parents. Maybe it would be simpler to just pass the user zone color down to children right there, so that no parent lookup is needed at all?

mcourteaux:
> One more thought: the zone timeline drawing routines are already passing through all the parents. Maybe it would be simpler to just pass the user zone color down to children right there, so that no parent lookup is needed at all?

I could do that, but that would require adding a `uint32_t custom_color` field to `struct TimelineDraw`, which you've been trying really hard to make it small it seems.

> I would name it "Inherit parent colors".
> 
> I would also rename `GetZoneParentsUntilTopLevel` to `GetAllZoneParents`, with an assert that the final depth is the depth passed as an parameter.

Will do.

wolfpld:
> I could do that, but that would require adding a `uint32_t custom_color` field to `struct TimelineDraw`, which you've been trying really hard to make it small it seems.

The amount of timeline items to draw turns out to be surprisingly low, so it's not a concern.

mcourteaux:
It's more annoying to implement than I expected. The utility function `View::GetZoneColor()` is `private` to `View`. So from within `TimelineItemThread::PreprocessZoneLevel()`, I can't use that function to get the color there and populate the `TimelineDraw::customColor` field.

wolfpld:
Just change it to `public`.

mcourteaux:
Changed it and reverted old changes (rewrote git history).

This new approach also allowed me to implement correctly the color for the Folded zones:

image

profiler/src/profiler/TracyView_Options.cpp
```diff
@@ -225,6 +225,9 @@ void View::DrawOptions()
     ImGui::SameLine();
     bool forceColors = m_vd.forceColors;
     if( SmallCheckbox( "Ignore custom", &forceColors ) ) m_vd.forceColors = forceColors;
+    ImGui::SameLine();
+    bool inheritColors = m_vd.inheritParentColors;
+    if( SmallCheckbox( "Inherit parent colors", &inheritColors) ) m_vd.inheritParentColors = inheritColors;
```
wolfpld, line 230: Add missing space.

mcourteaux:
Done 😄

wolfpld:
I forgot that zone colors may come from two different sources, so I took a deep dive to understand the code and b359936c04e496c2c5a5a75d4aa5b0c4110aea4e is the result. Please take a look if everything there works as you'd expect.

mcourteaux:
You introduced a bug somehow, I believe. Neighboring zones adopt the color as well, randomly. Will investigate:

image

The zone pointed at does not have any color.

Issue #888, opened 2024-09-23
Crash in HandleSymbolCodeQuery if module is unloaded


jkriegshauser:
We have a heavily modularized application. When running with privileges to get sys-tracing, the query backlog can grow quite high.

The shutdown order looks something like:
1. Unload most modules.
2. Unload our profiling-tracy module (blocks until `tracy::ShutdownProfiler()` returns)
3. Unload the rest of our modules
4. Exit application

Unfortunately, if a SymbolCodeQuery query is processed for a module that unloaded, we usually end up in the CrashHandler and the application is hung at this point.

Issue #887, opened 2024-09-23
Tracy Profiler crash on startup in Worker mutex lock with access violation on Windows


egorodet:
`tracy-profiler.exe` downloaded from GitHub releases v0.11.0 and v0.11.1 does not start silently.

When I build Tracy Profiler from sources, it works normally in `Debug` build, but crashes in `Release` and `ReleaseWithDebInfo` builds on attempt to lock mutex from thread loop of `ResolvService::Worker()`. Issue is reproducible on `master`, `v0.11.1` and `v0.11.0`. Previous release of Tracy `v0.10` was working fine.

![Tracy-Mutex-Crash-2](https://github.com/user-attachments/assets/db9fe796-998f-4ff8-b364-21cafdfffa1e)

![Tracy-Mutex-Crash-1](https://github.com/user-attachments/assets/9ed5a5c6-9611-44d6-b9eb-6c0f99aff0c1)

**Build command line**:
```cmd
cd X:\Git\Tracy\profiler
cmake -S . -B Build -G "Visual Studio 17 2022" -A x64
cmake --build Build --config Release
```

**System Information**:
- Windows 11 Pro 23H2 22631.4169
- Microsoft Visual Studio Enterprise 2022 v17.11.3

wolfpld:
See #862. The fact that you have MSVC 17.11 doesn't fit though.

egorodet:
> See #862. The fact that you have MSVC 17.11 doesn't fit though.

Oh, now I see. Looks like a new disaster :((( I though that MSVC Runtime is auto-updated by Visual Studio Installer along with other updates, but seems like it does not... 

Maybe instead of updating runtimes, it would be better to add definition `_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR` (as [suggested by Microsoft](https://developercommunity.visualstudio.com/t/Visual-Studio-17100-Update-leads-to-Pr/10669759?sort=newest)) to return compatibility with old runtime versions?

egorodet:
Btw, I've downloaded latest MSVC runtime installer `v14.40.33810.0` [from here](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#visual-studio-2015-2017-2019-and-2022) and it says that it is already installed! What's next?

![image](https://github.com/user-attachments/assets/b2e1e105-9c99-4a9f-bd6f-73ef6f2513ac)

wolfpld:
[...]

PR #889, opened 2024-09-23
Safe copy to fix issue #888

jkriegshauser:
Fix for issue #888 

Instead of attempting to read potentially unsafe memory that may be in a module that unloaded, this uses Structured Exception Handling on Windows or pipes on POSIX platforms to ensure that the entire memory region can be read.

### DIFF
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -111,6 +111,7 @@ extern "C" typedef char* (WINAPI *t_WineGetBuildId)();
 #else
 #  include <unistd.h>
 #  include <limits.h>
+#  include <fcntl.h>
 #endif
 #if defined __linux__
 #  include <sys/sysinfo.h>
@@ -1462,6 +1463,22 @@ Profiler::Profiler()
         m_userPort = atoi( userPort );
     }
 
+    m_safeSendBufferSize = 65536;
+    m_safeSendBuffer = (char*)tracy_malloc( m_safeSendBufferSize );
+
+#ifndef _WIN32
+    pipe(m_pipe);
+#  if defined __APPLE__ || defined BSD
+    // FreeBSD/XNU don't have F_SETPIPE_SZ, so use the default 
+    m_pipeBufSize = 16384;
+#  else
+    m_pipeBufSize = (int)(ptrdiff_t)m_safeSendBufferSize;
+    while( fcntl( m_pipe[0], F_SETPIPE_SZ, m_pipeBufSize ) == -1 && errno == EPERM )
+        m_pipeBufSize /= 2; // too big; reduce
+    m_pipeBufSize = fcntl( m_pipe[0], F_GETPIPE_SZ );
+#  endif
+#endif
+
 #if !defined(TRACY_DELAYED_INIT) || !defined(TRACY_MANUAL_LIFETIME)
     SpawnWorkerThreads();
 #endif
@@ -1487,7 +1504,7 @@ void Profiler::InstallCrashHandler()
 #endif
 
 #if defined _WIN32 && !defined TRACY_UWP && !defined TRACY_NO_CRASH_HANDLER
-    m_exceptionHandler = AddVectoredExceptionHandler( 1, CrashFilter );
+    m_exceptionHandler = AddVectoredExceptionHandler( 0, CrashFilter );
 #endif
 
 #ifndef TRACY_NO_CRASH_HANDLER
@@ -1599,6 +1616,12 @@ Profiler::~Profiler()
     m_kcore->~KCore();
     tracy_free( m_kcore );
 #endif
+
+#ifndef _WIN32
+    close( m_pipe[0] );
+    close( m_pipe[1] );
+#endif
+    tracy_free( m_safeSendBuffer );
 
     tracy_free( m_lz4Buf );
     tracy_free( m_buffer );
@@ -4017,13 +4040,13 @@ void Profiler::HandleSymbolCodeQuery( uint64_t symbol, uint32_t size )
     }
     else
     {
-        if( !EnsureReadable( symbol ) )
-        {
-            AckSymbolCodeNotAvailable();
+        // 'symbol' may have come from a module that has since unloaded, perform a safe copy before sending
+        if( withSafeCopy( (const char*)symbol, size, [this, symbol]( const char* buf, size_t size ) {
+            SendLongString( symbol, buf, size, QueueType::SymbolCode );
+        }))
             return;
-        }
-
-        SendLongString( symbol, (const char*)symbol, size, QueueType::SymbolCode );
+        
+        AckSymbolCodeNotAvailable();
     }
 }
 
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -1501,14 +1501,17 @@ void Profiler::InstallCrashHandler()
     sigaction( SIGPIPE, &crashHandler, &m_prevSignal.pipe );
     sigaction( SIGBUS, &crashHandler, &m_prevSignal.bus );
     sigaction( SIGABRT, &crashHandler, &m_prevSignal.abrt );
+
+    m_crashHandlerInstalled = true;
 #endif
 
 #if defined _WIN32 && !defined TRACY_UWP && !defined TRACY_NO_CRASH_HANDLER
-    m_exceptionHandler = AddVectoredExceptionHandler( 0, CrashFilter );
-#endif
-
-#ifndef TRACY_NO_CRASH_HANDLER
-    m_crashHandlerInstalled = true;
+    // We cannot use Vectored Exception handling because it catches application-wide frame-based SEH blocks. We only
+    // want to catch unhandled exceptions in the event that there is not already an unhandled exception filter.
+    if( auto prev = SetUnhandledExceptionFilter( CrashFilter ) )
+        SetUnhandledExceptionFilter( prev ); // Already had a handler => put it back
+    else
+        m_crashHandlerInstalled = true;
 #endif
 
 }
@@ -1516,7 +1519,12 @@ void Profiler::InstallCrashHandler()
 void Profiler::RemoveCrashHandler()
 {
 #if defined _WIN32 && !defined TRACY_UWP
-    if( m_crashHandlerInstalled ) RemoveVectoredExceptionHandler( m_exceptionHandler );
+    if( m_crashHandlerInstalled )
+    {
+        auto prev = SetUnhandledExceptionFilter( NULL );
+        if( prev != CrashFilter )
+            SetUnhandledExceptionFilter( prev ); // A different exception filter was installed over ours => put it back
+    }
 #endif
 
 #if defined __linux__ && !defined TRACY_NO_CRASH_HANDLER
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -3095,6 +3095,69 @@ bool Profiler::CommitData()
     return ret;
 }
 
+char* Profiler::SafeCopyProlog( const char* data, size_t size )
+{
+    bool success = true;
+    char* buf = m_safeSendBuffer;
+#ifndef NDEBUG
+    assert( !m_inUse.exchange(true) );
+#endif
+
+    if( size > m_safeSendBufferSize )
+    {
+        buf = (char*)tracy_malloc( size );
+    }
+
+#ifdef _WIN32
+    __try
+    {
+        memcpy( buf, data, size );
+    }
+    __except( 1 /*EXCEPTION_EXECUTE_HANDLER*/ )
+    {
+        success = false;
+    }
+#else
+    // Send through the pipe to ensure safe reads
+    for( size_t offset = 0; offset != size; /*in loop*/ )
+    {
+        size_t sendsize = size - offset;
+        ssize_t result1, result2;
+        while( ( result1 = write( m_pipe[1], data + offset, sendsize ) ) < 0 && errno == EINTR )
+            /* retry */;
+        if( result1 < 0 )
+        {
+            success = false;
+            break;
+        }
+        while( ( result2 = read( m_pipe[0], buf + offset, result1 ) ) < 0 && errno == EINTR )
+            /* retry */;
+        if( result2 != result1 )
+        {
+            success = false;
+            break;
+        }
+        offset += result1;
+    }
+#endif
+
+    if( success )
+        return buf;
+
+    SafeCopyEpilog( buf );
+    return nullptr;
+}
+
+void Profiler::SafeCopyEpilog( char* buf )
+{
+    if( buf != m_safeSendBuffer )
+        tracy_free( buf );
+    
+#ifndef NDEBUG
+    m_inUse.store( false );
+#endif
+}
+
 bool Profiler::SendData( const char* data, size_t len )
 {
     const lz4sz_t lz4sz = LZ4_compress_fast_continue( (LZ4_stream_t*)m_stream, data, m_lz4Buf + sizeof( lz4sz_t ), (int)len, LZ4Size, 1 );
@@ -4050,7 +4113,7 @@ void Profiler::HandleSymbolCodeQuery( uint64_t symbol, uint32_t size )
     else
     {
         // 'symbol' may have come from a module that has since unloaded, perform a safe copy before sending
-        if( withSafeCopy( (const char*)symbol, size, [this, symbol]( const char* buf, size_t size ) {
+        if( WithSafeCopy( (const char*)symbol, size, [this, symbol]( const char* buf, size_t size ) {
             SendLongString( symbol, buf, size, QueueType::SymbolCode );
         }))
             return;
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -1473,10 +1473,11 @@ Profiler::Profiler()
     m_pipeBufSize = 16384;
 #  else
     m_pipeBufSize = (int)(ptrdiff_t)m_safeSendBufferSize;
-    while( fcntl( m_pipe[0], F_SETPIPE_SZ, m_pipeBufSize ) == -1 && errno == EPERM )
+    while( fcntl( m_pipe[0], F_SETPIPE_SZ, m_pipeBufSize ) < 0 && errno == EPERM )
         m_pipeBufSize /= 2; // too big; reduce
     m_pipeBufSize = fcntl( m_pipe[0], F_GETPIPE_SZ );
 #  endif
+    fcntl( m_pipe[1], F_SETFL, O_NONBLOCK );
 #endif
 
 #if !defined(TRACY_DELAYED_INIT) || !defined(TRACY_MANUAL_LIFETIME)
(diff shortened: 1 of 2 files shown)


wolfpld:
What exactly is this code doing?

jkriegshauser:
Basically it treats memory requested by HandleSymbolCodeQuery as unsafe to read:
* On Windows, it does `memcpy()` inside a structured exception block. If an access violation occurs during the memcpy, the `__except` portion of the block is triggered and the memory cannot be safely read.
* On Linux/Mac, it writes the memory to a pipe which causes the kernel to attempt to read the memory. If the memory could not be read, `write()` sets `errno` to `EFAULT` and the memory cannot be safely read.

In both of these instances, if memory cannot be safely read, `withSafeCopy()` returns false and causes `HandleSymbolCodeQuery` to reply with `AckSymbolCodeNotAvailable()`. If memory was safely copied, `SendLongString()` is invoked with the symbol code.

jkriegshauser:
Looks like there are issues with Vectored exception handling on Windows blocking the frame-based handler. More info [here](https://learn.microsoft.com/en-us/answers/questions/1636330/some-doubt-in-exception-handling). I’ll address tomorrow.

wolfpld:
The `pipe` trick relying on `EFAULT` won't work.

> EFAULT buf is outside your accessible address space.

Glibc is forwarding the `write` call to a sys call, which is implemented in `fs/read_write.c` as:

```c
SYSCALL_DEFINE3(write, unsigned int, fd, const char __user *, buf,
		size_t, count)
{
	return ksys_write(fd, buf, count);
}
```

This goes to:

```c
ssize_t ksys_write(unsigned int fd, const char __user *buf, size_t count)
{
	struct fd f = fdget_pos(fd);
	ssize_t ret = -EBADF;

	if (fd_file(f)) {
		loff_t pos, *ppos = file_ppos(fd_file(f));
		if (ppos) {
			pos = *ppos;
			ppos = &pos;
		}
		ret = vfs_write(fd_file(f), buf, count, ppos);
		if (ret >= 0 && ppos)
			fd_file(f)->f_pos = pos;
		fdput_pos(f);
	}

	return ret;
}
```

Which goes to:

```c
ssize_t vfs_write(struct file *file, const char __user *buf, size_t count, loff_t *pos)
{
	ssize_t ret;

	if (!(file->f_mode & FMODE_WRITE))
		return -EBADF;
	if (!(file->f_mode & FMODE_CAN_WRITE))
		return -EINVAL;
	if (unlikely(!access_ok(buf, count)))
		return -EFAULT;
...
```

And `access_ok` is in `arch/x86/include/asm/uaccess_64.h`:

```c
/*
 * User pointers can have tag bits on x86-64.  This scheme tolerates
 * arbitrary values in those bits rather then masking them off.
 *
 * Enforce two rules:
 * 1. 'ptr' must be in the user half of the address space
 * 2. 'ptr+size' must not overflow into kernel addresses
 *
 * Note that addresses around the sign change are not valid addresses,
 * and will GP-fault even with LAM enabled if the sign bit is set (see
 * "CR3.LAM_SUP" that can narrow the canonicality check if we ever
 * enable it, but not remove it entirely).
 *
 * So the "overflow into kernel addresses" does not imply some sudden
 * exact boundary at the sign bit, and we can allow a lot of slop on the
 * size check.
 *
 * In fact, we could probably remove the size check entirely, since
 * any kernel accesses will be in increasing address order starting
 * at 'ptr', and even if the end might be in kernel space, we'll
 * hit the GP faults for non-canonical accesses before we ever get
 * there.
 *
 * That's a separate optimization, for now just handle the small
 * constant case.
 */
static inline bool __access_ok(const void __user *ptr, unsigned long size)
{
	if (__builtin_constant_p(size <= PAGE_SIZE) && size <= PAGE_SIZE) {
		return valid_user_address(ptr);
	} else {
		unsigned long sum = size + (__force unsigned long)ptr;

		return valid_user_address(sum) && sum >= (__force unsigned long)ptr;
	}
}
```

So you will only get `EFAULT` if you would want to step over into kernel space, not if you would want to access an unmapped memory address.

FWIW, Tracy on Linux is already protected against this problem by overriding `dlclose` to do nothing, which is conformant with the documented behavior:

> A successful return from dlclose() does not guarantee that the symbols associated with handle are  removed  from  the  caller's  address space.  In addition to references resulting from explicit dlopen() calls, a shared object may have been implicitly loaded (and reference counted) because of dependencies in other shared objects.  Only when all references have been released can the shared object be removed from the address space.

https://github.com/wolfpld/tracy/blob/master/public/client/TracyOverride.cpp

jkriegshauser:
We want the write to go through the pipe since the kernel can safely validate the memory for us. And yes, as you pointed out it does an initial access check through `__access_ok()`, but deeper in `pipe_write()` it will return `EFAULT` if copying from any page fails:
```
static ssize_t
pipe_write(struct kiocb *iocb, struct iov_iter *from)
{
    //...
			ret = copy_page_from_iter(buf->page, offset, chars, from);
			if (unlikely(ret < chars)) {
				ret = -EFAULT;
				goto out;
			}
    //...
			copied = copy_page_from_iter(page, 0, PAGE_SIZE, from);
			if (unlikely(copied < PAGE_SIZE && iov_iter_count(from))) {
				if (!ret)
					ret = -EFAULT;
				break;
			}
    //...
}
```
This is the behavior that we want since all of this is happening in the Tracy Profiler thread, and other threads are off doing whatever--we want to avoid the race condition. In any case, I'm not specifically looking for `errno == EFAULT` in my change, merely anything that causes a write to fail other than `EINTR` (though it should be `EFAULT` for the type of sanitization this is doing).

As to the _TracyOverride.cpp_ behavior, this is never actually getting triggered for us, possibly because we build the Tracy client as a module and it optionally gets loaded. And our plugin system notices when libraries don't unload as this makes hot-reloading impossible.

jkriegshauser:
I tested with this and discovered that a Linux pipe doesn't actually allow you to use the full amount of memory without blocking, so I now set the `O_NONBLOCK` flag and handle it appropriately. This also confirms that partially accessible memory produces an `EFAULT` from `write()`.

```cpp
    // Test withSafeCopy
    char* p = "test";
    // 0 bytes: valid
    assert(withSafeCopy(p, 0, [&](char* v, size_t s) { assert(v != p); assert(s == 0); }));
    // sizeof(p) bytes: valid with copy
    assert(withSafeCopy(p, 5, [&](char* v, size_t s) { assert(v != p); assert(s == 5); assert(strcmp(v, "test") == 0); }));
    // 1 byte at nullptr: invalid
    assert(!withSafeCopy(nullptr, 1, [](char*, size_t) { assert(false); }));

    // Create a region that is 4096 bytes of valid memory followed by 4096 bytes of unmapped memory.
    // [region+2048, region+4096+2048): invalid
    p = (char*)mmap(NULL, 4096 * 2, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    munmap(p + 4096, 4096);
    assert(!withSafeCopy(p + 2048, 4096, [](char*, size_t) { assert(false); }));
    munmap(p, 4096);

    // Create a region that is 2MB of valid memory followed by 4096 bytes of unmapped memory.
    p = (char*)mmap(NULL, 2 * 1024 * 1024 + 4096, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    munmap(p + 2 * 1024 * 1024, 4096);
    char const begin = 99;
    for (char* v = p, b = begin; v != (p + 2 * 1024 * 1024); ++v)
        *v = b++;
    // [region, region+2MB): valid (and bytes checked)
    assert(withSafeCopy(p, 2 * 1024 * 1024, [&](char* v, size_t s) {
        assert(v != p); assert(s == 2 * 1024 * 1024);
        bool match = true;
        for (char* u = v, b = begin; u != (v + s); ++u)
            match |= (*u == b++);
        assert(match);
    }));
    // [region+2048, region+2MB+2048): invalid
    assert(!withSafeCopy(p + 2048, 2 * 1024 * 1024, [](char*, size_t) { assert(false); }));
    munmap(p, 2 * 1024 * 1024);
```

public/client/TracyProfiler.cpp
```diff
@@ -1484,22 +1502,30 @@ void Profiler::InstallCrashHandler()
...
+    m_crashHandlerInstalled = true;
 #endif
 
-#ifndef TRACY_NO_CRASH_HANDLER
-    m_crashHandlerInstalled = true;
+#if defined _WIN32 && !defined TRACY_UWP && !defined TRACY_NO_CRASH_HANDLER
+    // We cannot use Vectored Exception handling because it catches application-wide frame-based SEH blocks. We only
+    // want to catch unhandled exceptions in the event that there is not already an unhandled exception filter.
+    if( auto prev = SetUnhandledExceptionFilter( CrashFilter ) )
+        SetUnhandledExceptionFilter( prev ); // Already had a handler => put it back
```
wolfpld, line 1513: What is the reason for this change?

public/client/TracyProfiler.hpp
```diff
@@ -833,6 +833,21 @@ class Profiler
         m_bufferOffset += int( len );
     }
 
+    char* SafeCopyProlog( const char* p, size_t size );
+    void SafeCopyEpilog( char* buf );
+
+    template<class Callable> // must be void( const char* buf, size_t size )
+    bool WithSafeCopy( const char* p, size_t size, Callable&& callable )
```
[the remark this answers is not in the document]
jkriegshauser: The template allows a lambda with captures to be used, otherwise we would have to pass a context.

public/client/TracyProfiler.cpp
jkriegshauser: Vectored exception handlers are called for every exception, including frame-based exception handlers that handle them. So any Windows application that would use Tracy along with frame-based SEH blocks would inadvertently cause `CrashFilter` to be called.

Instead, we want `CrashFilter` to be called only for unhandled exceptions, hence the change to `SetUnhandledExceptionFilter`.

The answer on [this page](https://stackoverflow.com/questions/54806802/filtering-critical-fatal-exceptions-for-use-with-addvectoredexceptionhandler/54879453#54879453) gives more information, specifically this portion:
> VEH handlers added by AddVectoredExceptionHandler will handle exceptions before they reach frame-based handlers.
> 
> Filter set by SetUnhandledExceptionFilter will handle exceptions after frame-based handlers failed.

jkriegshauser: Btw, here's a simple program that exemplifies what I'm talking about with vectored exceptions. If Tracy is connected, `CrashFilter` is called when the `memcpy` causes an access violation. If Tracy is not connected, the `CrashFilter` isn't installed and the program runs fine.
```cpp
int main(int argc, char** argv)
{
    tracy::StartupProfiler();

    std::this_thread::sleep_for(std::chrono::seconds(1)); // give time to connect

    {
        ZoneNamed(temp, true);

        __try
        {
            memcpy(nullptr, &temp, 1); // should crash?
        }
        __except (1)
        {
            // nope, control should end up here instead preventing the crash
        }

        std::this_thread::sleep_for(std::chrono::milliseconds(10));
    }

    tracy::ShutdownProfiler();

    return 0;   
}
```

wolfpld: Sigh. The `memcpy` above is ub, so any reasoning about what should happen is nonsensical. After #887 I don't have any energy left for Microsoft bullshit right now.

jkriegshauser: Yes, this simple example is UB, but the point is to exemplify a scenario where any Structured Exception is triggered and instead of being handled, the `CrashFilter` is invoked instead.

And yeah, understandable--887 looks rough.

jkriegshauser: Alright, here's another example. Integer divide-by-zero can be intercepted and handled with structured exceptions. If this happens while Tracy is connected with its vectored exception handler however, the profiler/viewer reports a crash and the application hangs. If Tracy is not connected, the vectored exception handler is not installed, and the application prints "Divide-by-zero without crash" as the exception was handled.
```cpp
        __try
        {
            int i = 1;
            int j = 1;
            --j;
            fprintf(stdout, "%d\n", i / j);
        }
        __except (1)
        {
            fprintf(stdout, "Divide-by-zero without crash\n");
        }
```

This of course merely illustrates that using vectored exception handlers will override _any_ frame-based exception handlers in an undesirable fashion, when what we really want is to handle _unhandled_ exceptions.

Yet another example is the old method of setting thread names for Visual Studio:
```cpp
        THREADNAME_INFO info = {};
        info.dwType = 0x1000;
        info.szName = name;
        info.dwThreadID = threadId;
        __try
        {
            RaiseException(MS_VC_EXCEPTION, 0, sizeof(info) / sizeof(ULONG_PTR), (ULONG_PTR*)&info);
        }
        __except (EXCEPTION_EXECUTE_HANDLER)
        {
        }
```
this too inadvertently results in Tracy's `CrashFilter` being invoked as currently written, however in this case it is not one of the recognized exceptions and returns `EXCEPTION_CONTINUE_SEARCH` which allows the frame-based filter to be called.

In all of these cases, my change to use `SetUnhandledExceptionFilter` causes `CrashFilter` to be invoked only in the event of an actual unhandled exception.

wolfpld: Thanks for the explanation, it does make sense.

My initial question was actually for the "Already had a handler => put it back" bit, which seems to deny enabling the crash handler, when something else is already hooked up. The `TRACY_NO_CRASH_HANDLER` macro is intended to handle this case.

jkriegshauser: I pushed a change that makes Windows now always set the unhandled exception filter, similar to how Linux always sets the signal handlers.

In a modern modular world the way these handlers work is not great, but such is the OS design that we have to work with. The fact that we have `TRACY_NO_CRASH_HANDLER` gives a work-around which is good.

I also changed the Linux side so that it matches the Windows removal-side now: if something else overrode our signal/SEH handler we put it back.

jkriegshauser:
Anything else I need to do for this?

wolfpld:
The general outlook of the changes is good, but I will still need to take a look into the details before the merge. For now, please add braces to scope where only indentation is used, or use single-liners, as you see fit.

public/client/TracyProfiler.cpp
```diff
@@ -1462,6 +1463,22 @@ Profiler::Profiler()
         m_userPort = atoi( userPort );
     }
 
+    m_safeSendBufferSize = 65536;
```
wolfpld, line 1466: This can be a contexpr and not a variable.

public/client/TracyProfiler.cpp
```diff
@@ -1462,6 +1463,22 @@ Profiler::Profiler()
...
+    m_safeSendBuffer = (char*)tracy_malloc( m_safeSendBufferSize );
+
+#ifndef _WIN32
+    pipe(m_pipe);
+#  if defined __APPLE__ || defined BSD
+    // FreeBSD/XNU don't have F_SETPIPE_SZ, so use the default 
+    m_pipeBufSize = 16384;
+#  else
+    m_pipeBufSize = (int)(ptrdiff_t)m_safeSendBufferSize;
```
wolfpld, line 1475: Why is double casting needed?

public/client/TracyProfiler.cpp
```diff
@@ -3063,6 +3097,62 @@ bool Profiler::CommitData()
...
 
+char* Profiler::SafeCopyProlog( const char* data, size_t size )
+{
+    bool success = true;
+    char* buf = m_safeSendBuffer;
+#ifndef NDEBUG
+    assert( !m_inUse.exchange(true) );
+#endif
+
+    if( size > m_safeSendBufferSize ) buf = (char*)tracy_malloc( size );
```
wolfpld, line 3108: What's the typical histogram of buffer sizes? How often is a larger temporary buffer allocated?

public/client/TracyProfiler.cpp
```diff
@@ -1498,20 +1517,29 @@ void Profiler::InstallCrashHandler()
...
     if( m_crashHandlerInstalled )
     {
-        sigaction( TRACY_CRASH_SIGNAL, &m_prevSignal.pwr, nullptr );
-        sigaction( SIGILL, &m_prevSignal.ill, nullptr );
-        sigaction( SIGFPE, &m_prevSignal.fpe, nullptr );
-        sigaction( SIGSEGV, &m_prevSignal.segv, nullptr );
-        sigaction( SIGPIPE, &m_prevSignal.pipe, nullptr );
-        sigaction( SIGBUS, &m_prevSignal.bus, nullptr );
-        sigaction( SIGABRT, &m_prevSignal.abrt, nullptr );
+        auto restore = [ this ]( int signum, struct sigaction* prev ) {
```
wolfpld, line 1531: Why is this captured?

public/client/TracyProfiler.cpp
jkriegshauser: I generally do this when both size and signedness are changing to (a) be explicit and (b) prevent pedantic warnings

public/client/TracyProfiler.cpp
jkriegshauser: Inadvertently left in--will remove with next push.

public/client/TracyProfiler.cpp
jkriegshauser: I did a test on a simple application and got the following:
  | 1 << ? | Count | Comment
-- | -- | -- | --
  | [0] | 0 | 
  | [1] | 0 | 
  | [2] | 8 | 
  | [3] | 85 | 
  | [4] | 280 | 
  | [5] | 685 | 
  | [6] | 863 | 
  | [7] | 917 | 
  | [8] | 856 | 
  | [9] | 799 | 
  | [10] | 619 | 
  | [11] | 449 | 
  | [12] | 239 | 
  | [13] | 124 | 
  | [14] | 58 | 
  | [15] | 34 | 
  | [16] | 17 | 
  | [17] | 2 | (exceeds buffer)

So at 64k in my simple test only 2 requests required temporary heap allocation as they exceeded the buffer size

wolfpld:
Thanks!

mncat77:
These changes seem to break the build with MinGW
```
tracy/public/client/TracyProfiler.cpp:1510:48: error: invalid conversion from 'LPTOP_LEVEL_EXCEPTION_FILTER' {aka 'long int (*)(_EXCEPTION_POINTERS*)'} to 'void*' [-fpermissive]
 1510 |     m_prevHandler = SetUnhandledExceptionFilter( CrashFilter );
 
tracy/public/client/TracyProfiler.cpp:3116:5: error: expected 'catch' before '__except'
 3116 |     __except( 1 /*EXCEPTION_EXECUTE_HANDLER*/ )
```

MSVC-specific code should probably be guarded with `_MSC_VER` instead of `_WIN32`

wolfpld:
Can you make a PR?

jkriegshauser:
```
tracy/public/client/TracyProfiler.cpp:1510:48: error: invalid conversion from 'LPTOP_LEVEL_EXCEPTION_FILTER' {aka 'long int (*)(_EXCEPTION_POINTERS*)'} to 'void*' [-fpermissive]
 1510 |     m_prevHandler = SetUnhandledExceptionFilter( CrashFilter );
```
This is easily fixable with a cast.

```
tracy/public/client/TracyProfiler.cpp:3116:5: error: expected 'catch' before '__except'
 3116 |     __except( 1 /*EXCEPTION_EXECUTE_HANDLER*/ )
```
This might be [more problematic](https://stackoverflow.com/questions/7244645/porting-vcs-try-except-exception-stack-overflow-to-mingw). You can see if `__try1` / `__except1` work otherwise it's going to be harder to work around.

wolfpld:
See #914.

PR #897, opened 2024-10-02
Added Custom GPU Context Type

Devaniti:
It is possible to implement custom integration in GPU zones submission
That works, but unfortunately you need to pretend to be one of existing integrations for such setup to work
This PR adds "Custom" GPU Context Type specifically for this usecase
It is by design will not be implemented in Tracy repo, and user is free to use it for custom integrations
image


wolfpld:
Protocol bump is required for this.

Devaniti:
Bumped protocol version to 71

PR #908, opened 2024-10-09
Fix upper limit of unique SourceLocations w.r.t. manual

slomp:
The manual mentions 65534 unique source locations:
https://github.com/wolfpld/tracy/blob/1bd84419c09dae2b5f9e2c8f5e29697e911fed38/manual/tracy.tex#L667

So it looks like the current assert is wrong:
https://github.com/wolfpld/tracy/blob/1bd84419c09dae2b5f9e2c8f5e29697e911fed38/server/TracyWorker.cpp#L3336

and should be:
```
assert( m_data.sourceLocationExpand.size() < std::numeric_limits<uint16_t>::max() );
```


slomp:
Ah, sorry, just realized about the split between static and dynamic. My bad.

PR #914, opened 2024-10-20
Fix MinGW build

mncat77:
@wolfpld: Just made changes to fix the build for now, should probably not be merged until it is tested properly - is there a standard procedure for that already?


wolfpld:
> should probably not be merged until it is tested properly - is there a standard procedure for that already?

There is nothing formalized for that.

mncat77:
> > should probably not be merged until it is tested properly - is there a standard procedure for that already?
> 
> There is nothing formalized for that.

Alright, will see if I can find more time to test this properly, should probably not be merged until it is tested properly by someone.

Issue #779, opened 2024-04-30
Android UE5: dlopen failed due to TLS model in tracy_rpmalloc.cpp


natepaynefb:
Hi. Thanks for the great profiler! I'm trying to use it in a UE5 project built for Android and deployed onto a Meta Queset VR headset. It crashes immediately with an error of the form:
```
java.lang.UnsatisfiedLinkError: dlopen failed: TLS symbol "(null)" in dlopened "<path>/libtracy.so" referenced from "<path>/libtracy.so" using IE access model
```

The dynamic linker is trying to handle the following variable from tracy_rpmalloc.cpp:
```
static _Thread_local heap_t* _memory_thread_heap TLS_MODEL;
```
where `TLS_MODEL` is defined as:
```
define TLS_MODEL __attribute__((tls_model("initial-exec")))
```

I can work around this problem by removing the `tls_model` attribute. Doing so fixes the crash and does not cause any other problems that I have been able to observe. However, I don't feel confident about this fix because I don't know why the specific TLS model was used in the first place.

I also tried compiling with linker flag `-fvisibility=hidden`. I was hoping that would hide the variable from the dynamic linker entirely, but it did not.

What would you suggest? Would the following revision to tracy_rpmalloc.cpp cause any problems that you are aware of?
```
#    if !defined(__HAIKU__) && !defined(__ANDROID__)
#      define TLS_MODEL __attribute__((tls_model("initial-exec")))
#    else
#      define TLS_MODEL
#    endif
```

In case it matters, here are the other defines used when building Tracy:
```
         -DTRACY_ENABLE=1,
        -DTRACY_FIBERS=1,
        -DTRACY_ON_DEMAND=1,
        -DTRACY_DELAYED_INIT=1,
        -DTRACY_MANUAL_LIFETIME=1,
        -DTRACY_EXPORTS=1,
```

Thanks for your help.

wolfpld:
It would be best to ask about this in rpmalloc: https://github.com/mjansson/rpmalloc

wolfpld:
339a92a7b2

PR #925, opened 2024-11-15
Fix crash on Android due to wrong TLS model

TroutZhang:
For android ndk, it is always a dynamic .so to let java load. 
So "initial-exec" would either not work, or crash for ndk 26+ with minSdk 29+.

check here: https://github.com/android/ndk/wiki/Changelog-r26#changes

[Issue 1679](https://github.com/android/ndk/issues/1679): Clang will now automatically enable ELF TLS for minSdkVersion 29 or higher.

### DIFF
--- a/public/client/tracy_rpmalloc.cpp
+++ b/public/client/tracy_rpmalloc.cpp
@@ -690,7 +690,9 @@ static pthread_key_t _memory_thread_heap;
 #    define _Thread_local __declspec(thread)
 #    define TLS_MODEL
 #  else
-#    ifndef __HAIKU__
+#    if defined(__ANDROID__) && __ANDROID_API__ >= 29 && defined(__NDK_MAJOR__) && __NDK_MAJOR__ >= 26
+#      define TLS_MODEL __attribute__((tls_model("local-dynamic")))
+#    elif !defined(__HAIKU__)
 #      define TLS_MODEL __attribute__((tls_model("initial-exec")))
 #    else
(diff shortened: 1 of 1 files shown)


TroutZhang:
Here is the PR for upstream: https://github.com/mjansson/rpmalloc/pull/347

TroutZhang:
Fixes https://github.com/wolfpld/tracy/issues/779

wolfpld:
How does this interact with previous Android SDKs, where the change was not needed?

TroutZhang:
> How does this interact with previous Android SDKs, where the change was not needed?

It seems previously the emulated TLS doesn't give error at all.

wolfpld:
Ah, it shouldn't matter, there are ifdefs against the SDK and NDK version.

TroutZhang:
My local fix is like the below:
```
#if defined(__ANDROID__) && __ANDROID_API__ >= 29 && defined(__NDK_MAJOR__) && __NDK_MAJOR__ >= 26
    #define TLS_MODEL __attribute__((tls_model("local-dynamic")))
#else
    #define TLS_MODEL
#endif
```

TroutZhang:
Basically, only one of our games, that has minSdk 29, triggered this.  Others are fine when ndk 25c => 27c.

So, only ndk 26+ and minSdk 29+ will enable hardware ELF TLS, hence the crash in dlopen from Java's `System.loadLibrary`

PR #926, opened 2024-11-16
Percentiles p75, p90 for FindZone view

WernherVonData:
Small contribution proposal from my side: during work on TPEG decoding messages I found it useful to have percentiles values for zone statistics. It helped me better understand the nature of the problem that I'm facing and show the gain from the improvements that I'm currently working on.

Didn't add P75 and P90 as a lines to the histogram. If you feel my contribution brings any value to core Tracy, then I can add them as well (don't use them much).

### DIFF
--- a/profiler/src/profiler/TracyView_FindZone.cpp
+++ b/profiler/src/profiler/TracyView_FindZone.cpp
@@ -549,10 +549,13 @@ void View::DrawFindZone()
                 std::inplace_merge( vec.begin(), mid, vec.end() );
 
                 const auto vsz = vec.size();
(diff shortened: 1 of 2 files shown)


profiler/src/profiler/TracyView_FindZone.cpp
```diff
@@ -549,10 +549,13 @@ void View::DrawFindZone()
                 std::inplace_merge( vec.begin(), mid, vec.end() );
 
                 const auto vsz = vec.size();
+                const auto vszQuarter = vsz / 4;
```
wolfpld, line 552: I don't think this variable is needed.

PR #931, opened 2024-11-27
Add absolute time info to Zone info view

eibach:
For correlation with other traces (like Wireshark or tracy from another process) I find it really helpful to have an absolute timestamp in the zone info view. As all required data is available, the implementation is straightforward.
For my purposes second granularity is fine.

### DIFF
--- a/profiler/src/profiler/TracyView_ZoneInfo.cpp
+++ b/profiler/src/profiler/TracyView_ZoneInfo.cpp
@@ -457,6 +457,8 @@ void View::DrawZoneInfoWindow()
         const auto ztime = end - ev.Start();
         const auto selftime = GetZoneSelfTime( ev );
         TextFocused( "Time from start of program:", TimeToStringExact( ev.Start() ) );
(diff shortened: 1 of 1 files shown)


profiler/src/profiler/TracyView_ZoneInfo.cpp
```diff
@@ -457,6 +457,8 @@ void View::DrawZoneInfoWindow()
         const auto ztime = end - ev.Start();
         const auto selftime = GetZoneSelfTime( ev );
         TextFocused( "Time from start of program:", TimeToStringExact( ev.Start() ) );
+        const std::time_t ts = m_worker.GetCaptureTime() + ev.Start() / 1000000000;
+        TextFocused( "Abs time:", std::asctime( std::localtime( &ts) ) );
```
wolfpld, line 461: Maybe "wall clock time"?

eibach:
Thanks. Tracy is an invaluable tool for my daily work at the moment.

PR #938, opened 2024-12-03
RFC: Add blob feature

eibach:
This adds the ability to trace the status of arbitrary blobs in memory.

This is an RFC only, to see if I am thinking into the right direction. The encoding parameter is meant to add some decoding ability in the profiler later on. Documentation and timeline integration is still missing completely.

### DIFF
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -2462,11 +2462,13 @@ Profiler::DequeueStatus Profiler::Dequeue( moodycamel::ConsumerToken& token )
                         break;
                     case QueueType::Blob:
                     case QueueType::BlobCallstack:
+                    {
                         ptr = MemRead<uint64_t>( &item->blobData.data );
-                        size = MemRead<uint16_t>( &item->blobData.size );
+                        auto size = MemRead<uint32_t>( &item->blobData.size );
                         SendBlob( (const char*)ptr, size );
                         tracy_free_fast( (void*)ptr );
                         break;
+                    }
                     case QueueType::ZoneBeginAllocSrcLoc:
                     case QueueType::ZoneBeginAllocSrcLocCallstack:
                     {
@@ -2998,7 +3000,7 @@ Profiler::DequeueStatus Profiler::DequeueSerial()
                 {
                     ThreadCtxCheckSerial( blobDataThread );
                     ptr = MemRead<uint64_t>( &item->blobData.data );
-                    uint16_t size = MemRead<uint16_t>( &item->blobData.size );
+                    uint32_t size = MemRead<uint32_t>( &item->blobData.size );
                     SendBlob( (const char*)ptr, size );
                     tracy_free_fast( (void*)ptr );
                     break;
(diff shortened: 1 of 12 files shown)


wolfpld:
What is the intended use case?

Have you tried sending "large" blobs, i.e. larger than `enum { TargetFrameSize = 256 * 1024 }` bytes?

eibach:
My use case is recording structs or packets on the fly so I can inspect them later. Intended size is << 64k, with 64k as upper limit.

wolfpld:
I'm not sure about imposing a limit. You may want instead to send the 3D scene geometry for debug purposes, and that could be easily in the megabytes range. (I actually did such a thing: https://bitbucket.org/wolfpld/tracy/commits/branch/mesh-debug)

eibach:
Ok, I see. I have used the string infrastructure because I thought that identical blobs should only get stored and transferred once. Do you think that would also make sense for bigger blobs like the geometry from your usecase?

wolfpld:
Sorry, I don't remember the exact details of how things are transferred, so I can't comment on the specifics.

Hashing things for storage shouldn't be a problem, you can get tens of GB/s of hashing perf e.g. with xxh3.

Hashing things for transfer looks like a no-go for me, as you'd need to do it on the client, and doing extra calculations when you are profiling and want to keep impact minimal is something you'd generally want to avoid.

eibach:
Ah yes, you are right, hashing only is done on the server side. I got that mixed up. I will check if I hit any limits when I crank up blob size.

eibach:
Works only up to 64k - 1 bytes using the string infrastructure. So maybe I have to come up with a different transfer strategy...

wolfpld:
Frame image transfers are another example, but still limited to ~256 KB. Some sort of data part transfer would be needed, something like `Worker::QueryDataTransfer()` is doing, but in reverse.

(For some reason I'm not able to add a comment on GitHub.)

eibach:
Alright, I will see if I can get some inspiration from `Worker::QueryDataTransfer()`

eibach:
I am struggling with the implementation of `Worker::QueryDataTransfer()`. Why is everything divided in pieces up to 12 bytes and not larger chunks? Doesn't that generate a lot of overhead? Probably I am missing the point here.

eibach:
This is my idea for the client side. If you agree, I would start implementing the server side.

public/client/TracyProfiler.cpp
```diff
@@ -3245,6 +3267,23 @@ void Profiler::SendLongString( uint64_t str, const char* ptr, size_t len, QueueT
...
+void Profiler::SendBlob( const char* ptr, size_t len )
+{
+    QueueItem item;
+    MemWrite( &item.hdr.type, QueueType::BlobFragment );
+    while (len)
+    {
+        uint32_t fragment_size = len > TargetFrameSize ? TargetFrameSize : len;
+        len -= fragment_size;
+
+        NeedDataSize( QueueDataSize[(int)QueueType::BlobFragment] + sizeof( fragment_size ) + fragment_size );
```
wolfpld, line 3279: Shouldn't fragment_size take these additional sizes into account when comparing against TargetFrameSize?

public/client/TracyProfiler.hpp
```diff
@@ -484,6 +484,28 @@ class Profiler
...
+        if( callstack != 0 )
+        {
+            tracy::GetProfiler().SendCallstack( callstack );
+        }
+
+        auto ptr = tracy_malloc( size );
+        memcpy( ptr, data, size );
+
+        TracyQueuePrepare( callstack == 0 ? QueueType::Blob : QueueType::BlobCallstack );
+        MemWrite( &item->blobData.time, GetTime() );
```
wolfpld, line 502: Time value needs to be send as a delta over the wire in Dequeue().

public/common/TracyQueue.hpp
```diff
@@ -395,6 +398,23 @@ struct QueueMessageColorFatThread : public QueueMessageColorFat
...
+struct QueueBlob
+{
+    int64_t time;
+};
+
+struct QueueBlobData : public QueueBlob
+{
+    uint64_t encoding;
+    uint64_t data;      // ptr
+    uint16_t size;
```
wolfpld, line 410: 16 bits is probably too small.

public/common/TracyQueue.hpp
```diff
@@ -395,6 +398,23 @@ struct QueueMessageColorFatThread : public QueueMessageColorFat
...
 };
 
+struct QueueBlob
+{
+    int64_t time;
+};
+
+struct QueueBlobData : public QueueBlob
+{
+    uint64_t encoding;
```
wolfpld, line 408: What is the meaning of the encoding values? The TracyBlob() macro requires setting a value. How will user know what to set there? If "the encoding parameter is meant to add some decoding ability in the profiler later on", how will this future functionality be able to work, when there is no "restricted for future use" set of values that shouldn't be used by users?

public/client/TracyProfiler.hpp
eibach: This implementation is taken from the Message...() functions. As far as my understanding goes, it is converted in TracyWorker using the TscTime() function.

public/common/TracyQueue.hpp
eibach: I am not completely done with this. My (totally untested) idea was to have a facility for executing user supplied dissectors, like in Wireshark, maybe even Lua based. This facility could use the encoding value to choose the dissector.

public/client/TracyProfiler.cpp
eibach: Done

public/common/TracyQueue.hpp
eibach: Done

public/client/TracyProfiler.hpp
wolfpld: The missing processing step is done in Deqeue(). Look for refCtx for examples of this.

eibach:
I have added support for fragmented transfer of blob data. Now uint32_t should be the limit ;)

wolfpld:
Why is `m_serverQuerySpaceLeft` increased? The client should send all the blob fragments in succession, and the server doesn't make any queries?

PR #939, opened 2024-12-10
Display zone values also in hex

eibach:
For better readability display zone values not only in decimal but also in hex.


wolfpld:
Can you provide screenshots of how this looks?

eibach:
image

PR #941, opened 2024-12-12
[Meson][Windows] detect MSVC usage

UnixY2K:
previous check tested for compiler ID in meson, detection of an specific MSVC macro is preferred so it works in any compiler that includes the MSVC stdlib.

I noticed this error while trying to compile the `0.11.0` version

- changes the compiler check to macro check `_MSC_VER` which is an [Microsoft specific macro](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170#microsoft-specific-predefined-macros)
- changes the current defined version to `0.11.1`, previous was `0.11.0`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('tracy', ['cpp'], version: '0.11.0', meson_version: '>=1.1.0')
+project('tracy', ['cpp'], version: '0.11.1', meson_version: '>=1.3.0', default_options : ['cpp_std=c++11'])
 
 # internal compiler flags
 tracy_compile_args = []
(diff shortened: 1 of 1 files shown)


wolfpld:
Wouldn't it be better to just remove that whole condition and add a `default_options: ['cpp_std=c++11']` parameter to the project definition?

UnixY2K:
> Wouldn't it be better to just remove that whole condition and add a `default_options: ['cpp_std=c++11']` parameter to the project definition?

the issue is that MSVC stdlib is not fully C++11 compliant, so compilation will fail in Windows unless C++14 is used.

I set the std by default to C++11 and increased the meson version as I noticed some warnings with the new macro check, any other platform will just use C++11 by default.

PR #957, opened 2024-12-27
Simplify API by decreasing a number of entry points

### LINKED ISSUE #954: Simplify API by merging callstack / no callstack functions
The two following functions are very close to each other:
https://github.com/wolfpld/tracy/blob/6452690e575a17c11962f2245e5ee3d73f807525/public/client/TracyProfiler.cpp#L4247-L4273

https://github.com/wolfpld/tracy/blob/6452690e575a17c11962f2245e5ee3d73f807525/public/client/TracyProfiler.cpp#L4275-L4302

I think they can be merged into one. With depth=0 (or with negative integers), callstack can be omitted. It will simplify TracyC.h/Tracy.hpp files by avoiding replication code for cases when callstack is available and is unavailable.

foxtran:
I have removed most of TRACY_HAS_CALLSTACK in C/C++ public headers with implementing related constexpr (if possible) checks in functions' implementations. 

I have also set TRACY_CALLSTACK to zero if it is not set. It is set to zero, stack will not be collected. I hope compiler will be able to determine that corresponding branches are always false and remove related code. 

Note: only C/C++ headers are updated. I did not touch other APIs which can be also reduced in the same way.

Fixes #954


wolfpld:
Can you show disassembly of the code that would use functions you have changed? I don't really want the if depth > 0 check to be present, if it's known at compile time it's not needed.

foxtran:
I do not know how to show it in a good way...

Anyway, there is a not a simple example: https://godbolt.org/z/5h7jz5dzz

I have inserted `exit(depth)` calls into public/client/TracyScoped.hpp in two branches. As you can see, it will generate an extremely simple code that will only pass 0 or 5 (I set this depth) into exit call with -O3 optimization level. With -DCMAKE_BUILD_TYPE=Debug, TRACY_CALLSTACK=0 will get a cost but starting from -O1 it will work with a simple asm :)

public/client/TracyProfiler.cpp
```diff
@@ -4291,9 +4291,16 @@ TRACY_API TracyCZoneCtx ___tracy_emit_zone_begin_callstack( const struct ___trac
         TracyQueueCommitC( zoneValidationThread );
     }
 #endif
-    tracy::GetProfiler().SendCallstack( depth );
-    {
-        TracyQueuePrepareC( tracy::QueueType::ZoneBeginCallstack );
+    if (depth > 0 && tracy::has_callstack()) {
+        tracy::GetProfiler().SendCallstack( depth );
```
wolfpld, line 4295: Please adapt to the coding style (and in similar places).

public/client/TracyProfiler.cpp
```diff
@@ -4291,9 +4291,16 @@ TRACY_API TracyCZoneCtx ___tracy_emit_zone_begin_callstack( const struct ___trac
...
+        {
+            TracyQueuePrepareC( tracy::QueueType::ZoneBeginCallstack );
+            tracy::MemWrite( &item->zoneBegin.time, tracy::Profiler::GetTime() );
+            tracy::MemWrite( &item->zoneBegin.srcloc, (uint64_t)srcloc );
+            TracyQueueCommitC( zoneBeginThread );
+        }
```
wolfpld, line 4301: This extra scope is not needed.

public/client/TracyProfiler.cpp
```diff
@@ -4291,9 +4291,16 @@ TRACY_API TracyCZoneCtx ___tracy_emit_zone_begin_callstack( const struct ___trac
...
+    } else {
+        TracyQueuePrepareC( tracy::QueueType::ZoneBegin );
         tracy::MemWrite( &item->zoneBegin.time, tracy::Profiler::GetTime() );
         tracy::MemWrite( &item->zoneBegin.srcloc, (uint64_t)srcloc );
         TracyQueueCommitC( zoneBeginThread );
```
wolfpld, line 4306: There is shared code between the branches, maybe it should be moved out of the condition?

foxtran: `TracyQueuePrepareC` is a macro and can generate variables which can be destroyed at the end of scope. At least I do not see a definition of item variable. I can merge both branches with ternary expression, but it will be overcomplicated.

public/client/TracyProfiler.cpp
foxtran: Done :)

PR #961, opened 2025-01-02
Use `int32_t` type instead of `int`

### LINKED ISSUE #953: C API: avoid usage of `int` type
Hello! 

In some places of C API `int` type is used which may be not consistent between compilers (usually, it is). For example, here:
https://github.com/wolfpld/tracy/blob/6452690e575a17c11962f2245e5ee3d73f807525/public/tracy/TracyC.h#L160-L165

It would be nice to use consistent `int32_t` (or even `bool`, where it is necessary) instead of `int`.

foxtran:
This patch updates C API (and others too) for using `int32_t` type. After the patch,  tracy should be more portable for cases if `int` type has a size different from 4 bytes.   

Fixes #953

### DIFF
--- a/public/client/TracyCallstack.hpp
+++ b/public/client/TracyCallstack.hpp
@@ -10,7 +10,7 @@
 namespace tracy
 {
 static constexpr bool has_callstack() { return false; }
-static tracy_force_inline void* Callstack( int /*depth*/ ) { return nullptr; }
+static tracy_force_inline void* Callstack( int32_t /*depth*/ ) { return nullptr; }
 }
 
 #else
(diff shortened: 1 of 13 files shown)


wolfpld:
Why does this matter? I don't think there is an expectation for int to be 32 bit in the cases you have changed (I have not looked thoroughly though).

Also, why have you added explicit conversion between bool and int?

foxtran:
Ooops. I forget about Python API.

wolfpld:
> Unfortunately, one of compilers have its own understanding what is the size of `int` type in C (it can be changed between 32 or 64 bits by compiler flag), while normally it is 32 bit for GCC/LLVM. So, compiling of Fortran-app+Tracy using this compiler will not produce a binary which will work correctly if `int` type is 64 bit long in Fortran side and Tracy was compiled with GCC/LLVM.

This seems to me like a defect in the compiler? I would expect the compiler to follow the 64 bit data model of the platform it targets, instead of randomly selecting a model that will produce binaries that do not match the ABI.

foxtran:
> This seems to me like a defect in the compiler?

It is a feature. It switches between LP64/ILP64 models for both, Fortran and C, languages and it does not care about linking with binaries produced by other compilers.

PR #962, opened 2025-01-03
Add initial native Fortran support 

foxtran:
This PR adds support of Fortran language. Most of features which available in C should be available in Fortran.

Fortran does not support mutexes/locks, so Fortran API also does not support them.

Uses API from #961 

TODO:
- [x] Add documentation
- [x] Add Tracy colors
- [x] ~~Add GPU support~~ (not in this PR)

Tested with:
- Classic Intel Fortran compiler
- Intel LLVM Fortran compiler
- GCC 14.1 (x86_64)
- Oracle Sun Studio 12.6
- Cray Fortran 16
- AOCC 4.1
- NVFortran 23.11


foxtran:
@wolfpld, finally it works for Fortran with OpenMP :)

It is one of [MRCC (Quantum Chemistry software)](http://mrcc.hu) module. 

It does not have all possible info since I'm using an HPC cluster with Android lunary problem :(

image

PR #967, opened 2025-01-11
Add cross-project LTO

foxtran:
This PR adds `TRACY_LTO` flag that enables Link-Time optimizations for Tracy. It should improve performance of Tracy calls.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,17 +15,30 @@ endif()
 
 option(TRACY_STATIC "Whether to build Tracy as a static library" ${DEFAULT_STATIC})
 option(TRACY_Fortran "Build Fortran bindings" OFF)
+option(TRACY_LTO "Enable Link-Time optimization" OFF)
 
 if(TRACY_Fortran)
   enable_language(Fortran)
   set(CMAKE_Fortran_VERSION 2003)
 endif()
 
+if(TRACY_LTO)
+    include(CheckIPOSupported)
+    check_ipo_supported(RESULT LTO_SUPPORTED)
+    if(NOT LTO_SUPPORTED)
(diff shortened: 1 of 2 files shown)


wolfpld:
I would expect users to set `set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)` in their build scripts.

foxtran:
Yep, setting `CMAKE_INTERPROCEDURAL_OPTIMIZATION ` is more convenient way, however not all projects can be compiled with LTO (compilation time, linking issues, compiler bugs, and so on), so it would be useful to provide an LTO option for users who can not use LTO on whole project.

CMakeLists.txt
```diff
@@ -8,31 +8,45 @@ project(Tracy LANGUAGES CXX VERSION ${TRACY_VERSION_STRING})
...
+elseif(TRACY_STATIC)
+    set(TRACY_VISIBILITY "STATIC")
 else()
-	set(TRACY_VISIBILITY "SHARED")
+    set(TRACY_VISIBILITY "SHARED")
 endif()
 
 add_library(TracyClient ${TRACY_VISIBILITY} "${TRACY_PUBLIC_DIR}/TracyClient.cpp")
 target_compile_features(TracyClient PUBLIC cxx_std_11)
+set_target_properties(TracyClient PROPERTIES INTERPROCEDURAL_OPTIMIZATION ${LTO_SUPPORTED})
```
wolfpld, line 49: This would interfere with the global option.

foxtran: Fixed by https://github.com/wolfpld/tracy/pull/967/commits/b87a935c58a89ec9f7b73b5a3c88b7f9b7f23fbe

Issue #966, opened 2025-01-07
Profiling MPI applications with Tracy can cause long spikes in non-blocking send/receive operations


psalz:
We are using Tracy to profile our distributed GPU runtime system [Celerity](https://github.com/celerity/celerity-runtime), and it's mostly working great. However, during some recent benchmarking runs on the Leonardo supercomputer we've noticed that traces often contain very long spikes for MPI non-blocking send / receive operations, with some transfers taking several thousand times longer than they should (e.g. 30ms instead of 10us, sometimes even > 100ms).

Here's an example trace for a run with 32 ranks. Notice how there are several small gaps throughout the run and a few very long ones towards the end, caused by long transfers (in the "p2p" fibers at the very bottom). The application is a simple stencil code executed over 10000 iterations, with each iteration performing exactly the same set of operations (point to point transfers between ranks, some copies as well as GPU kernel executions).

![image](https://github.com/user-attachments/assets/b7233d65-358f-4baf-8527-e1829b94eeee)

**Long story short, it turns out that those spikes only happen while profiling with Tracy**, and therefore seem to be due to some unfortunate interaction between the Tracy client and MPI.

What is very curious is that the gaps happen at seemingly predictable phases of the program's execution.

<details>
<summary>Here is another trace of the same application / configuration. Notice how the pattern of gaps looks very similar, although in this case the long gap towards the end is quite a bit shorter.</summary>

![image](https://github.com/user-attachments/assets/0f50e7e2-aa8a-4c78-9bb7-afc87397c23c)

</details>

I've managed to create a small-ish reproducer program, in case anyone is interested:

```c++
#include <numeric>
#include <optional>
#include <functional>
#include <algorithm>
#include <array>
#include <vector>
#include <chrono>
#include <cstdio>
#include <cstdlib>
#include <thread>
#include <deque>

#include <mpi.h>
#include <tracy/Tracy.hpp>
#include <tracy/TracyC.h>

using clk = std::chrono::steady_clock;
using namespace std::chrono_literals;

int main(int argc, char* argv[]) {
	const size_t transfer_bytes = argc > 1 ? std::atol(argv[1]) : 16384 * 4;
	const size_t iterations = 10000;
	const size_t warmup = 100;

	int provided = -1;
	MPI_Init_thread(nullptr, nullptr, MPI_THREAD_MULTIPLE, &provided);

	int rank = -1;
	int size = -1;
	MPI_Comm_rank(MPI_COMM_WORLD, &rank);
[...]

PR #968, opened 2025-01-14
Add suspend/resume profiler

### LINKED ISSUE #952: C API: manual lifetime: pause and resume profiler
Hello!

I have found two methods `___tracy_startup_profiler` and `___tracy_shutdown_profiler`. It would be nice to expand them with pause/resume methods for applications which produces a lot of profiler data. 

For implementation, pause method may just skip data collection and resume will enable data collection back

foxtran:
This patch implements suspend/resume calls for profiler that stops generating data for server. It is useful for long-living applications which can produce extremely huge data.

Users must be careful about usage TracySuspend/TracyResume inside of zones.

I have fixed couple bugs with proper exit of Tracy client.

Closes #952


wolfpld:
I don't really see how this could work correctly in a multithreaded application.

foxtran:
It just stops collecting of data from Tracy calls for whole application, so it works fine.

For example, there is 4-threaded OpenMP application, which I'm using for testing:

image

Client does not get Tracy events and therefore it is just a long bar comes after TracySuspend call. When TracyResume was called, Tracy works as before.

P.S. I still did not get why main thread does not have OpenMP sections :(

wolfpld:
> It just stops collecting of data from Tracy calls for whole application, so it works fine.

This is exactly why it won't work fine.

T1: suspend profiling
T2: enter zone, ignore event
T1: resume profiling
T2: leave zone, send event

At this point the state on client and server is desynchronized and things can't work properly.

foxtran:
Yep, that is true. Luckily, some applications (mostly, HPC) use control-flow thread only on which suspending/resuming can happen. 
So,
T1: suspends profiling
T1: spawns T2-Tn
T1-Tn: do collective work
T1-Tn: may enter and exit from zones (according to zone scope rules)
T2-Tn: die (actually, just sleeping)
T1: resumes profiling

foxtran:
If one will add some unique info about zones that would be unique for each zone (even the same in cycle), it will be possible to improve understanding by GUI app, that zone was created in suspend mode, but exited in resumed. Unfortunately, it will not help, if zone was created in resumed mode and ended in suspended.

wolfpld:
It seems to me like what you are looking for here is the `active` parameter in the macros.

foxtran:
`active` is not a solution, since it still produce a lot of data (mostly from callstacks, I think).

wolfpld:
Callstack collection should be paused when on demand mode is used and no connection is established. You should be able to extend this to you use case.

foxtran:
[These is an example how I'm using Tracy on HPC cluster.](https://github.com/wolfpld/tracy/issues/966#issuecomment-2585418424) It would be a bit hard to detect a right time when to start data collection (more specifically, start tracy-capture) and when to stop for avoiding out-of-memory errors from logs of application. I think I can start tracy-capture from my app for debugging itself :-)

slomp:
> active is not a solution, since it still produces a lot of data

Are you sure? If the `active` argument is false, it's an early out and nothing gets sent:
https://github.com/wolfpld/tracy/blob/8e388d6d4edacd579a2486fd8aae6ad53f8628b3/public/client/TracyScoped.hpp#L33

PR #973, opened 2025-01-20
Use correct header on NetBSD.

0-wiz-0:
Fixes
In file included from .../public/TracyClient.cpp:14:
.../public/common/TracySystem.cpp: In function ‘uint32_t tracy::detail::GetThreadHandleImpl()’:
.../public/common/TracySystem.cpp:79:12: error: ‘_lwp_self’ was not declared in this scope
   79 |     return _lwp_self();
      |            ^~~~~~~~~


0-wiz-0:
That was fast, thank you!

PR #989, opened 2025-02-23
Fix a build of OpelCLVectorAdd example on linux

FrK5E:


wolfpld:
Why is this needed?

FrK5E:
Whilst compiling the example with gcc on Ubuntu, had to include <cmath>. because of sqrt.

wolfpld:
Change cmath to math.h.

PR #1004, opened 2025-03-08
wayland: Activate window with `XDG_ACTIVATION_TOKEN`.

bbb651:
Tested on niri with:
```kdl
debug {
    strict-new-window-focus-policy
}
```


wolfpld:
And what is this doing?

bbb651:
This allows tracy to gain focus when it starts, `XDG_ACTIVATION_TOKEN` is set by spawn binds, launchers, etc. (currently not terminals). In practice compositors have a "smart" focus prevention mode that gives focus to newly created windows regardless, which does. Gnome has a similar off-by-default strict focus mode and are also interested in improving support for this, see [this mutter blog post](https://blogs.gnome.org/shell-dev/2024/09/20/understanding-gnome-shells-focus-stealing-prevention/).

PR #1016, opened 2025-03-28
Fix: Hang on shutdown with enabled system tracing

6yry6e:
The issue occur when the client wants to shutdown profiling but transfer all data to the server before exiting.
The client code looks like it:
`tracy::GetProfiler().RequestShutdown();`
`while (!tracy::GetProfiler().HasShutdonwFinished())`
`{`
`    yield...`
`}`

If the client doesn't send any zones/data to the queue, it works fine. But when system tracing is enabled it constantly fills queue with new data, the queue may never become empty, so we never reach exit conditions and the app hangs.

It look like solution to that problem is shutting down system tracing before sending remaining data to the server.
Another possible fix is to check `ShouldExit` in the systrace callback, but that's appear to me as a waste of cycles.

### DIFF
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -1155,6 +1155,38 @@ static void CrashHandler( int signal, siginfo_t* info, void* /*ucontext*/ )
 }
 #endif
 
+#ifdef TRACY_HAS_SYSTEM_TRACING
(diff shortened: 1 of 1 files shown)


Xnuman:
@wolfpld any updates on this review? It would be nice if this were merged

PR #1021, opened 2025-04-08
Thread wakeup visualization

Lectem:
This PR does the following things:
- Allow to lock the highlighted thread using left mouse click or lane popup menu
- Add more descriptions for wait reasons in View::DecodeContextSwitchReason
- Stores the Wakeup CPU so that we may display it in the CPU data view
- `QueueThreadWakeup` now contains cpu and adjust info (currently unused). Fits in the padding, no struct size change.
- `QueueContextSwitch` now also contains thread priorities information (windows+linux) and CState (windows), but currently unused by the profiler. Fits in the padding, no struct size change.
- Fixes Linux `EventWakeup` which tries to remove PERF_SAMPLE_CALLCHAIN on the wrong member. Set the whole `perf_event_attr` explicitely instead.
- Fixes Windows ReadyThread event being dropped if triggered before the thread switches out (this is confirmed to be normal behaviour, probably due to thread being switched out but context switch actually happening later)
- Bumps tracy file version and protocol versions

Screenshots of the wakeup viz:

image
image

In the future we should be able to retrieve the readying thread information, such as its callstack, by iterating the per-core context switch data. Ideally we would also have a way to draw an arrow between the thread lanes and not just the CPU cores data ?


profiler/src/profiler/TracyTimelineItemThread.cpp
```diff
@@ -144,7 +144,7 @@ int64_t TimelineItemThread::RangeEnd() const
 
 void TimelineItemThread::HeaderTooltip( const char* label ) const
 {
-    m_view.HighlightThread( m_thread->id );
+    m_view.HighlightThread(m_thread->id, false);
```
wolfpld, line 147: Please use the correct code style across the changes.

public/common/TracyVersion.hpp
```diff
@@ -6,8 +6,8 @@ namespace tracy
 namespace Version
 {
 enum { Major = 0 };
-enum { Minor = 11 };
-enum { Patch = 2 };
+enum { Minor = 12 };
```
wolfpld, line 9: Keep minor at 11.

wolfpld:
Please update the manual to explain the changes you've made. As it now is I don't really know what the screenshots show. Having a proper writeup will help with code review.

profiler/src/profiler/TracyTimelineItemThread.cpp
Lectem: Done, sorry !

Lectem:
I added a new paragraph and illustration to the manual (pages 78-79 for me).
Please tell me what you think about it.

manual/tracy.tex
```diff
@@ -3542,9 +3542,45 @@ \subsubsection{Timeline view}
...
 \end{itemize}
 
-When the \faMousePointer{}~mouse pointer is hovered over either the CPU data zone or the thread timeline label, Tracy will display a line connecting all zones associated with the selected thread. This can be used to quickly see how the thread migrated across the CPU cores.
+When the \faMousePointer{}~mouse pointer is hovered over either the CPU data zone or the thread timeline label, Tracy will display a line connecting all zones associated with the selected thread. This can be used to quickly see how the thread migrated across the CPU cores. 
 
-Clicking the \LMB{}~left mouse button on a tracked thread will make it visible on the timeline if it was either hidden or collapsed before.
+It will also add lines starting with a filed circle to denote wake up events. Those are useful to pinpoint the origin of a thread waking up, especially when holding locks. It may also start from an empty region, denoting the time at which the kernel made chose to schedule or boost the priority of your thread. Wake ups will have a different color based on the reason for which the thread was waiting to be scheduled.
```
wolfpld, line 3547: > Wake ups will have a different color based on the reason for which the thread was waiting to be scheduled. 

Is there a way to know what these colors will be?

wolfpld, line 3547: > made chose to

made choice to

manual/tracy.tex
```diff
@@ -3542,9 +3542,45 @@ \subsubsection{Timeline view}
...
+
+\draw (12,12) -- (14,13.5);
+	\end{tikzpicture}
+	\caption{Thread migration and wake up.}
+	\label{plot}
+\end{figure}
+
+In the above picture, \emph{Thread B} migrates from \emph{Core 3} to \emph{Core 4} due to a wake up from \emph{Thread A}. Then it migrates from \emph{Core 4} to \emph{Core 1}.
+
+Clicking the \LMB{}~left mouse button on a tracked thread will make it visible on the timeline if it was either hidden or collapsed before. It will also lock the selected thread so that you may move explore data while retaining the visualization of thread migrations and wake up events.
```
wolfpld, line 3583: > you may move explore data

pan explored

manual/tracy.tex
Lectem: Right now not really, but the "wait reason" can be seen from the thread lanes. The idea behind coloring it was to quickly see the different kinds of wait reason, but I think in the long term we might want to categorize each wait reason and regroup them ?

Lectem:
Fixed the latest comments (I think), and also added 

> Clicking again somewhere empty on the timeline with the left mouse button will unlock the selection.

profiler/src/profiler/TracyTimelineItemThread.cpp
```diff
@@ -275,6 +275,14 @@ void TimelineItemThread::DrawOverlay( const ImVec2& ul, const ImVec2& dr )
     m_view.DrawThreadOverlays( *m_thread, ul, dr );
 }
 
+void TimelineItemThread::DrawExtraPopupItems()
+{
+    if( ImGui::MenuItem( ICON_FA_TIMELINE " Select in CPU timeline" ) )
```
wolfpld, line 280: Is this also shown when there is no CPU timeline?

server/TracyEvent.hpp
```diff
@@ -530,36 +530,95 @@ struct CrashEvent
...
+        Win32_WrFastMutex       = 34,
+        Win32_WrGuardedMutex    = 35,
+        Win32_WrRundown         = 36,
+        Win32_WrAlertByThreadId = 37,
+        Win32_WrDeferredPreempt = 38,
+        Win32_WrPhysicalFault   = 39,
+        Win32_WrIoRing          = 40,
+        Win32_WrMdlCache        = 41,
+        Win32_WrRcu             = 42,
+        Win32_MaximumWaitReason,
```
wolfpld, line 588: Where is this list sourced from?

server/TracyEvent.hpp
```diff
@@ -530,36 +530,95 @@ struct CrashEvent
...
+    tracy_force_inline void SetState( int8_t state ) { _state = state; }
     tracy_force_inline int64_t WakeupVal() const { return _wakeup.Val(); }
     tracy_force_inline void SetWakeup( int64_t wakeup ) { assert( wakeup < (int64_t)( 1ull << 47 ) ); _wakeup.SetVal( wakeup ); }
     tracy_force_inline uint16_t Thread() const { return _thread; }
     tracy_force_inline void SetThread( uint16_t thread ) { _thread = thread; }
 
-    tracy_force_inline void SetStartCpu( int64_t start, uint8_t cpu ) { assert( start < (int64_t)( 1ull << 47 ) ); _start_cpu = ( uint64_t( start ) << 16 ) | cpu; }
-    tracy_force_inline void SetEndReasonState( int64_t end, int8_t reason, int8_t state ) { assert( end < (int64_t)( 1ull << 47 ) ); _end_reason_state = ( uint64_t( end ) << 16 ) | ( uint64_t( reason ) << 8 ) | uint8_t( state ); }
+    tracy_force_inline void SetStartCpu( int64_t start, uint8_t cpu ) { assert( start < (int64_t)( 1ull << 47 ) ); _start.SetVal( start ); _cpu = cpu; }
+    tracy_force_inline void SetEndReasonState( int64_t end, int8_t reason, int8_t state ) { assert( end < (int64_t)( 1ull << 47 ) ); _end = end; _reason = reason; _state = state; }
```
wolfpld, line 610: These functions existed only to have the setting operation as a single store. With the way data is now split it makes no sense to keep these.

server/TracyWorker.cpp
```diff
@@ -6772,9 +6781,12 @@ void Worker::ProcessContextSwitch( const QueueContextSwitch& ev )
             auto& item = data.back();
             assert( item.Start() <= time );
             assert( item.End() == -1 );
+            //TODO: it may happen that events are being dropped (for example due to breaking in the debugger, or we are simply too slow to handle the events)
+            //      We should handle this properly in some way, but it is unclear how. We can't even really detect it properly other than when cpu doesn't match.
+            assert( item.Cpu() == ev.cpu );
```
wolfpld, line 6786: Yeah, it will happen. This shouldn't be an assert, because some people will run with no asserts and it will be ignored anyways, and other people will complain that the profiler crashes.

server/TracyEvent.hpp
Lectem: This comes from the windows driver kit, though not documented online you can find the Rust versions here https://microsoft.github.io/windows-docs-rs/doc/windows/Wdk/System/SystemServices/constant.WrIoRing.html or from the Process Hacker docs https://ntdoc.m417z.com/kwait_reason

server/TracyWorker.cpp
Lectem: Should I just remove the assert and ignore the issue or do we actually want to generate a new context switch?
It's not really clear how to fill the data if we want to generate a new cswitch. (what time do we use for end of the previous cswitch, and for start of the new one).

wolfpld: It's all guesswork when data is missing. The capture should ideally be able to eventually recover proper state after more data is received.

Lectem: I suppose we could detect dropped events from ETW display it somehow in the UI but it seems out of scope for this PR. Will leave the comment (with a bit more detail if needed) but remove the assert, is that OK with you? (ie, same behavior as before, but documented)

profiler/src/profiler/TracyTimelineItemThread.cpp
Lectem: Good point, I'll try to make it conditional

server/TracyWorker.cpp
wolfpld: It's ok.

wolfpld:
Lines on the CPU data graph no longer work as intended.

Correct behavior:

obraz

After moving the leftmost element out of view:

obraz

wolfpld:
Elements too small to be visible are not culled, which results in an unreadable display and severely degraded performance.

obraz

wolfpld:
Selecting a thread on the CPU data list hijacks the highlight functionality. Threads hovered over with the mouse will not be highlighted until the selection is removed. These should be two separate things.

wolfpld:
I want to see how the two wakeup colors differ, but there is no explanation in the tooltip.

obraz

Lectem:
> Lines on the CPU data graph no longer work as intended.

Oversight on my part due to wrong using the worng signedness for the time. Fixed.

> The vertical lines are drawn on top of the above and below CPU lanes

This was intentional as I felt it looked better, but I changed it to fit the height of the core.

> Elements too small to be visible are not culled, which results in an unreadable display and severely degraded performance.

I am now only displaying those if the wait duration spans more than 30px or if the element would be > 0.5px. That way we hit a nice spot between usability and performance. I'm open to suggestions if you think there are better ways to do it.

> I want to see how the two wakeup colors differ, but there is no explanation in the tooltip.

I added information about the waits and wakeups in the tooltip

> Selecting a thread on the CPU data list hijacks the highlight functionality. Threads hovered over with the mouse will not be highlighted until the selection is removed. These should be two separate things.

This brings a few questions:
If I select a thread using the click on the timeline, then hover another (from either the cpu data timeline or the thread lane)
- Should I display both? 
- Should the hovered one have priority over the selected one?
- Should I fade out the selected one?

wolfpld:
> This brings a few questions: If I select a thread using the click on the timeline, then hover another (from either the cpu data timeline or the thread lane)
> 
>     * Should I display both?
> 
>     * Should the hovered one have priority over the selected one?
> 
>     * Should I fade out the selected one?

Display both, with different colors. See GPU zones or locks for reference.

profiler/src/profiler/TracyTimelineItemThread.cpp
```diff
@@ -275,6 +275,14 @@ void TimelineItemThread::DrawOverlay( const ImVec2& ul, const ImVec2& dr )
     m_view.DrawThreadOverlays( *m_thread, ul, dr );
 }
 
+void TimelineItemThread::DrawExtraPopupItems()
+{
+    if( m_view.GetViewData().drawCpuUsageGraph && ImGui::MenuItem( ICON_FA_TIMELINE " Select in CPU timeline" ) )
```
wolfpld, line 280: That's not correct. You need to check if the cpu usage data exists. Looking at the queries the info window makes is a good starting point.

wolfpld:
Unselecting doesn't work properly.
- Clicking in a window (not on the timeline) will unselect the thread.
- Unselecting only works if CPU data is drawn. It's enough to collapse the CPU data section to not be able to unselect the thread.

wolfpld:
The marker has incorrect offset.

image

wolfpld:
The wakeup line does not adapt to DPI.

image

wolfpld:
Is the data gathered correctly on Linux?

I have worker threads that are sleeping on a condition variable, and the job dispatching thread is waking the workers. But all I see is that the worker threads are "woken by kernel".

image

What I see instead are some strange wakeup attributions that don't make much sense.

image

Worker#16 is seemingly woken by Worker#12, but the worker threads do not communicate with each other. Workers may block each other on access to the job queue (as seen on the lock at the bottom of the image), but when Worker#16 obtains the lock, the wakeup reason is still "woken by kernel".

Lectem:
It seems that there are subtle differences between sched_wakeup and sched_waking (there's also sched_wakeup_new) and that the origin CPU might not be the one of the wakee depending on various things ...

See https://perfetto.dev/docs/data-sources/cpu-scheduling

I'll investigate a bit more.

Note that on windows, it is possible that a thread seemingly unrelated readies (wakes up) another if it triggers the scheduler by calling any kind of synchronization or syscall. (this mostly happens when thread was already ready to be scheduled). From what I saw on windows, we're now displaying the same things as WPA and Pix.

Lectem:
> The marker has incorrect offset.
> image

On my PC it I had a different issue, the marker was not covering the bottom line, while on your screenshot it's the top line. Could it be something related to inverted Y-axis on some renderers ? Anyway, I added +1 to the end of the line Y since `sstep = sty + 1`, could you confirm this works on your end ?

Lectem:
> It seems that there are subtle differences between sched_wakeup and sched_waking (there's also sched_wakeup_new) and that the origin CPU might not be the one of the wakee depending on various things ...
> 
> See https://perfetto.dev/docs/data-sources/cpu-scheduling
> 
> I'll investigate a bit more.
> 
> Note that on windows, it is possible that a thread seemingly unrelated readies (wakes up) another if it triggers the scheduler by calling any kind of synchronization or syscall. (this mostly happens when thread was already ready to be scheduled). From what I saw on windows, we're now displaying the same things as WPA and Pix.

In the end it seems that `sched_waking` is the event we needed, not `sched_wakeup`. 

`sched_wakeup` is only useful if you need to compute the scheduling latency of the context switch itself (ie: how long it takes for the thread to wakeup since the kernel actually decided to do it). Note that all `sched_wakeup` events are always followed by a `sched_switch` event, so it's not necessary to keep it as we do not compute the scheduling latency. 

Instead, I replaced `sched_wakeup` by `sched_waking`, which does give the information about the thread actually triggering the wakeup. This way now have the same information as the `ReadyThread` event on Windows. The only impact this has on current behavior is that the "Thread is waking up" part on the thread timeline will now indicate the same thing on both platforms, while it used to show only the scheduling latency on Linux.
image 

This was confirmed by analysing the ftrace output (for example here using `traceshark` (https://github.com/cunctator/traceshark).
image )

Lectem:
> > The marker has incorrect offset.
> > image
> 
> On my PC it I had a different issue, the marker was not covering the bottom line, while on your screenshot it's the top line. Could it be something related to inverted Y-axis on some renderers ? Anyway, I added +1 to the end of the line Y since `sstep = sty + 1`, could you confirm this works on your end ?

So this was actually a half pixel issue.

I have a setup with the following screens:

- 2560x1440(1x) (main) + 2560x1600(1.5x) + 1080p (1x)
  When using this setup, whatever screen the application opens up on first, I had the following result (same as you):
  image

- If I remove the first screen, and thus use the following setup:
   2560x1600(1.5x) (main) + 1080p (1x)
   image

- Then if I use only the 1080p (1x) screen, issue comes back.
   image.


# Using a half pixel offset:

- 2560x1440(1x) (main) + 2560x1600(1.5x) + 1080p (1x)
   image

-  2560x1600(1.5x) (main) + 1080p (1x)
  image

- 1080p only:
   image

This was also tested with different zoom levels and seems correct now.

wolfpld:
```
/home/wolf/tracy/profiler/src/profiler/TracyView_ContextSwitch.cpp:79:53: warning: expression result unused [-Wunused-value]
   79 |     case ContextSwitchData::Win32_WrPoolAllocation: "(Thread is waiting for a system allocation)";
      |                                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

server/TracyWorker.cpp
```diff
@@ -1419,7 +1419,8 @@ Worker::Worker( FileRead& f, EventType::Type eventMask, bool bgTasks, bool allow
 
     s_loadProgress.subTotal.store( 0, std::memory_order_relaxed );
     s_loadProgress.progress.store( LoadProgress::ContextSwitches, std::memory_order_relaxed );
-
+
```
wolfpld, line 1422: Spaces are added here.

Lectem: Fixed, weirdly neither Github nor my diff tool showed those

Lectem:
I think with this last fix there should not be any pending issues anymore (as far as I'm concerned).
I've been using this for a while now (both on Windows and Linux), and didn't encounter other issues (not saying there aren't, but I've not seen them if there are).

Do you need me to address anything else for this PR ?

wolfpld:
I'll try to take a look.

gglin001:
can we add support wakeup vis for `tracy-import-chrome` ?

PR #1022, opened 2025-04-10
Change TracyLine to be evaluate as a constant even with Edit-and-Cont…

GabrielSiliceum:
With this change Tracy line should work with Edit and continue

### DIFF
--- a/public/tracy/Tracy.hpp
+++ b/public/tracy/Tracy.hpp
@@ -13,7 +13,7 @@
 #endif
 
 #ifndef TracyLine
-#  define TracyLine __LINE__
+#  define TracyLine TracyConcat(__LINE__,U) // MSVC Edit and continue __LINE__ is non-constant. See https://developercommunity.visualstudio.com/t/-line-cannot-be-used-as-an-argument-for-constexpr/195665
 #endif
 
 #ifndef TRACY_ENABLE


wolfpld:
What are the implications of changing this? The error you see with e&c is valid.

GabrielSiliceum:
With this change, people using MSVC with Edit and Continue can at least compile and run their application.
If you try to run and change some Tracy zones, hot reload fails as it doesn't support modification of most static variables data ("[https://learn.microsoft.com/en-us/visualstudio/debugger/supported-code-changes-cpp?view=vs-2022"](https://learn.microsoft.com/en-us/visualstudio/debugger/supported-code-changes-cpp?view=vs-2022%22)) .

On my tests I get this error:

> error NEN1500 : New global or static variable 'PC_CORE::App::WorldLoop'::2'::__tracy_source_location84U' was added. 
(c:\users\gabriel\code\paraconquer\core\x64\debug\app.new.obj.enc)

So even if instrumented code can not be reloaded, it can at least compile and run without code modification.

Eventually, we have two options. 

The first one: ignore the issue and prevent people from  compiling with edit and continue.
The second: allow people to at least compile and run their app, with the knowledge that any line change of a TracyZone will prevent reloading.

PR #1024, opened 2025-04-14
tracy-update: Batch addr2line calls per 1024 addresses and escape paths

Lectem:
`addr2line` starts failing when given too many addresses to resolve. Batching per 1024 addresses allows not only to see progress but also fixes resolution for images with high number of addresses to resolve. 
Escaping is also required when image paths have spaces for example.

### DIFF
--- a/update/src/OfflineSymbolResolverAddr2Line.cpp
+++ b/update/src/OfflineSymbolResolverAddr2Line.cpp
@@ -44,57 +44,102 @@ public:
         }
     }
 
+    static void escapeShellParam(std::string const& s, std::string& out)
+    {
+        out.reserve( s.size() + 2 );
+        out.push_back( '"' );
+        for( unsigned char c : s )
+        {
+            if( ' ' <= c and c <= '~' and c != '\\' and c != '"' )
+            {
+                out.push_back( c );
+            }
+            else
+            {
+                out.push_back( '\\' );
+                switch( c )
+                {
+                    case '"':  out.push_back( '"' );  break;
+                    case '\\': out.push_back( '\\' ); break;
+                    case '\t': out.push_back( 't' );  break;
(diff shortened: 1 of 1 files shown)


wolfpld:
@tiago-rodrigues can you take a look?

tiago-rodrigues (commented):
Looks ok to me.

Lectem:
> Somewhat related, but I've actually moved to use LibDw in a local branch which is many times faster than the addr2line codepath, I'll try to do a PR for it later this week, I'm not 100% sure of the license implications of using libDw.

I'd be interested in having a look for https://github.com/wolfpld/tracy/pull/1009 if you ever want to integrate some symbol resolution mechanism using a library. I was thinking of maybe just modifying *libbacktrace* since it already supports it. From what I saw it wouldn't require that many changes, but it's obviously less robust than libdw (but supports more formats).

wolfpld:
> I'm not 100% sure of the license implications of using libDw.

lgpl is ok.

tiago-rodrigues:
> > Somewhat related, but I've actually moved to use LibDw in a local branch which is many times faster than the addr2line codepath, I'll try to do a PR for it later this week, I'm not 100% sure of the license implications of using libDw.
> 
> I'd be interested in having a look for #1009 if you ever want to integrate some symbol resolution mechanism using a library. I was thinking of maybe just modifying _libbacktrace_ since it already supports it. From what I saw it wouldn't require that many changes, but it's obviously less robust than libdw (but supports more formats).

I'll tag you in the PR when I get to it later this week.
I actually tried a few different resolution backends (Liberty - has licensing issues, libdw, libbacktrace - the version tracy uses now, but couldn't get it to work properly for offline symbol resolution, needed to load the .so in mem which can fail due to deps).

PR #1026, opened 2025-04-14
Offline symbol resolving improvements (parallel image resolving, new fast libdw based resolving backend)

tiago-rodrigues:
- Add parallel implementation of offline symbol resolving (process images in parallel, use '-m' to specify the level of parallelism, -1 [default] is max core count)
- Add a couple of more offline symbol resolving backend implementations for linux (use '-n' to choose the backend to use): 
   * a libdw based backend that is ~100x faster than addr2line and LGPL licensed (requires libdw package to be installed and LIBDW_OFFLINE_SYMBOL_RESOLUTION_SUPPORT to be enabled)
   * a libacktrace based backend (using the libbacktrace version that exists in the tracy codebase). This backend is not fully functional as it need to load the .so into memory to resolve the symbols which is not always possible (enable with LIBBACKTRACE_OFFLINE_SYMBOL_RESOLUTION_SUPPORT)
- Add option to exclude certain images from being resolved (specified via -x)

### DIFF
--- a/update/src/OfflineSymbolResolver.cpp
+++ b/update/src/OfflineSymbolResolver.cpp
@@ -207,8 +207,13 @@ bool PatchSymbolsWithRegex( tracy::Worker& worker, const PathSubstitutionList& p
         totalEntries += entries.size();
     }
 
+#ifdef _WIN32
+    // DbgHelper is not thread safe
+    unsigned int maxConcurrent = 1;
+#else
     unsigned int maxConcurrent = 
         (options.maxParallelism <= 0) ? std::thread::hardware_concurrency() : int(options.maxParallelism);
+#endif
 
     std::cout << "* Running " << resolvedResults.size() << " resolution jobs in parallel (batches of " << maxConcurrent << ")" << std::endl;    
     Stopwatch parallelResolveStopwatch;
--- a/update/src/OfflineSymbolResolver.cpp
+++ b/update/src/OfflineSymbolResolver.cpp
@@ -3,6 +3,7 @@
 #include <vector>
 #include <future>
 #include <thread>
+#include <chrono>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
--- a/update/src/OfflineSymbolResolver.cpp
+++ b/update/src/OfflineSymbolResolver.cpp
@@ -14,20 +14,54 @@
 
 #include "OfflineSymbolResolver.h"
 
+struct OfflineResolverEntry
+{
+    const char* name;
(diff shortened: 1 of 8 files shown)


tiago-rodrigues:
@Lectem : thjis is the PR I was menioning here: https://github.com/wolfpld/tracy/pull/1024#issuecomment-2801954325

@wolfpld : what would be the right way to add the dependency to "libdw"? As it's significantly faster than addr2line, I would like to make it the default.

I left the backtrace implementation for reference as @Lectem mentioned he wanted to attempt implementing it, I however never succeeded in getting it to work for all .so's maybe there is a way. Still I think libdw is the best option at the cost of this extra dependency.

update/src/OfflineSymbolResolver.cpp
```diff
@@ -24,19 +43,107 @@ bool ApplyPathSubstitutions( std::string& path, const PathSubstitutionList& path
...
+            auto future = std::async( std::launch::async, std::bind( func, index ) ); 
+            results.push_back( { index, std::move(future) } );
+            index++;
+        }
+
+        // Remove completed jobs
+        results.erase( std::remove_if( results.begin(), results.end(),
+            [onJobEnd]( JobEntry& entry )
+            { 
+                const bool finished = entry.future.wait_for( std::chrono::milliseconds( 100 ) ) == std::future_status::ready;
```
Lectem, line 106: Any reason to use wait_for with a timeout of 100ms here ?

Lectem:
Interesting, so from what I saw, we could use the `offline_callbacks` to check that the build-id is correct ? One of the main issues I'd like to fix is the that we might not be using the correct debug information. On Windows this can be done by providing the PDB GUID, and on linux using the build-id.
If we do integrate libDw and that it supports build-id checks and we can do file lookup with debuginfod, then there'd be little need for modifying libbacktrace. (Except that libbacktrace supports more formats and could be used on windows too)

wolfpld:
> If we do integrate libDw and that it supports build-id checks and we can do file lookup with debuginfod, then there'd be less/little need for modifying libbacktrace. (Except that libbacktrace supports more formats, could be used on platforms other than Linux, and doesn't require installing a package).

There are changes in Tracy libbacktrace that are required for inline functions to be properly reported.

tiago-rodrigues:
I've made the new back-ends optional so it passes on CI and still allows users to enable them if they want. Although I would like the libdw backend to the the default, unfortunately I don't see how I would do that with CPM and pulling and building elfutils to do this.

PR #1028, opened 2025-04-16
Remove wp_viewport_set_source call in Wayland backend

### LINKED ISSUE #905: wp_viewport@33: error 2: source rectangle out of buffer bounds
A similar issue as in SDL, please check https://github.com/libsdl-org/SDL/pull/9285

Supreeeme:
There seems to be some sort of race condition between setting the egl buffer size and the viewport size, causing the viewport source size to be larger than the current buffer size, and leading to a crash. It turns out setting the viewport source size is unnecessary anyway, so we can just avoid it.
Fixes #905

### DIFF
--- a/profiler/src/BackendWayland.cpp
+++ b/profiler/src/BackendWayland.cpp
@@ -1139,7 +1139,6 @@ void Backend::NewFrame( int& w, int& h )
         wl_egl_window_resize( s_eglWin, s_width * s_maxScale / 120, s_height * s_maxScale / 120, 0, 0 );
         if( s_fracSurf )
         {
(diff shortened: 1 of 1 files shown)


wolfpld:
> It turns out setting the viewport source size is unnecessary anyway, so we can just avoid it.

That's just wrong.

Supreeeme:
Care to explain? Does `wp_viewport_set_source` not imply cropping? Does tracy need to crop the surface? Because it certainly works on my end, and it definitely says the source can be unset in the protocol spec.

wolfpld:
https://github.com/user-attachments/assets/e2a9f07f-2ca1-4a2f-b573-432467ee22ef

wolfpld:
How does all this relate to the issue the video shows?

Supreeeme:
I have no idea what issue that video was even supposed to show, there was no context and nothing appears obviously wrong to me.

Issue #1034, opened 2025-04-24
Make Tracy dpi aware per monitor on windows.


simonvanbernem:
On Windows 10, I have a 1440p monitor at 100% scaling and a 4k monitor at 150% scaling. I often move Tracy windows between the screen to compare runs. Every time I do this, I have to manually change the GUI scaling in tracy to make things appropriately sized again, since tracy does not automatically adjust its GUI scaling based on monitor DPI.

This can be achieved by passing DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2  to [SetThreadDpiAwarenessContext](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setthreaddpiawarenesscontext) and handling the WM_DPICHANGED window message. Depending on the backend it should be fairly easy to implement. I could take a stab at implementing this myself and submitting a pull request, if wanted.

wolfpld:
Please do.

PR #1037, opened 2025-04-25
Adding per-monitor DPI awareness

simonvanbernem:
This pull requests adds per monitor dpi awareness, as requested in #1034 . Specifically, if you have two monitors with different scaling set in the Windows display settings, this change will make it so the window and its contents are automatically scaled correctly according to the OS setting, when moved between monitors.

Luckily, glfw supports this directly and the changes are minimal:
- Instead of glfwGetMonitorContentScale, glfwGetWindowContentScale is used, which updates when the window is moved to a different window with different display scale setting in windows.
- By passing GLFW_SCALE_TO_MONITOR to glfwWindowHint, glfw will automatically resize the window as well. Without this the contents would be enlarged when moved to a monitor with higher scaling, but the window would stay the same size, so the UI would look more cramped.

I have only tested this change on Windows 10, where everything worked as intended.


wolfpld:
Works on win11 too.

PR #1038, opened 2025-04-27
Added saving of user ui scale via global configuration.

### LINKED ISSUE #1035: Have tracy remember display scaling
Currently, when opening tracy on Windows 10, it always starts at 100% scaling. Ideally, I'd like tracy to remember the last scaling it was set at, so that I don't need to reconfigure it in a tiny UI on a 4k screen every time I start a new instance.

This would also be helpful, because I believe right now tracy does remember window positions, which leads to strange behavior: If I had tracy on my 4k screen, and open it on the 1440p, the connect window inside the main tracy window might be completely to the lower right, almost off screen, because that was an appropriate position on the big 4k screen, but on a smaller screen, it is almost off screen. If tracy also respected per-monitor dpi scaling, this would not happen anymore.

simonvanbernem:
This pull requests resolves #1035. 

- Moved userScale into s_config.
- Added "ui" config section for userScale and saveUserScale.
- Prevented "Get started" and "Loading trace..." windows from having their position saved between restarts, because if you change scale and then restart, they might get pushed to the edge of the window, which is confusing. And since saved window positions just don't respect ui scale in general, it also was pretty confusing before, with different edge cases imo.

As a result, changing the UI scale takes longer, because the config file is now saved when it happens.

### DIFF
--- a/profiler/src/main.cpp
+++ b/profiler/src/main.cpp
@@ -101,7 +101,6 @@ static std::atomic<ViewShutdown> viewShutdown { ViewShutdown::False };
 static double animTime = 0;
 static float dpiScale = -1.f;
 static bool dpiScaleOverriddenFromEnv = false;
-static float userScale = 1.f;
 static float prevScale = 1.f;
 static int dpiChanged = 0;
 static bool dpiFirstSetup = true;
@@ -162,7 +161,7 @@ static void ScaleWindow(ImGuiWindow* window, float scale)
 
 static void SetupDPIScale()
 {
-    auto scale = dpiScale * userScale;
+    auto scale = dpiScale * s_config.userScale;
 
     if( !dpiFirstSetup && prevScale == scale ) return;
     dpiFirstSetup = false;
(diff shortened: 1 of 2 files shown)


wolfpld:
> As a result, changing the UI scale takes longer, because the config file is now saved (if the save UI Scale ist activated) when it happens.

Maybe this could be moved to when the program is exiting?

simonvanbernem:
I actually expected that the config would be saved on exit in general, but when I looked it was eplicit in existing code. So I wasn't sure if that change would be welcomed or not. I'll add it to the pull request.

simonvanbernem:
I added an unconditional saving of the config on exit.

However, I have questions: In the general settings section and Achievements popup, SaveConfig is called. Should I then remove those as well, as they aren't strictly necessary? On the other hand, if a second instance is started, the changes to config would not be visible until the instance that made the changes is closed, which might be confusing. Additionally, I noticed that the delay I saw wasn't due to saving the config (your mileage may vary with a different drive), but apparently rescaling the UI just takes that long in debug mode.

So, for me, this change didn't make anything faster, but introduced an inconsistent style to the codebase (instant save vs at shutdown). Doesn't seem worth it?

simonvanbernem:
Ok, I reverted.

PR #1040, opened 2025-05-06
Time to string exact

blegat:
On my system, I am getting a segmentation fault. The issue is that all these are evaluated to `false`
https://github.com/wolfpld/tracy/blob/a03c7580b9101df3225937524e09a9a885d5464a/server/TracyPrint.cpp#L262
https://github.com/wolfpld/tracy/blob/a03c7580b9101df3225937524e09a9a885d5464a/server/TracyPrint.cpp#L285
https://github.com/wolfpld/tracy/blob/a03c7580b9101df3225937524e09a9a885d5464a/server/TracyPrint.cpp#L297
https://github.com/wolfpld/tracy/blob/a03c7580b9101df3225937524e09a9a885d5464a/server/TracyPrint.cpp#L306
and then the errors occurs in
https://github.com/wolfpld/tracy/blob/a03c7580b9101df3225937524e09a9a885d5464a/server/TracyPrint.cpp#L72-L91
Here is a minimal reproducer
```c++
#include <stdio.h>
#include <stdint.h>
#include <iostream>
#include <limits>

void ko(int64_t _ns)
{
    std::cout << "KO" << std::endl;
    std::cout << _ns << std::endl;
    uint64_t ns;
    if( _ns < 0 )
    {
        ns = -_ns;
    }
    else
    {
        ns = _ns;
    }
    std::cout << ns << std::endl;
    std::cout << 1000ll << std::endl;
    std::cout << (ns >= 1000ll) << std::endl;
}

void ok(int64_t _ns)
{
    std::cout << "OK" << std::endl;
    std::cout << _ns << std::endl;
    uint64_t ns;
    if( _ns == std::numeric_limits<int64_t>::min() )
    {
        ns = -(_ns + 1) + 1;
    }
    else if( _ns < 0 )
    {
        ns = -_ns;
    }
    else
    {
        ns = _ns;
    }
    std::cout << ns << std::endl;
    std::cout << 1000ll << std::endl;
    std::cout << (ns >= 1000ll) << std::endl;
}

int main() {
    ko(-9223372036854775808ULL);
    ok(-9223372036854775808ULL);
    return 0;
}
```
Then I can reproduce the issue with `g++` with `-O2` or `-O3` but not `clang++`
```
$ g++ -O3 main.cpp && ./a.out
KO
-9223372036854775808
9223372036854775808
1000
0
OK
-9223372036854775808
9223372036854775808
1000
1
$ g++ -O1 main.cpp && ./a.out
KO
-9223372036854775808
9223372036854775808
1000
1
OK
-9223372036854775808
9223372036854775808
1000
1
$ clang++ -O3 main.cpp && ./a.out
KO
-9223372036854775808
9223372036854775808
1000
1
OK
-9223372036854775808
9223372036854775808
1000
1
```
The way I understand the issue is that `-_ns` does not fit in `int64_t` even though it fits in `uint64_t`.
So the idea is to first convert to `uint64_t` and then do `+1`.
My gcc version is
```
$ g++ --version
g++ (GCC) 15.1.1 20250425
$ uname -a
Linux Precision5570 6.14.4-arch1-2 #1 SMP PREEMPT_DYNAMIC Tue, 29 Apr 2025 09:23:13 +0000 x86_64 GNU/Linux
```

### DIFF
--- a/server/TracyPrint.cpp
+++ b/server/TracyPrint.cpp
@@ -246,7 +246,11 @@ const char* TimeToStringExact( int64_t _ns )
     bufsel = ( bufsel + 1 ) % Pool;
 
     uint64_t ns;
-    if( _ns < 0 )
+    if( _ns == std::numeric_limits<int64_t>::min() )
+    {
+        ns = -(_ns + 1) + 1;
+    }
+    else if( _ns < 0 )
     {
         *buf = '-';
         buf++;
--- a/server/TracyPrint.cpp
+++ b/server/TracyPrint.cpp
@@ -71,7 +71,7 @@ static inline void PrintSmallInt( char*& buf, uint64_t v )
 
 static inline void PrintSmallInt0( char*& buf, uint64_t v )
 {
-    assert( v < uint64_t(1000ll) );
+    assert( v < 1000 );
     if( v >= 100 )
     {
         memcpy( buf, IntTable100 + v/10*2, 2 );
@@ -157,13 +157,13 @@ static inline void PrintSecondsFrac( char*& buf, uint64_t v )
 
 uint64_t _int64_abs( int64_t x )
 {
-    if( x == std::numeric_limits<int64_t>::min() )
+    if( x < 0 )
     {
-        return -(x + 1) + 1;
-    }
-    else if( x < 0 )
-    {
-        return -x;
+        // `-x` does not work if `x` is `std::numeric_limits<int64_t>::min()`,
+        // see https://github.com/wolfpld/tracy/pull/1040
+        // This works though:
+        // https://graphics.stanford.edu/~seander/bithacks.html#IntegerAbs
+        return -(uint64_t)x;
     }
     else
     {
--- a/server/TracyPrint.cpp
+++ b/server/TracyPrint.cpp
@@ -71,7 +71,7 @@ static inline void PrintSmallInt( char*& buf, uint64_t v )
 
 static inline void PrintSmallInt0( char*& buf, uint64_t v )
 {
-    assert( v < 1000 );
+    assert( v < uint64_t(1000ll) );
     if( v >= 100 )
     {
         memcpy( buf, IntTable100 + v/10*2, 2 );
@@ -155,6 +155,22 @@ static inline void PrintSecondsFrac( char*& buf, uint64_t v )
     }
 }
 
+uint64_t _int64_abs( int64_t x )
+{
+    if( x == std::numeric_limits<int64_t>::min() )
+    {
+        return -(x + 1) + 1;
+    }
+    else if( x < 0 )
+    {
+        return -x;
+    }
+    else
+    {
+        return x;
+    }
+}
+
 const char* TimeToString( int64_t _ns )
 {
     enum { Pool = 8 };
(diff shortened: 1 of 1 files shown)


server/TracyPrint.cpp
```diff
@@ -155,6 +155,22 @@ static inline void PrintSecondsFrac( char*& buf, uint64_t v )
     }
 }
 
+uint64_t _int64_abs( int64_t x )
+{
+    if( x == std::numeric_limits<int64_t>::min() )
+    {
+        return -(x + 1) + 1;
```
DanielGibson, line 162: Shouldn't this be casted to uint64_t before adding 1?
Example with int8_t so it's readable:
```
int8_t x  == -128
=> (x+1)  == -127
=> -(x+1) == 127 // ok, fits into int8_t
=> -(x+1) + 1 == ??? // still a signed int8_t, right?
                    // as 128 doesn't fit into int8_t, it probably wraps around to -128

=> should be `uint64_t(-(x + 1)) + 1;`
```

However, https://graphics.stanford.edu/~seander/bithacks.html#IntegerAbs claims that that for **every _negative_ x** the following works (so you don't need the INT64_MIN special case):
```c++
return -(uint64_t)x;
```
They write (using `unsigned` and `int` instead of (u)int64_t):
> But citing the ISO C99 spec on July 9, 2008, Vincent Lefèvre convinced me to remove it becasue even on non-2s-complement machines -(unsigned)v will do the right thing.
> The evaluation of `-(unsigned)v` first converts the negative value of `v` to an unsigned by adding `2**N`, yielding a 2s complement representation of v's value that I'll call U. Then, U is negated, giving the desired result,
> `-U = 0 - U = 2**N - U = 2**N - (v+2**N) = -v = abs(v)`.

DanielGibson:
By the way, do you have any idea why the time is INT64_MIN?
I ran into the same problem (see #1048) but can't properly debug it because for some reason Debug builds of tracy-profiler don't link on my system..

Your fix here works, but this is such a weird value that there may be is another underlying bug.

blegat:
Not sure what the source is but it may be
https://github.com/wolfpld/tracy/blob/c530219e46120ed26efcb539a45098b0dacd20c2/profiler/src/profiler/TracyView.hpp#L727

server/TracyPrint.cpp
blegat: Thanks, it indeed works

DanielGibson:
I just remembered that I don't need a Debug build to debug, I can just put `#pragma GCC optimize("O0")` at the top of the sourcefiles I want to inspect stuff in.
Doing that in TracyPrint.cpp makes the crash go away, but TracyView_FrameTimeline.cpp (for `View::DrawTimelineFramesHeader()`) is more interesting anyway, because there the int64_t that's passed to `PrintSmallInt0()` is calculated.

I turned that `TooltipIfHovered(...)` line into three so I can inspect the involved values:
```c++
ImVec2 mousepos = ImGui::GetIO().MousePos;
int64_t ns =  m_vd.zvStart + ( mousepos.x - wpos.x ) * nspx ;
TooltipIfHovered( TimeToStringExact(ns) );
```
So `ns` is indeed `-9223372036854775808` (-2^63 - INT64_MIN).
Most other values are normal: `m_vd.zvStart = 1761226325805`, `wpos.x = 5`, `nspx = 29962.384`.
This leaves `mousepos` - which for some reason is `{x = -3.40282347e+38, y = -3.40282347e+38}` ?!

`-3.40282347e+38` is `- FLT_MAX`, which must be from `BackendWayland.cpp`:
```c++
static void PointerLeave( void*, struct wl_pointer* pointer, uint32_t serial, struct wl_surface* surf )
{
    ImGuiIO& io = ImGui::GetIO();
    io.AddMousePosEvent( -FLT_MAX, -FLT_MAX );
}
```

I don't know what that is supposed to achieve exactly or what `TimeToStringExact()` (or any other function) is supposed to do with that value; probably @wolfpld should look into this.

wolfpld:
As a commentary I will just link this interesting paper: https://web.ist.utl.pt/nuno.lopes/pubs/ub-pldi25.pdf

PR #1050, opened 2025-05-23
Add option to toggle installation

rachit7645:
This PR adds an option `TRACY_INSTALL` to enable/disable installation. I've set it to `ON` by default to match older behavior.


wolfpld:
What does this do? Are you looking for https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html?

rachit7645:
I've tried that, but it didn't work for me, so I added an explicit option to tracy instead

Issue #631, opened 2023-10-03
Missing vertical scrollbar on macOS


slomp:
It appears that the vertical scrollbar of the main window is missing on macOS; have a look at the attached screenshot (a capture of the right side of the profiler window):
<img width="240" alt="Screenshot 2023-10-03 at 7 49 53 AM" src="https://github.com/wolfpld/tracy/assets/526886/ab79931b-b983-4697-a00a-c8306f8bac08">

wolfpld:
ImGui bug.

lysannschlegel:
Could you link the related imgui issue?

IanButterworth:
Does anyone know if this is fixed on 0.10.0? I would test it but I've not got any compatible data to load and our client is on 0.9.1

Without the vertical scrollbar Tracy is almost unusable on a MacOS macbook with a trackpad. 
In addition to the lack of vertical scroll, the vertical drag zoom is HYPER sensitive, so much so that you cannot really drag sideways while maintaining the vertical region of interest.

IanButterworth:
It was pointed out to me that https://tracy.nereid.pl/Tracy-release.data is a tracy file with a different extension.

Loading that I don't see the vertical scrollbar on v0.10.0

![Screenshot 2024-07-14 at 7 33 08 PM](https://github.com/user-attachments/assets/2a07a2ac-0233-4705-9410-cedcf9e4fbde)

Is anyone aware of a way to scroll vertically without it? I thought a key combo with a 2 finger drag would work but no luck yet.

wolfpld:
> Does anyone know if this is fixed on 0.10.0?

It's not fixed on master.

> In addition to the lack of vertical scroll, the vertical drag zoom is HYPER sensitive, so much so that you cannot really drag sideways while maintaining the vertical region of interest.

Can you check if #815 makes this better for you?

undefz:
Hey @wolfpld,
I was experiencing the same issue with the missing scrollbar on Mac and tried applying the patch from #815.
I can confirm that it makes Tracy usable in this scenario, as it enables vertical scrolling using a two-finger press on the touchpad.
Any chances of merging that pull request to main? 🙏

wolfpld:
I have merged it.

wolfpld:
The scrollbar is now visible as intended.

PR #1055, opened 2025-05-31
fix vertical scroll bar lost  on macos

gglin001:
trying to fix 
- https://github.com/wolfpld/tracy/issues/631

on macos 15.5 with 4k monitor and 1920x1080 resolution

before

Screenshot 2025-05-31 at 23 41 52

after this pr
Screenshot 2025-05-31 at 23 42 24


profiler/src/profiler/TracyView.cpp
```diff
@@ -808,6 +808,7 @@ bool View::DrawImpl()
     sprintf( tmp, "%s###Profiler", m_worker.GetCaptureName().c_str() );
     ImGui::SetNextWindowSize( ImVec2( 1550, 800 ), ImGuiCond_FirstUseEver );
     ImGui::Begin( tmp, keepOpenPtr, ImGuiWindowFlags_NoScrollbar | ImGuiWindowFlags_NoBringToFrontOnFocus );
+    ImGuiViewport* viewport = ImGui::GetMainViewport();
```
gglin001, line 811: if define `TRACY_NO_ROOT_WINDOW`

`viewport` will be used but not defined

PR #1056, opened 2025-05-31
GCC and Clang compatible on Windows

FilipNur:
Made simple modifications so that project can be compiled wiith Clang and GCC on Windows


public/client/TracyProfiler.cpp
```diff
@@ -3122,14 +3122,20 @@ char* Profiler::SafeCopyProlog( const char* data, size_t size )
...
-    {
-        memcpy( buf, data, size );
-    }
-    __except( 1 /*EXCEPTION_EXECUTE_HANDLER*/ )
-    {
-        success = false;
-    }
+
+	#if defined(__clang__) || defined(__GNUC__)
+		// memory size pointed to by buf variable is checked above
```
wolfpld, line 3127: Where is this check?

FilipNur: From what I see, on line 3122 in ```public/client/TracyProfiler.cpp```, there is ```if``` statement that checks if ```buf``` pointer will be able to contain new data and if not, then memory is allocated with size equal to size of data that will be copied.

FilipNur:
@wolfpld, I finished changing source so it compiles with gcc and clang on Windows. I tested it on both compilers, test program that uses Tracy, and Tracy profiler itself compile successfully. The initial problem with safe ```memcpy``` is solved using pipes for data transfer similar to linux implementation. I wasn't able to find fix for too many sections that I mentioned before and just disabled ```CMAKE_INTERPROCEDURAL_OPTIMIZATION``` when compiling with gcc.

cmake/config.cmake
```diff
@@ -30,15 +34,20 @@ endif()
...
 endif()
 
 if(EMSCRIPTEN)
     add_compile_options(-pthread -DIMGUI_IMPL_OPENGL_ES2)
 endif()
 
 if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT EMSCRIPTEN)
-    set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
+	# Mingw gcc on windows can't handle section count resulting during compilation of profiler/src/profiler/TracyMicroArchitecture.cpp
+	if(NOT (MINGW OR "${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU") AND WIN32 AND NOT ("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang"))
```
wolfpld, line 48: Please move `WIN32` as first condition for better readability.

import/src/json.hpp
```diff
@@ -16799,7 +16799,7 @@ class binary_writer
 
     void write_compact_float(const number_float_t n, detail::input_format_t format)
     {
-#ifdef __GNUC__
+#if defined __GNUC__ || defined __clang__
```
wolfpld, line 16802: Why is this needed?

```sh
% clang -dM -E -x c /dev/null | grep __GNUC__
#define __GNUC__ 4
```

profiler/src/profiler/TracyView_Timeline.cpp
```diff
@@ -10,6 +10,14 @@
 #include "TracyTimelineItemThread.hpp"
 #include "TracyView.hpp"
 
+#if defined _WIN32 && (defined __GNUC__ && !defined __clang__)
+//_WIN32 - only tested on windows
+
+// gcc throws error for not present std::pow function,
+// clang does not have that problem
+#include <cmath>
+#endif
```
wolfpld, line 19: The correct solution would be to change `std::pow` to `pow` and include `math.h` (with no platform checks).

profiler/src/stb_image.h
```diff
@@ -1,4 +1,4 @@
-/* stb_image - v2.29 - public domain image loader - http://nothings.org/stb
+/* stb_image - v2.28 - public domain image loader - http://nothings.org/stb
```
wolfpld, line 1: ?

FilipNur: I actually don't know. I did repulling, maybe something there, will fix it.

import/src/import-chrome.cpp
```diff
@@ -11,9 +11,11 @@
 #include <unordered_map>
 #include <zstd.h>
 
-#ifdef _MSC_VER
+#if defined _MSC_VER || defined __clang__ || defined __GNUC__
+// all checked compilers contain _stat64
```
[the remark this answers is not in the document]
FilipNur: Will fix it.

import/src/json.hpp
FilipNur: Just being explicit. I can change it if needed.

profiler/src/main.cpp
```diff
@@ -19,6 +19,14 @@
...
 #endif
 
+#if defined _WIN32 && (defined __GNUC__ && !defined __clang__)
+//_WIN32 - only tested on windows
+
+// gcc throws error for not present std::pow function,
+// clang does not have that problem
+#include <cmath>
+#endif
+
```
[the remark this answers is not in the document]
FilipNur: Tested, and turns out that using cmake from clang toolchain and gcc compiler created errors. So not needed, will remove.

profiler/src/stb_image.h
FilipNur: Now i know, in the current repo there are 2 ```stb_image.h``` files, ```test/stb_image.h``` has ```2.28``` and ```profiler/src/stb_image.h``` has ```2.29```. When applying changes I just copied one file in place of the other to not do the same changes again. Now I checked that they are different. Will fix.

import/src/json.hpp
FilipNur: I understand that this would be due to having unified ```#if``` conditions. I removed obviously not needed checks for ```__clang__```, only left ones that coexist with version checking of ```GNU```, as I don't know if there can be any differences, just to be sure.

FilipNur: Ok, does that mean that if ```__GNUC__``` version is checked then for sure ```clang``` will not be supported for this block of code, or it still can be checked via ```#if (<GNUC version check>) || defined __clang__```? Because if only version check done in this way is not valid then it still should be possible to check if ```clang``` is compiler that is used.

FilipNur:
@wolfpld Fixed things you pointed out and some other minor things.

import/src/json.hpp
wolfpld: clang defines a very old version of gcc version macro, which may affect what features are available. This can be problematic in rare circumstances.

FilipNur:
@mcourteaux I know and I don't know why. I will push changes when I resolve this issue.

mcourteaux:
As I can't assess what your PR does due to this issue, make sure it's still relevant after my last PR was merged.

FilipNur:
@mcourteaux I see that your pull request is also about ```mingw``` compatibility so my pull request might be not needed, as I was trying to achieve something similar. But I have one question: in your changes to ```TracyProfiler.cpp``` you did
```
#ifdef _WIN32
+ #  ifdef _MSC_VER
      __try
      {
          memcpy( buf, data, size );
      {
          success = false;
      }
+ #  else
+     memcpy( buf, data, size );
+ #  endif
    #else
    ...
```
Where the ```memcpy```, that is executed if ```_MSC_VER``` is not defined, is not protected against potential ```segfault```. I am just curious why was it merged, because when I proposed just bare ```memcpy``` in first request, @wolfpld sent me to #889, whetre issue says about ensuring safe ```memcpy```.

wolfpld:
It's an incremental step.

FilipNur:
Fixed line endings in ```public/client/TracyProfiler.cpp``` and rebased project to latest master branch.

update/src/OfflineSymbolResolverAddr2Line.cpp
```diff
@@ -89,7 +89,7 @@ class SymbolResolver
             const size_t startIdx = entryIdx;
             const size_t batchEndIdx = std::min( inputEntryList.size(), startIdx + (size_t)1024 );
 
-            printf( "Resolving symbols [%zu-%zu[\n", startIdx, batchEndIdx );
+            printf( "Resolving symbols [%zu-%zu]\n", startIdx, batchEndIdx );
```
wolfpld, line 92: Why are there changes that are already applied on master?

FilipNur: From what I saw the difference was ```space``` vs ```tab```.

FilipNur:
@wolfpld Could you check it?

wolfpld:
You are still changing way too much code for no reason, especially in external libraries, and there are comments you have not addressed properly. What's there to check?

PR #1065, opened 2025-06-07
Added support for host query reset when collecting Vulkan traces

### LINKED ISSUE #1053: `vkCmdResetQueryPool` called in `TracyVkCollect` despite using `VK_EXT_host_query_reset`
`vkCmdResetQueryPool` is called instead of `vkResetQueryPool` in `TracyVkCollect` even when the context was created with `TracyVkContextHostCalibrated`:
https://github.com/wolfpld/tracy/blob/f14ff2b994dac1de1d99acfcdced4e9dced47d97/public/tracy/TracyVulkan.hpp#L328

I believe forcing the "Cmd" variant is not needed here?

Also, this would allow removing the cmdbuf parameter of `TracyVkCollect` when host query reset are used, making it more flexible in when it is called.

maxime-modulopi:
Closes #1053

The manual should probably be updated but I'm not sure how to build it so...

### DIFF
--- a/public/tracy/TracyVulkan.hpp
+++ b/public/tracy/TracyVulkan.hpp
@@ -16,6 +16,7 @@
 #define TracyVkZoneC(c,x,y,z)
 #define TracyVkZoneTransient(c,x,y,z,w)
 #define TracyVkCollect(c,x)
+#define TracyVkCollectHost(c)
 
 #define TracyVkNamedZoneS(c,x,y,z,w,a)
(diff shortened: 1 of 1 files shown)


wolfpld:
> The manual should probably be updated but I'm not sure how to build it so...

Yes, do update. It's a text file and it should be fairly easy to figure out what the expected syntax is, even without building it (CI will verify it anyways).

maxime-modulopi:
Done

PR #1082, opened 2025-06-21
Fix vendor.cmake

gav2xlin:
Your latest releases cannot be used (without hacks and patches) with vcpkg and Homebrew because they prohibit external downloads that CPM.cmake does

pkg_check_modules(CAPSTONE **capstone**)
pkg_check_modules(GLFW **glfw3**)
pkg_check_modules(FREETYPE **freetype2**)
...
[https://github.com/Homebrew/homebrew-core/blob/badf764aab60b495bb699e23ab3b1633f588df50/Formula/t/tracy.rb](https://github.com/Homebrew/homebrew-core/blob/badf764aab60b495bb699e23ab3b1633f588df50/Formula/t/tracy.rb)
depends_on "**glfw**"
depends_on "**capstone**"
depends_on "**freetype**"

[https://github.com/microsoft/vcpkg/blob/master/ports/tracy/vcpkg.json](https://github.com/microsoft/vcpkg/blob/master/ports/tracy/vcpkg.json)
"dependencies": [
        {
          "name": "**capstone**",
          "features": [
            "arm",
            "arm64",
            "x86"
          ]
        },
        ...
        "**freetype**",
        "**glfw3**",
        ...
      ]

Please, use the same approach and check that modules are already installed:

pkg_check_modules(ZSTD zstd)
pkg_check_modules(ImGui imgui)
pkg_check_modules(NFD nativefiledialog-extended)
pkg_check_modules(PPQSORT ppqsort)

(I'm checking that these names and CMake variables are correct)

### DIFF
--- a/cmake/server.cmake
+++ b/cmake/server.cmake
@@ -29,7 +29,7 @@ list(TRANSFORM TRACY_SERVER_SOURCES PREPEND "${TRACY_SERVER_DIR}/")
 
 add_library(TracyServer STATIC EXCLUDE_FROM_ALL ${TRACY_COMMON_SOURCES} ${TRACY_SERVER_SOURCES})
 target_include_directories(TracyServer PUBLIC ${TRACY_COMMON_DIR} ${TRACY_SERVER_DIR})
-target_link_libraries(TracyServer PUBLIC TracyCapstone libzstd PPQSort::PPQSort)
+target_link_libraries(TracyServer PUBLIC TracyCapstone TracyZstd TracyPPQSort)
 if(NO_STATISTICS)
     target_compile_definitions(TracyServer PUBLIC TRACY_NO_STATISTICS)
 endif()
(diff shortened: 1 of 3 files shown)


wolfpld:
I don't understand how you expect this to work, when there's a need for specific patches to be applied (which you are just ignoring), or when the library explicitly forbids what you want to do (as explained in #1079).

The amount of libraries retrieved via CPM will only increase, and downstream needs to adapt to this.

See https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=moderncore if you need guidance.

dg0yt:
> The amount of libraries retrieved via CPM will only increase, and downstream needs to adapt to this.

Will this increase be limited to what is used by the executables, or will it also affect the libraries?

wolfpld:
I don't expect the libraries to be affected. The intention is that you can just dump the Tracy client sources into your application and have everything working with minimal effort required.

PR #1093, opened 2025-07-07
Add tooltip to Limit Range checkboxes: Ctrl+drag

MrAntoine77:
This PR adds a tooltip to the “Limit range” checkboxes in the Find Zone panel.
When the user hovers over the checkbox, the tooltip displays the message:

image

This clarifies an alternative way to select the zone range, improving discoverability and user experience.


wolfpld:
Also, coming soon:

image

(It says "holding a key" without specifying the ctrl key due to issues with latex to markdown conversion process, which omits the key descriptions specified with the menukeys package.)

PR #1096, opened 2025-07-07
Add lock struct with condition variable to main thread

MrAntoine77:
> The links you have given do not work.

**Here are the working links, I hope this will make it clearer.**

In certain situations (notably during Tracy initialization), we observed delays caused by a form of starvation of the main thread.
The issue occurs when the worker thread quickly reacquires the mutex after releasing it — before the main thread has had a chance to wake up and acquire it itself.

This leads to a situation where the main thread is repeatedly preempted, causing noticeable lag or even temporary freezes.
Solution

To address this, we introduce a lock coordination mechanism using:
- An atomic flag: `mainThreadWantsLock`
- A condition_variable to wake the worker after the main thread is done

The `MainThreadDataLockGuard`:
- Signals that the main thread wants the lock
- Acquires the lock
- Upon destruction, clears the flag and notifies the worker thread

On the worker side:
- Before continuing processing, it checks the flag
- If the main thread is requesting the lock, it calls wait_for on the condition variable (1ms timeout), giving the main thread a chance to take the lock first

This ensures the main thread gets temporary priority when needed, reducing contention and preventing starvation.

We can see the problem by self profiling :
Before:
Capture d'écran 2025-07-21 093332
After:
Capture d'écran 2025-07-21 093824

This significantly improves responsiveness, especially during initialization, by ensuring the main thread is not indefinitely delayed by worker lock contention.

Before:

https://github.com/user-attachments/assets/fcaa9555-1bd5-45e0-9b72-b9c9f91dca25

After: 

https://github.com/user-attachments/assets/143cb626-f593-4e7c-a23f-4462919190c0

### DIFF
--- a/profiler/src/profiler/TracyView.cpp
+++ b/profiler/src/profiler/TracyView.cpp
@@ -833,7 +833,7 @@ bool View::DrawImpl()
             ImGui::EndPopup();
         }
     }
-    std::lock_guard<std::mutex> lock( m_worker.GetDataLock() );
+    Worker::MainThreadDataLockGuard lock = m_worker.ObtainLockForMainThread();
     m_worker.DoPostponedWork();
     if( !m_worker.IsDataStatic() )
     {
@@ -1435,7 +1435,7 @@ bool View::Save( const char* fn, FileCompression comp, int zlevel, bool buildDic
     m_userData.StateShouldBePreserved();
     m_saveThreadState.store( SaveThreadState::Saving, std::memory_order_relaxed );
     m_saveThread = std::thread( [this, f{std::move( f )}, buildDict] {
-        std::lock_guard<std::mutex> lock( m_worker.GetDataLock() );
+        Worker::MainThreadDataLockGuard lock = m_worker.ObtainLockForMainThread();
         m_worker.Write( *f, buildDict );
         f->Finish();
         const auto stats = f->GetCompressionStatistics();
--- a/profiler/src/profiler/TracyView_ConnectionState.cpp
+++ b/profiler/src/profiler/TracyView_ConnectionState.cpp
@@ -78,7 +78,7 @@ bool View::DrawConnection()
     }
 
     {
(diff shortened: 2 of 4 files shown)


wolfpld:
The links you have given do not work.

How does this change interact with saving traces (possible while the connection is still active)? If I remember correctly, the save process currently blocks the UI.

MrAntoine77:
> The links you have given do not work.

**Here are the working links, I hope this will make it clearer.**

In certain situations (notably during Tracy initialization), we observed delays caused by a form of starvation of the main thread.
The issue occurs when the worker thread quickly reacquires the mutex after releasing it — before the main thread has had a chance to wake up and acquire it itself.

This leads to a situation where the main thread is repeatedly preempted, causing noticeable lag or even temporary freezes.
Solution

To address this, we introduce a lock coordination mechanism using:
- An atomic flag: `mainThreadWantsLock`
- A condition_variable to wake the worker after the main thread is done

The `MainThreadDataLockGuard`:
- Signals that the main thread wants the lock
- Acquires the lock
- Upon destruction, clears the flag and notifies the worker thread

On the worker side:
- Before continuing processing, it checks the flag
- If the main thread is requesting the lock, it calls wait_for on the condition variable (1ms timeout), giving the main thread a chance to take the lock first

This ensures the main thread gets temporary priority when needed, reducing contention and preventing starvation.

We can see the problem by self profiling :
Before:
Capture d'écran 2025-07-21 093332
After:
Capture d'écran 2025-07-21 093824

This significantly improves responsiveness, especially during initialization, by ensuring the main thread is not indefinitely delayed by worker lock contention.

Before:

https://github.com/user-attachments/assets/fcaa9555-1bd5-45e0-9b72-b9c9f91dca25

After: 

https://github.com/user-attachments/assets/143cb626-f593-4e7c-a23f-4462919190c0

wolfpld:
This needs to follow the coding style.

Lectem:
I took the liberty to fix formatting, adding a comment and rebase to the latest commit from master

PR #1097, opened 2025-07-12
Collect dispatches and counter values with Rocprofv3

erieaton-amd:
This uses (rocprofiler-sdk)[https://github.com/ROCm/rocprofiler-sdk] to collect the dispatch times and GPU performance counters. I had to use a potentially expensive method to attach the counter values to zones, and would like some feedback on the approach that is taken here.


server/TracyWorker.cpp
```diff
@@ -8511,6 +8588,7 @@ void Worker::WriteTimelineImpl( FileWrite& f, const V& vec, int64_t& refTime, in
 
         WriteTimeOffset( f, refTime, v.CpuEnd() );
         WriteTimeOffset( f, refGpuTime, v.GpuEnd() );
+        f.Write( &v.query_id, sizeof( v.query_id ) );
```
wolfpld, line 8591: Tracy version needs to be bumped.

wolfpld:
User manual needs to be updated.

PR #1103, opened 2025-07-17
Server-side symbol request cache to reduce the amount of symbol queries

slomp:
This PR adds a symbol _request_ cache to reduce the amount of symbol queries that the server needs to make to the client.
In my local experiments profiling a very large application, on a 32c/64t processor, at 8KHz, on Windows 11, I was able to reduce the size of the query backlog by 2 orders of magnitude (from O(10M) to O(100K)).


wolfpld:
Implemented in an alternative way on master.

PR #1105, opened 2025-07-18
[Draft] Making callstack symbol queries faster by deprioritizing external name queries

slomp:
Currently, `SymbolQueueItemType::ExternalName` queries are quite lengthy to process, and when they bundle up in the request queue, they can delay responding to other symbol queries considerably.
Given that external name queries (i.e., name of threads belonging to other processes) are generally not as relevant to the profiling session, it makes sense to defer their processing and to respond to more relevant queries first.

### DIFF
--- a/public/client/TracyCallstack.cpp
+++ b/public/client/TracyCallstack.cpp
@@ -9,6 +9,7 @@
 #include "../common/TracyAlloc.hpp"
 #include "../common/TracySystem.hpp"
 
+#include <set>
 
 #ifdef TRACY_HAS_CALLSTACK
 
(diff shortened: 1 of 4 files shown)


Lectem:
I think it would be better to start by making things faster instead of adding another priority mechanism (there is one on the profiler side, albeit less efficient):
1. Avoid querying the process name if we already have it. Currently any TID external name query does both thread and process name querying + send `TidToPid` (we could get TidToPid from from ETW). This is a lot of unnecessary work if we already have the process information.
2. Avoid querying thread names by reading the thread SetName events as suggested in #1110

slomp:
> Avoid querying the process name if we already have it.
Yes, this is something I noticed too, we don't really "cache" process names.
I also noticed that if we "prioritize" just process names, and not thread names, the backlog build-up is much reduced, so maybe we could do something about it.

Are thread names from external processes generally useful? (Asking because so far I never had a need, and usually the process name is enough to identify external intervention in the app trace, so I'd be curious to learn more about use cases for external thread names).

Lectem:
When working on games it has been useful to me only a few times in the past to know what it's doing and why it's interfering. 
But recently I had to deal with and embedded app and it did help a lot more as there are many multi-process components and services running, and you need to know what they do. 

But clearly I think it depends on your usecase and process name already helps a lot. This could be an option I think? But we do need to send TidToPid using ETW at least.

public/client/TracyCallstack.cpp
```diff
@@ -9,6 +9,7 @@
 #include "../common/TracyAlloc.hpp"
 #include "../common/TracySystem.hpp"
 
+#include <set>
```
wolfpld, line 12: Tracy can't allocate using the default allocator.

PR #1108, opened 2025-07-22
Allow to limit zone depth

Lectem:
This PR introduces a new feature that allows users to crop the timeline on a per-thread basis, improving focus and readability when analyzing a thread with many sub-zones.

- Interactive crop toggles: Small round handles are displayed per thread on the left side of the timeline.
- Hovering over a handle displays a red guideline indicating where the crop will begin.
- Clicking a handle highlights it in white and crops the visible timeline to that thread's active zone.
- Clicking again removes the crop, restoring the full view.

This helps reduce visual clutter and makes it easier to focus on deeply nested or dense thread activity.

https://github.com/user-attachments/assets/620f3d77-3da6-461e-9246-8b0a38b76ec8

### DIFF
--- a/profiler/src/profiler/TracyView_ZoneTimeline.cpp
+++ b/profiler/src/profiler/TracyView_ZoneTimeline.cpp
@@ -56,12 +56,12 @@ void View::DrawThread( const TimelineContext& ctx, const ThreadData& thread, con
     }
 
     const auto yPos = wpos.y + offset;
-    const float cropperWidth = ImGui::CalcTextSize(ICON_FA_CARET_DOWN).x;
-    const float cropperCircleRadius = (cropperWidth - 2.0f * GetScale() ) / 2.0f ;
+    const float cropperWidth = ImGui::CalcTextSize( ICON_FA_CARET_DOWN ).x;
+    const float cropperCircleRadius = ( cropperWidth - 2.0f * GetScale() ) / 2.0f ;
     const float cropperAdditionalMargin = cropperWidth + wpos.x; // We add the left window margin for symmetry
     
     // Display cropper if currently limited or if hovering the cropper area
-    const auto threadDepthLimitIt = m_threadDepthLimit.find(thread.id);
+    const auto threadDepthLimitIt = m_threadDepthLimit.find( thread.id );
     const bool displayCropper = ( threadDepthLimitIt != m_threadDepthLimit.end() && threadDepthLimitIt->second <= depth )
         || ( ImGui::GetMousePos().x < wpos.x + cropperAdditionalMargin );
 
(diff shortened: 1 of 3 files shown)


Lectem:
I just pushed the auto-hide/show version. 
The Alt key seems to be captured by the OS somehow (at least on windows, mouse movement stops registering when Alt is pressed), so I preferred not to got that way.

The only downside of hiding the cropper is that it makes the whole lane move a bit, but that's necessary if we want to be able to keep reading the zone names (otherwise they can be hidden).

slomp:
Very handy feature. Maybe for the future (and not as important) but it'd be nice to also be able to chop frames from the bottom (top?) of the call-stack.

Lectem:
Yeah I also thought about this, but not sure how it should look/work.
I suppose we could simply have a key ( Ctrl?) to set the minimum, or use another mouse button for that.

Otherwise we could use some drag'n'drop mechanism from the min/max handles. I'm not sure, would probably need testing.

slomp:
How about right-clicking at a a zone in a level and having an option like "collapse zones below"?
You'd then render a small icon (e.g. an arrow down) in the zones at that level that, when clicked, would expand the hidden levels.

Lectem:
>  option like "collapse zones below"?
Do you mean above ? 

So instead of just having the continuation of the line like this:
image

We would draw this ? 

image

And for the above this ?

image

slomp:
I am no UI expert, and while I like the dots-and-arrows look, based on the amount of new UI code introduced and how specific this UI widget is (and having to hold down a button to make it show up), I am more inclined to think that having these options in the right-click popup menu is a simpler and more familiar way to introduce the feature. It's also less "mouse-travel" on average.

slomp:
Hmmm, yeah, I don't know. Neither PIX nor Superluminal are regarded as providing a good UI experience in my opinion.
One thing I personally like about Tracy is how spacious the timeline view is.

It is ultimately your call, since this is a handy feature and the dots-and-arrow does not take much space when active. You seem pleased with this design, so it's worth giving it a shot in the wild and see what kind of feedback we get.
I would ask to at least add a command in the "right-click popup menu of the thread name" (the one that has "Hide" and "Select in CPU timeline") to toggle on/off the dots-and-arrows widget (instead of just relying on a keyboard key).

Lectem:
> (instead of just relying on a keyboard key

The current version does not use a key ! It is hidden as long as the mouse isn't near the widget (far left of the window), or the crop is active on the lane.

But I don't mind adding controls with the right click ;)

Lectem:
Here's the latest version in action:

https://github.com/user-attachments/assets/9d889f91-ed7c-4223-8cd9-a257f107c605

slomp:
LGTM!

profiler/src/profiler/TracyView_ZoneTimeline.cpp
```diff
@@ -614,4 +601,69 @@ void View::DrawZoneList( const TimelineContext& ctx, const std::vector<TimelineD
...
+    bool isCropped = ( m_threadDepthLimit.find( tid ) != m_threadDepthLimit.end() );
+    const int depthLimit = isCropped ? m_threadDepthLimit[tid] : depth;
+    // If user changes settings to hide Ctx Switches, he would be unable to remove the limit, so set the value to its minimum
+    if( !hasCtxSwitches && isCropped && depthLimit == 0 ) m_threadDepthLimit[tid] = 1;
+
+    const float cropperCenterX = xPos + cropperWidth / 2.0;
+    
+    const auto CircleCenterYForLine = [=]( int lane ) {
+        return yPos + ostep * ( lane + 0.5 );
+        };
```
mcourteaux, line 619: Wrong indent.

PR #1112, opened 2025-07-26
Fix profiler visual glitching on Wayland on NVIDIA because of the lack of implicit DRM sync

ProfFan:
This PR adds support for [Wayland Explicit Sync Protocol](https://wayland.app/protocols/linux-drm-syncobj-v1) so that the Profiler will not have visual glitches when being used with Wayland on NVIDIA GPUs.

Tested under Hyprland.

### DIFF
--- a/profiler/src/BackendWayland.cpp
+++ b/profiler/src/BackendWayland.cpp
@@ -18,6 +18,8 @@
 #include <wayland-client.h>
 #include <wayland-cursor.h>
 #include <wayland-egl.h>
+#include <sys/eventfd.h>
+#include <sys/epoll.h>
 
 #include "wayland-xdg-activation-client-protocol.h"
 #include "wayland-xdg-decoration-client-protocol.h"
@@ -26,6 +28,7 @@
 #include "wayland-viewporter-client-protocol.h"
 #include "wayland-cursor-shape-client-protocol.h"
 #include "wayland-xdg-toplevel-icon-client-protocol.h"
+#include "wayland-linux-drm-syncobj-client-protocol.h"
 
 #include "profiler/TracyImGui.hpp"
 #include "stb_image_resize.h"
@@ -220,6 +223,16 @@ static std::vector<int> s_iconSizes;
 static int s_keyRepeatRate = 0;
 static int s_keyRepeatDelay = 0;
 
+// Explicit sync support using linux-drm-syncobj-v1
+static struct wp_linux_drm_syncobj_manager_v1* s_syncobjMgr;
+static struct wp_linux_drm_syncobj_surface_v1* s_syncobjSurf;
+static struct wp_linux_drm_syncobj_timeline_v1* s_acquireTimeline;
+static struct wp_linux_drm_syncobj_timeline_v1* s_releaseTimeline;
+static int s_acquireTimelineFd = -1;
+static int s_releaseTimelineFd = -1;
+static uint64_t s_acquirePoint = 0;
+static bool s_explicitSyncEnabled = false;
+
 struct Output
 {
     int32_t scale;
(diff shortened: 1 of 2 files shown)


profiler/src/BackendWayland.cpp
```diff
@@ -1169,7 +1256,6 @@ void Backend::NewFrame( int& w, int& h )
         wl_egl_window_resize( s_eglWin, s_width * s_maxScale / 120, s_height * s_maxScale / 120, 0, 0 );
         if( s_fracSurf )
         {
-            wp_viewport_set_source( s_viewport, 0, 0, wl_fixed_from_double( s_width * s_maxScale / 120. ), wl_fixed_from_double( s_height * s_maxScale / 120. ) );
```
wolfpld, line 1172: WTF is that?!

ProfFan: Sorry, accidentally committed some unrelated change here...

wolfpld:
I will need you to explain why this is needed. I see no glitches on Nvidia on my side. Using the protocol directly by the client seems to be a workaround for a deeper problem (in the compositor?), as the client is not directly manipulating Wayland buffers, but rather it uses the OpenGL / WSI interfaces which should be handling this internally.

ProfFan:
> That's not how any of this works.

To be honest I'm not an expert on this topic. Could you elaborate more on this? There is really very little information on the internet on how this thing actually works.

That aside, this patch fixed the glitches that I see but if you are sure that the compositor is the problem I can report to the compositor issue tracker instead.

wolfpld:
Tracy uses OpenGL to render the window contents, and EGL as a Window System Integration (WSI) layer. And that's it.

When submitting the frame render commands is done, a call to `eglSwapBuffers` is made, which **must** follow the [specification](https://registry.khronos.org/EGL/sdk/docs/man/html/eglSwapBuffers.xhtml):

> eglSwapBuffers performs an implicit flush operation on the context (glFlush for an OpenGL ES or OpenGL context, vgFlush for an OpenVG context) bound to surface before swapping.

This is all Tracy needs to care about. The wording above necessitates that no external synchronization needs to happen in the application code.

The Wayland protocol extension you want to use has its place in a completely different layer of that graphics stack. This is explicitly stated in the [protocol documentation](https://wayland.app/protocols/linux-drm-syncobj-v1) you linked in the OP:

> Graphics APIs, like EGL or Vulkan, that manage the buffer queue and commits of a wl_surface themselves, are likely to be using this extension internally. If a client is using such an API for a wl_surface, it should not directly use this extension on that surface, to avoid raising a surface_exists protocol error.

You see the glitches fixed, because your patch adds the functionality that is missing somewhere else. I don't know where that place would be on your system. But realistically, your options are that either Chrome, Xwayland, the EGL WSI implementation, or something else that is used by *many* users is broken in a way that only manifests on your system. Or, that the obscure compositor you use, made by dev who doesn't play well with others and wants to do all on its own, doesn't do something the way it should be done. Take your pick.

ProfFan:
Yes it seems to be an issue with Hyprland itself. I tested on KDE Plasma and it seems KDE plasma's EGL is indeed issuing the sync commands itself:
```
[1960969.536]  -> wp_linux_drm_syncobj_surface_v1#27.set_acquire_point(wp_linux_drm_syncobj_timeline_v1#28, 0, 51)
[1960969.537]  -> wp_linux_drm_syncobj_surface_v1#27.set_release_point(wp_linux_drm_syncobj_timeline_v1#36, 0, 13)
```

I switched to KDE for the moment in the hope of finding a good tiling compositor that does not have issues, thanks!

wolfpld:
> I switched to KDE for the moment in the hope of finding a good tiling compositor that does not have issues, thanks!

I hear good things about https://github.com/YaLTeR/niri/

PR #1116, opened 2025-07-30
Fix clipboard button and add code preview in Zone Info

Lectem:
(Opening this PR on behalf of @MrAntoine77)

---

This PR improves the usability of the ZoneInfo component:

- Moved the clipboard copy button to the left of the text 
  - It would be problematic with long file paths as it would be out of the ZoneInfo window (or if the window is narrow)
  - no more need to move the mouse all the way to the right to copy the content.
- Added code preview on right-click of the zone path, similar to the behavior in FindZone.

These changes aim to make interaction faster and more consistent with other parts of the UI.

Before:
468511050-b0438915-7fc3-4447-8db0-d50259e7c55f

After:
468510301-49d52cdd-eabc-4ccf-a69f-141e775c7a2c

468510328-c2d1d1ba-0c0d-4394-ba61-ab328385fc55

### DIFF
--- a/profiler/src/profiler/TracyImGui.hpp
+++ b/profiler/src/profiler/TracyImGui.hpp
@@ -240,6 +240,15 @@ static constexpr const uint32_t AsmSyntaxColors[] = {
     return res;
 }
 
+[[maybe_unused]] static inline void TextFocusedClipboard( const char* label, const char* value, const char* clipboard, const int clipboardButtonId )
+{
+    TextDisabledUnformatted( label );
+    ImGui::SameLine();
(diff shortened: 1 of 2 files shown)


Lectem:
Right, I forgot to mention the issue sorry.

We saw that but couldn't get ImGui to align the button properly under a font size change...
If we force the button to have the font with "normal" size we're getting this (button not aligned):

image

It didn't really seem better, so we left it like this.

We did try `ImGui::AlignTextToFramePadding();` and without the flag `ImGuiButtonFlags_AlignTextBaseLine` but it doesn't help.

However after looking a bit I found [this](https://github.com/ocornut/imgui/issues/1284), so I just pushed a version that aligns vertically based on the previous font size. Here's the new version:

image

PR #1117, opened 2025-07-30
If we have samples in view and they are displayed, timeline should be displayed even without zones

Lectem:
This is kind of misleading as you may think a thread is inactive but is not. It also made it hard to inspect samples that are at the beginning of a thread where a zone might not have been called yet.

## Current behavior:

https://github.com/user-attachments/assets/a5317f1e-d2e8-4eec-8131-4a2748d7e2c6

## With this PR changes:

https://github.com/user-attachments/assets/6026ce70-a260-4036-823c-797c3f23ef20


Lectem:
This is what the check I added is doing, `m_hasSamples` comes from the track preprocessing and means `hasAnySampleInView`.
Or do you mean I should always display it regardless of the value of `drawSamples` (view mode) ?

PR #1121, opened 2025-07-31
Add faster module caching fallback by using GetModuleHandleExA

GabrielSiliceum:
The main feature of this PR is to avoid calling `EnumProcessModules` when we don't find a module in the cache.
Instead we use the function `GetModuleHandleExA` with the flag `GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS`. 
Thus, we do not iterate through each process's modules, resulting in performance improvement.

For a module at the index 65 out of 70:
- Before: ~2300 microseconds.
- Now : ~100 microseconds.

NB: We also use the flag `GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT` to avoid incrementing the reference count.

### DIFF
--- a/public/client/TracyCallstack.cpp
+++ b/public/client/TracyCallstack.cpp
@@ -555,30 +555,28 @@ ModuleNameAndBaseAddress GetModuleNameAndPrepareSymbols( uint64_t addr )
         }
     }
 
-    HMODULE mod[1024];
-    DWORD needed;
     HANDLE proc = GetCurrentProcess();
+    // Do not use FreeLibrary because we set the flag GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT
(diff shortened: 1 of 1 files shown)


slomp (approved):

slomp:
Made some more exhaustive tests here on a large app, and I can verify it's much faster.
Here are some stats for `tracy::GetModuleNameAndPrepareSymbols()`:
| method | Count | MTPC |
| --- | --- | --- |
| main | ~85K | 554.35us (and growing...) |
| this PR | 250K+ | 95ns |
| [this experiment](https://github.com/wolfpld/tracy/pull/1105/commits/83dd8db972c0f4963ee36d75ff6a054843c7b8c6) | 250K+ | 16.22us |

slomp:
Just complementing my previous comment:
There were 22,900 calls to `GetModuleHandleEx()`, with MTPC 286ns.

PR #1123, opened 2025-07-31
Fix selPos warning on trace worker

EXtremeExploit:
Fixes a warning on selPos about it being possibly used uninitialized, so i decided to initialize it to -1 just like sel above it, hopefully its good enough of an init value

### DIFF
--- a/public/client/TracySysTrace.cpp
+++ b/public/client/TracySysTrace.cpp
@@ -1353,7 +1353,7 @@ void SysTraceWorker( void* ptr )
                 {
                     // Find the earliest event from the active buffers
                     int sel = -1;
-                    int selPos;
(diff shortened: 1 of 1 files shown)


EXtremeExploit:
On the first if inside the loop, if `hdr.type` is NOT PERF_RECORD_SAMPLE on all cases then it doesnt ever get set

wolfpld:
It would be much easier if you would refer to the actual lines in code.

EXtremeExploit:
In the memmove in line 1549

https://github.com/wolfpld/tracy/blob/be7b116a658c0576959609d5bf33f5c392b9753f/public/client/TracySysTrace.cpp#L1549

PR #1124, opened 2025-07-31
Use config-specific lib directory so files don't overwrite each other

AnyOldName3:


Lectem (approved):
This always add an additional `/` even in release. I think it would be better to have `$<IF:$<CONFIG:Release>,,/$<CONFIG>>` but it's a not a big deal.

Otherwise LGTM.

AnyOldName3:
I felt that it was easier to read with the `/` outside the generator expression, and that as there are no surviving operating systems that care about repeated directory separators, that wasn't worth worrying about.

PR #1134, opened 2025-08-12
Refactor Python bindings build

moritz-h:
This is a suggestion for refactoring how the Python bindings are built. I originally started to work on this because the current wheel package has an incorrect RUNPATH set in the binary (it points to the build dir instead of the installed libTracyClient.so), so as soon as the build dir is removed the installed wheel is broken. But I noticed a few more things while working on this, so changes grew a bit larger:
- The current wheel does install headers and binaries in the global include and lib folders. Especially with having very common include path names ("client" and "common") this may can lead to problems when installing the wheel in larger environments.
- The current CMake code does write a lot of files to the source directory. This contradicts CMake best practices. In addition, the code feels a bit like it is reimplementing what the CMake install feature is for.
- The current process to build a wheel feels more complex and error prone then needed. Its required to manually compile the tracy client library before running setuptools, In addition, the process expects a clean source directory, otherwise easily additional things can end up in the wheel.

This PR suggests the following changes:
- Replace setuptools by scikit-build-core to build the python package. Pybind11 supports scikit-build-core and it is the [first example in the docs](https://pybind11.readthedocs.io/en/stable/compiling.html). scikit-build-core is designed for using CMake and calls it automatically. So building the wheel is now as simple as:
```sh
git clone https://github.com/wolfpld/tracy.git
cd tracy/python
pip wheel .
```
- All files are installed with the regular CMake install feature in the background (automatically used by scikit-build-core). So the layout of the installed client library files within the wheel is exactly the same as when installing the client library standalone. This also makes patching the include paths in the headers obsolete. The client library is installed into the tracy_client directory, so not polluting global namespace anymore.
- Set RPath on TracyClientBindings*.so to find correct libTracyClient.so
- Updated Pybind11 to 2.13.6 (for Python 3.13 support)

I tested building the Wheel on Windows 11 with Python 3.13 and Ubuntu 22.04 with Python 3.10.

Docs are not yet updated, can do so if you like this changes.

CC @Chekov2k (seems you did most of the current Python setup, any opinions on this?)


Chekov2k:
On first glance it looks really good. How do you pass in CMake flags to the python wheel? For example `TRACY_ENABLE`?

moritz-h:
You can either edit the pyproject.toml file and write in the last section `[tool.scikit-build.cmake.define]` or pass it as command line parameter `pip wheel . --config-settings cmake.define.TRACY_ENABLE=OFF`.

PR #1141, opened 2025-08-24
Fix the Wayland render loop logic.

mcourteaux:
I asked Gemini Pro what could be wrong, and it pinpointed the use of `wl_display_dispatch()`:

> The key function here is wl_display_dispatch(s_dpy). According to the Wayland documentation, wl_display_dispatch() does the following:
> 
>   - It processes the Wayland event queue.
>   - If the queue is empty, it blocks and waits for new events to arrive from the compositor.
> 

I confirmed that this is the case here: https://www.systutorials.com/docs/linux/man/3-wl_display_dispatch/

This documentation also says that the correct loop logic should be:

>  A correct sequence would be:
> 
> while (wl_display_prepare_read(display) != 0) wl_display_dispatch_pending(display); wl_display_flush(display); poll(fds, nfds, -1); wl_display_read_events(display); wl_display_dispatch_pending(display);
> 

Which matches with the suggestion of Gemini. I tried this and the UI is updating again continuously, with normal CPU usage (i.e., no busy loops).

### DIFF
--- a/profiler/src/BackendWayland.cpp
+++ b/profiler/src/BackendWayland.cpp
@@ -1129,9 +1129,27 @@ void Backend::Show()
 
 void Backend::Run()
 {
-    while( s_running && wl_display_dispatch( s_dpy ) != -1 )
+    wl_display_dispatch( s_dpy );
+    while( s_running )
     {
-        if( tracy::s_config.focusLostLimit && !s_hasFocus ) std::this_thread::sleep_for( std::chrono::milliseconds( 50 ) );
+        // First, dispatch any pending events that have already been received.
+        while( wl_display_prepare_read( s_dpy ) != 0 )
+        {
+            wl_display_dispatch_pending( s_dpy );
+        }
+
+        if( wl_display_read_events( s_dpy ) < 0 )
+        {
+            s_running = false;
+            break;
+        }
+        wl_display_dispatch_pending( s_dpy );
+
+        if( tracy::s_config.focusLostLimit && !s_hasFocus )
+        {
+            std::this_thread::sleep_for( std::chrono::milliseconds( 50 ) );
+        }
+
         s_redraw();
         s_mainThreadTasks->Run();
     }
@@ -1147,7 +1165,7 @@ static void TokenDone( void*, xdg_activation_token_v1* token, const char* str )
 
 constexpr struct xdg_activation_token_v1_listener tokenListener = {
     .done = TokenDone
-}; 
+};
 
 
 void Backend::Attention()


mcourteaux:
What do you mean? It is what it says:

>  Dispatch the display's main event queue.
> 
> **If the main event queue is empty, this function blocks** until there are events to be read from the display fd. Events are read and queued on the appropriate event queues. Finally, events on the main event queue are dispatched. 

Emphasis is mine.

tesselslate:
I experienced the same issue in my own software a few days ago and bisected it back to [this libwayland commit](https://gitlab.freedesktop.org/wayland/wayland/-/commit/ddd348da7ea0889056843cf252729185d306b7b8), which is in libwayland 1.24. It doesn't happen on all compositors (it seems to at least occur on kwin, mutter, and hyprland, but not some wlroots-based ones). I fixed it with a [similar change](https://github.com/tesselslate/waywall/commit/8089fa28b4878d8be53cff1fe2ec7f2e9ce96e26) to this PR. Not sure if there's a more elegant solution (other than using `wl_surface.frame` yourself to always have some events on the main queue).

I didn't bother to investigate too deeply but my guess would be that the only events being sent by the compositor are going to the OpenGL driver's event queue (Mesa uses one, dunno about Nvidia), and the change to `wl_display_dispatch` means that it will not return after reading those events from the Wayland socket. This also explains why moving the mouse causes things to get temporarily unstuck, since it causes the compositor to send events which go to the default queue and allow `wl_display_dispatch` to return.

mcourteaux:
Cool git bisect work! Thanks for chiming in.

> I fixed it with a [similar change](https://github.com/tesselslate/waywall/commit/8089fa28b4878d8be53cff1fe2ec7f2e9ce96e26) to this PR.

Do you have any recommendations towards the use of `wl_display_flush()` and the `poll()`? Gemini didn't put those in, and it worked for me on Sway (which is wlroots, afaik). You seem to have put them in, in your PR. Are they required technically to be correct? It's a bit curious how it "just works" for me without those extra calls. But then again: I'm by no means knowledgeable on the Wayland protocol.

mcourteaux:
> No, you have not quoted that in the original post. I'm not going to read _all_ the documentation concerning a change you are proposing, on a whim. It is your job to properly describe what the change is doing and why.

Apologies. I see: the link I posted to the documentation for that function instead goes to a page that shows documentation for a whole group of functions instead of the one indicated in the URL for some reason.

> @mcourteaux Since `wl_display_dispatch()` is a wrapper around `wl_display_dispatch_queue()` (https://gitlab.freedesktop.org/wayland/wayland/-/blob/ddd348da7ea0889056843cf252729185d306b7b8/src/wayland-client.c#L2199), which is now a wrapper around `wl_display_dispatch_queue_timeout()`, can you try using the timeout version instead with a zero timeout, rather than doing all the internal calls manually?

I thought about that too once I read the comment by @tesselslate, but the commit they referenced shows that the `wl_display_dispatch_queue_timeout` function is *new*. So I think that would break Tracy on older Wayland systems?

I'm definitely open to change this PR in a way, but I don't think we should rely on new functions being present on the system. Perhaps a somewhat ugly C++ `if (requires(...))` clause can determine whether that new function exists, and therefore the behavior of the old one has changed?

> Ah yes, let's change existing behavior and break previously working programs. What the fuck.

Yeah... Well, in their defense, the documentation of that function didn't change. So now it just does what it states it does (i.e., it blocks). This is another unfortunate situation where [Hyrum's Law](https://www.hyrumslaw.com/) applies:

> With a sufficient number of users of an API,
> it does not matter what you promise in the contract:
> all observable behaviors of your system
> will be depended on by somebody.

PR #1142, opened 2025-08-25
Fix wayland loop

mcourteaux:
Continuation of #1141, as Github doesn't allow me to reopen that because of a git history rewrite.

This solution uses the Wayland version number to conditionally use the new version.

cc @tesselslate

### DIFF
--- a/profiler/src/BackendWayland.cpp
+++ b/profiler/src/BackendWayland.cpp
@@ -1129,9 +1129,27 @@ void Backend::Show()
 
 void Backend::Run()
 {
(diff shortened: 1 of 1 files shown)


mcourteaux:
Expect issues from people failing to compile. The latest Ubuntu LTS is not yet shipping wayland 1.24 (containing the function you call). It's still on 1.23, so there will be undefined symbol errors.

https://launchpad.net/ubuntu/+source/wayland

wolfpld:
Yes, this is the solution that will work at runtime. But as I said, I'm not really interested in having workarounds in my code just to fit to some distro's broken design.

PR #1148, opened 2025-09-09
Fix#1147: Cropper now displays properly when docking windows

### LINKED ISSUE #1147: Zone depth panel clips timeline when panels are docked in the left
The "limit zone depth UI" (https://github.com/wolfpld/tracy/pull/1108) ends up drawing over a large portion of the timeline when there is a panel docked to the left side of the window and you hover the mouse over it.

Lectem:
This PR fixes #1147 and another issue related to the red "cut" line which was using the foreground draw list instead of just drawing at the proper time.

@slomp can you test this ?


wolfpld (approved):

slomp:
@Lectem Confirming that the issue is no longer there with this PR! Thanks for the fix!

PR #1151, opened 2025-09-10
Make ZoneNameF() less error-prone by checking format against args

dpelle:
`ZoneNameF(fmt, ...)` was error-prone: if the format is inconsistent with arguments, the code compiled fine without errors or warnings. Use gcc/clang `__attribute__((format(printf, fmt_idx, arg_idx)))` function attribute, so that `ZoneNameF(fmt, ...)` can now check at compilation time the consistency between the format and the arguments.

See https://gcc.gnu.org/onlinedocs/gcc-3.1/gcc/Function-Attributes.html

### DIFF
--- a/public/client/TracyScoped.hpp
+++ b/public/client/TracyScoped.hpp
@@ -12,6 +12,12 @@
 #include "TracyProfiler.hpp"
 #include "TracyCallstack.hpp"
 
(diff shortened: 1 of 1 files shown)


wolfpld:
Why (2, 3) for indices?

PR #1155, opened 2025-09-15
Merge GpuEvent and ZoneEvent

### LINKED ISSUE #1109: CPU and GPU zone inconsistency
In the code, there seem to be a lot of similarity between CPU and GPU zones, except they use separate code paths. But GPU zones are more limited, you can't bring up a statistics window for them like you can with CPU zones. Do you think it would be feasible to unify the zone types code more to reduce duplication and ensure a consistent feature set?

erieaton-amd:
This patch changes GPU to use ZoneEvents, to reduce code duplication and allow consistent feature support between GPU and CPU. The CPU is given a context similar to how GPUs have a context.

The following features now work with GPU events:
Trace Comparison
Zone Search
Zone Histogram
Flame Graph

Fixes #1109

I'm posting this for some feedback on the concept. There are still some major to do items for this:
* Testing with different platforms and build configurations.
* It should be possible to have the extra bytes in ZoneExtra only for GPU contexts. But this will need a whole lot of additional refactoring.
* File format reverse compatibility is broken, but can probably be fixed.

Also, this rework could allow more than one CPU context, which could be used to support cluster tracing.

### DIFF
--- a/cmake/server.cmake
+++ b/cmake/server.cmake
@@ -22,6 +22,7 @@ set(TRACY_SERVER_SOURCES
     TracyTextureCompression.cpp
     TracyThreadCompress.cpp
     TracyWorker.cpp
+    TracyContext.cpp
 )
 
 list(TRANSFORM TRACY_SERVER_SOURCES PREPEND "${TRACY_SERVER_DIR}/")
--- a/profiler/src/profiler/TracyTimelineItemGpu.cpp
+++ b/profiler/src/profiler/TracyTimelineItemGpu.cpp
@@ -31,7 +31,7 @@ const char* TimelineItemGpu::HeaderLabel() const
     }
     else
     {
-        sprintf( buf, "%s context %i", GpuContextNames[(int)m_gpu->type], m_idx );
+        sprintf( buf, "%s context %i", ZoneContextNames[(int)m_gpu->type], m_idx );
     }
     return buf;
 }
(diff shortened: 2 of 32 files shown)


erieaton-amd:
The GpuEvent is removed and replaced with ZoneEvent. The extra fields of GpuEvent (the two extra timestamps, and query_id) are added to ZoneExtra. (This of course uses up extra bytes with CPU events, but that could be fixed I think). Replacing the GpuEvent with ZoneEvent allows a bunch of nearly identical functions to be consolidated.

TracyWorker.cpp contains some parallel data structures for CPU and GPU, like sourceLocationZones and gpuSourceLocationZones. These data structures are moved to a new ZoneContext type, which is a base class for CPU and GPU. GpuCtxData is made a subclass of ZoneContext, and a new context type CPUZoneContext is added for CPU. GpuSourceLocationZones and GpuZoneThreadData are redundant with non-gpu and are removed. The ZoneThreadData and SourceLocationZones are moved to TracyContext.hpp. Both contexts use the same ThreadData type now, but CPU has a subclass to store extra information for sampling data. Worker creates a default CPU context automatically, the context now contains some old Worker fields like m_threadMap, m_data.threads, threadCtx, threadCtxData.

Zone child data is still held by Worker. There is an assumption that the same allocation of ZoneEvent would never be in two contexts, so it should work. Also, all contexts share the original zoneExtra data storage in Worker. If that data structure were moved to contexts, the CPU version could avoid the extra GPU fields.

Most of the remaining changes are refactoring to deal with the above changes.

The UI is changed in some places to have a Combo box to select a context, and display the context with search results.

There is no change to the protocol, but there is to the file format.

PR #1157, opened 2025-09-25
Standardize Tracy macros to not end with a semicolon

### LINKED ISSUE #592: Inconsistent semicolon use in GPU macros
In TracyOpenGL.hpp, there are several macros that have a semicolon at the end, contrary to other macros in the API. This inconsistency causes some minor issues:
- Manually adding the semicolon after using the macro causes some linters to complain about a statement that has no effect
- Omitting the semicolon after the macro creates an (seemingly) inconsistent style, and causes some IDEs to add an extra indentation to the next line

The following macros are affected:
- TracyGpuNamedZone
- TracyGpuNamedZoneC
- TracyGpuZoneTransient
- TracyGpuCollect
- TracyGpuNamedZoneS
- TracyGpuNamedZoneCS
- TracyGpuZoneTransientS

Additionally, if `TRACY_HAS_CALLSTACK` is not defined, the latter four macros will not have an implicit semicolon at the end, leading to further inconsistent behavior (for example, the application not compiling if the user relied on the semicolon that was previously there).

At a glance, the issue appears to affect all of the other GPU headers (TracyVulkan.hpp, TracyD3D11.hpp, TracyD3D12.hpp, and TracyOpenCL.hpp).

BitwiseMage:
Fix #592 

Used the following regex to match the macros ```(#\s*define Tracy.*);{1}$``` and replaced it with group 1 itself ```$1```
This was a quick hack that I did locally and works on my small codebase
Further testing should be done

### DIFF
--- a/public/client/TracyDebug.hpp
+++ b/public/client/TracyDebug.hpp
@@ -3,7 +3,7 @@
 
 #ifdef TRACY_VERBOSE
 #  include <stdio.h>
-#  define TracyDebug(...) fprintf( stderr, __VA_ARGS__ );
+#  define TracyDebug(...) fprintf( stderr, __VA_ARGS__ )
 #else
 #  define TracyDebug(...)
 #endif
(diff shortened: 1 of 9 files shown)


BitwiseMage:
@Lectem I'm not sure if there's a plan to merge this, but by mistake I deleted the head repo.
I already recreated it in case this PR is wanted. Let me know if I need to do something else on my end

PR #1160, opened 2025-09-29
Suppress variable shadowing warnings

Voultapher:
This disables the warnings for MSVC, GCC and Clang in the ZoneScopedXX macros.

The warnings produced are both a false positive since they didn't find a bug *and* they don't happen in user written code, so the user couldn't even do much about it. The previous workaround of using ZoneNamedXXX is a poor solution since the Zone(Text|Name|etc.) macros all rely on the `___tracy_scoped_zone` name.

I hope this is a relatively uncontroversial change. I'm not sure if maybe some of the TracyC.h macros should also be changed. The documentation - which is refreshingly great btw - could probably stay as is since some non-standard compiler may still produce these warnings.

### DIFF
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -1397,7 +1397,7 @@ void Function()
 }
 \end{lstlisting}
 
-This doesn't stop some compilers from dispensing \emph{fashion advice} about variable shadowing (as both \texttt{ZoneScoped} calls create a variable with the same name, with the inner scope one shadowing the one in the outer scope). If you want to avoid these warnings, you will also need to use the \texttt{ZoneNamed} macros.
+This doesn't stop some compilers from dispensing \emph{fashion advice} about variable shadowing (as both \texttt{ZoneScoped} calls create a variable with the same name, with the inner scope one shadowing the one in the outer scope). By default the produced warnings are suppressed when using clang, gcc or MSVC. This behavior can be opted out of by defining \texttt{TRACY\_ALLOW\_SHADOW\_WARNING}. An alternative approach avoids variable name shadowing by manually defining zone names with \texttt{ZoneNamed}. Using this approach requires using the V variants of zone macros like \texttt{ZoneTextV}.
 
 \subsubsection{Exiting program from within a zone}
 
--- a/public/tracy/Tracy.hpp
+++ b/public/tracy/Tracy.hpp
@@ -149,24 +149,26 @@
 #define ZoneTransientN( varname, name, active ) tracy::ScopedZone varname( TracyLine, TracyFile, strlen( TracyFile ), TracyFunction, strlen( TracyFunction ), name, strlen( name ), TRACY_CALLSTACK, active )
 #define ZoneTransientNC( varname, name, color, active ) tracy::ScopedZone varname( TracyLine, TracyFile, strlen( TracyFile ), TracyFunction, strlen( TracyFunction ), name, strlen( name ), color, TRACY_CALLSTACK, active )
 
-#if defined(__clang__)
+#if defined(TRACY_ALLOW_SHADOW_WARNING)
+    #define SuppressVarShadowWarning(Expr) Expr
+#elif defined(__clang__)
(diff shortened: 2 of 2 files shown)


wolfpld:
What warnings?

wolfpld:
@Lectem @slomp what do you think?

slomp:
Hmmm, I'd prefer something more _opt-in_, like:
```
#ifdef TRACY_SUPRESS_WARNINGS
        ...
        _Pragma("clang diagnostic push"); \
        _Pragma("clang diagnostic ignored \"-Wshadow\""); \
        ...
#endif
```
and then leave the macros as they were before.

slomp:
All right, sounds good then!

manual/tracy.tex
```diff
@@ -1397,7 +1397,7 @@ \subsubsection{Variable shadowing}
 }
 \end{lstlisting}
 
-This doesn't stop some compilers from dispensing \emph{fashion advice} about variable shadowing (as both \texttt{ZoneScoped} calls create a variable with the same name, with the inner scope one shadowing the one in the outer scope). If you want to avoid these warnings, you will also need to use the \texttt{ZoneNamed} macros.
+This doesn't stop some compilers from dispensing \emph{fashion advice} about variable shadowing (as both \texttt{ZoneScoped} calls create a variable with the same name, with the inner scope one shadowing the one in the outer scope). By default the produced warnings are suppressed when using clang, gcc or MSVC. This behavior can be opted out of by defining \texttt{TRACY\_ALLOW\_SHADOW\_WARNING}. An alternative approach avoids variable name shadowing by manually defining zone names with \texttt{ZoneNamed}. However this approach is incompatible with \texttt{ZoneText}, \texttt{ZoneValue} and other macros that use the default zone name to reference the zone in the current scope.
```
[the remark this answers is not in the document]
Voultapher: I'm open to a better wording.

public/tracy/Tracy.hpp
```diff
@@ -149,10 +149,34 @@
...
+        _Pragma("GCC diagnostic push"); \
+        _Pragma("GCC diagnostic ignored \"-Wshadow\""); \
+        Expr; \
+        _Pragma("GCC diagnostic pop");
+#elif defined(_MSC_VER) 
+    #define SuppressVarShadowWarning(Expr) \
+        _Pragma("warning(push)"); \
+        _Pragma("warning(disable : 4456)"); \
+        Expr; \
+        _Pragma("warning(pop)");
```
Lectem, line 171: Since we have https://github.com/wolfpld/tracy/pull/1157, we'd better no end the macro with a semicolumn

manual/tracy.tex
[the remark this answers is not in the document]
Lectem: I would just remove the whole sentence I quoted, since all macros have a `Zone****V` variant to be used with `ZoneNamed`

public/tracy/Tracy.hpp
Voultapher: I see the point, at the same time I'm not sure if removing these semicolons here will do what we want. Since

A) We need a semicolon after `Expr` to properly separate it from `_Pragma`.
B) `_Pragma` doesn't need a semicolon after it to work correctly IIRC
C) This is not a macro intended for user use.

Maybe we should mark the macro as private, something like `__SuppressVarShadowWarning`?

Lectem: Right, but it will trigger `-Wextra-semi-stmt` on clang though. Introducing a new warning by suppressing another is a bit weird.

The only way I can think of on top of my head is by adding a `do {} while(0)` (test here https://godbolt.org/z/7b4hjv7Go ), but I don't know if this is something we want?

Lectem: > Yeah, that's why I was advocating for an opt-in toggle that would suppress these warnings right at the top of `Tracy.hpp` (and leave it enabled for the entire translation unit).

I think at that point it'd be better to let the users wrap it through their own headers :)

PR #1166, opened 2025-10-08
Merge ZoneEvent and GpuEvent (redo)

erieaton-amd:
This PR accomplishes part of what #1155 does with fewer changes.

The work is split across three changes:
* Replace GpuEvent with ZoneEvent: This is a smaller revision that accomplishes the removal of GpuEvent, while changing as little else as possible. It is hard to demonstrate the virtue of doing this without further changes; this patch does unify a few small functions.
* Eliminate the function GetZoneCtx: This function turns out to be unnecessary, this patch introduces a helper struct to carry around the context which is also used in the next patch.
* Merge the GetZoneColor functions: The GPU and CPU code is made to use the same code to determine zone color. The helper struct from the previous patch is used by the unified functions to distinguish CPU and GPU events. This allows GPU events to have a custom color (although the protocol doesn't allow it to be set yet).

Overall notes:
* The file format is not changed.
* The protocol is not changed.
* The UI is changed because GPU zones now appear as a different color.
* The original task of ensuring feature parity between GPU and CPU remains unfinished.
* Compared to the previous PR, the ZoneExtra data is expanded only for GPU, saving some memory. Also, the previous PR eliminated the thread field but this doesn't. That uses more memory but avoids having to go search for the thread.

### DIFF
--- a/csvexport/src/csvexport.cpp
+++ b/csvexport/src/csvexport.cpp
@@ -299,9 +299,9 @@ int main(int argc, char** argv)
             const auto& zone_data = it->second;
             for (const auto& zone_thread_data : zone_data.zones)
             {
-                tracy::GpuEvent* gpu_event = zone_thread_data.Zone();
-                const auto start = gpu_event->GpuStart();
-                const auto end = gpu_event->GpuEnd();
+                auto& gpu_event = worker.GetGpuExtra( *zone_thread_data.Zone() );
+                const auto start = gpu_event.GpuStart();
+                const auto end = gpu_event.GpuEnd();
 
                 values[2] = std::to_string( start );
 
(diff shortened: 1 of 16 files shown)


wolfpld:
The whole `struct EventAdapter` is very hard to read for me due to templates, and I don't paricularly like that instead of passing a single object (`ev`), it is now expected to construct a wrapper (`{ev, extra}`).

Given that you have made `GpuExtra` inherit from `ZoneExtra`, the `extra` data is now directly linked to from each GPU `ZoneEvent` (that's what makes sense for me, not something I have read from the proposed changes).

If the above is true, there is no need to construct the wrapper/adapter struct, as the GPU data can be accessed via the extra pointer, and I think it should be possible for the code to know which `ZoneEvent` is GPU and has the `GpuExtra` type extra data from the context of the call.

Does that make sense?

PR #1167, opened 2025-10-08
Remove resolution operators for nvcc compilation

Q-Minh:
[nvcc](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/) can't compile member function calls written as `block->ConcurrentQueue::Block::<member function>()`. Is there a reason to use this calling convention?

### DIFF
--- a/public/client/tracy_concurrentqueue.h
+++ b/public/client/tracy_concurrentqueue.h
@@ -975,7 +975,7 @@ private:
 				auto block = this->tailBlock;
 				do {
 					block = block->next;
-					if (block->ConcurrentQueue::Block::is_empty()) {
+					if (block->is_empty()) {
 						continue;
 					}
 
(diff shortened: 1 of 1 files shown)


Lectem:
See https://github.com/cameron314/concurrentqueue/commit/594bae453a25f83def391763a5c301cf1e36dbd0

Could you try reverting https://github.com/wolfpld/tracy/commit/d3783ae359f3584c5bbb685179aab2b6f7abe410 and see if it fixes your issue on nvcc ?

The namespace was originally added for an old version of GCC, not sure what version does not need this.

WangFengtu1996:
I have same problem!
# env
* win11 ` Visual Studio 17 2022 `
* nvidai  `cuda 12.9`

WangFengtu1996:
> I have same problem!我也有同样的问题！
> 
> # env  环境
> * win11 `Visual Studio 17 2022`
> * nvidai  `cuda 12.9`

But when I use gcc-11.4.0 in ubuntu, it can build successfully!!

slomp:
So, it appears this is the corresponding issue in moodycamel upstream that prompted the namespace scoping: https://github.com/cameron314/concurrentqueue/issues/8

From the discussion, if I understand correctly, the scping had to be added because author of the issue was `using namespace std` somewhere in his code, which caused problems with `block->template is_empty<explicit_context>()` because the compiler was somehow confusing it with `std::is_empty`.

The moodycamel author then decided to add the full scope qualifier to sidestep the issue.

Back to Tracy, I think my arguments in https://github.com/wolfpld/tracy/pull/1372 still hold (reproduced verbatim below), unless I am missing something:
> The modified version in Tracy drops the template argument of the methods in question, so there's no dependent name lookup, and we can just call the methods directly.

wolfpld:
Merged as #1372.

PR #1175, opened 2025-10-16
fix macos ui scaling: fix defaults and rescaling

redthing1:
issues:
- when run on macos right now, it double-scales everything so the UI looks ballooned.
- the 50%/100% scale presets do nothing on macOS

fixes:
- skip GLFW retina DPI when no override, to prevent launching at 2x on macos
- use font scale separately from style scale

this shouldn't affect anything for other platforms, just fixes macos

### DIFF
--- a/profiler/src/main.cpp
+++ b/profiler/src/main.cpp
@@ -100,6 +100,7 @@ static double animTime = 0;
 static float dpiScale = -1.f;
 static bool dpiScaleOverriddenFromEnv = false;
 static float prevScale = 1.f;
+static float prevFontScale = 1.f;
 static int dpiChanged = 0;
 static bool dpiFirstSetup = true;
 static Filters* filt;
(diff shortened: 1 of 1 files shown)


pvande:
I've been building with this patch applied, and have found it helpful.

redthing1:
yeah, it solves the issue for me. would be nice if it could be merged.

wolfpld:
For a macos-only fix this does change a bit too much outside of the apple-specific code paths.

slomp:
Have you folks tried a master branch build of the profiler, after the Dear ImGUI bump to 1.92.6?
https://github.com/wolfpld/tracy/commit/a754bbe3a48a97ea0fae0a1532f723905e9b17e5

I have not experienced the zoom issues anymore, and I can move the Tracy profiler window between the retina display and external displays (with no dpi scaling) without any wonkiness to the UI scaling.

wolfpld:
This is fixed now?

PR #1176, opened 2025-10-16
fix window icon/dock integration with objc on macos

redthing1:
glfw doesn't properly support cocoa, so we see this:
```
Error 65548: Cocoa: Regular windows do not have icons on macOS
```

so this fix adds proper support to set up dock icon on macos


profiler/src/BackendGlfw.cpp
```diff
@@ -190,11 +245,19 @@ void Backend::EndFrame()
 
 void Backend::SetIcon( uint8_t* data, int w, int h )
 {
+#ifdef __APPLE__
+    (void)data;
+    (void)w;
+    (void)h;
```
wolfpld: Remove these.

PR #1180, opened 2025-10-24
Support source/severity for messages

Lectem:
This pull request introduces some new metadata to messages: sources and severity. The main objective was to provide a way to send internal information to the user by default, such as errors related to system sampling for example.

Currently sources are User/Tracy, and severity levels are `Trace, Debug, Info, Warning, Error, Fatal`, based on https://arxiv.org/abs/2109.01192 .

Note that some commits are not finalized / meant to be merged as is, as I'd like some feedback first on the following questions.

1. Should we allow user-defined levels / channels?
2. Should we break the API&ABI to add this to the current `TracyMessage*` defines? 
  - If not, are `tracy::MessageSourceType::User` and `tracy::MessageSeverity::Info` the correct defaults for those?
  - If yes, 
3. This has already been enhanced and discussed a bit on discord, but I'd like feedback on the filters
  - Version A image
  - Version B. Note that currently, toggling the filter requires to click the button or the text precisely image

Still needs to be done:
- Display source/severity of the messages
- Adjust severity level of `TracyDebug` messages by splitting into multiple macros
- Update manual
- Polish code ?

Any kind of feedback is welcome.

### DIFF
--- a/public/common/TracyQueue.hpp
+++ b/public/common/TracyQueue.hpp
@@ -345,16 +345,14 @@ struct QueuePlotDataDouble : public QueuePlotDataBase
     double val;
 };
 
-using MessageMetadata = uint8_t;
-
-enum class MessageSourceType : MessageMetadata
+enum class MessageSourceType : uint8_t
 {
     User,
     Tracy,
     COUNT
 };
 
-enum class MessageSeverity : MessageMetadata
+enum class MessageSeverity : uint8_t
 {
     Trace,   // Broadly track variable states and events in the software program.
     Debug,   // Describes variable states and details about specific internal events in the software, that are useful for investigations.
@@ -365,22 +363,22 @@ enum class MessageSeverity : MessageMetadata
     COUNT
 };
 
-tracy_force_inline MessageMetadata MakeMessageMetadata(MessageSourceType source, MessageSeverity severity)
+tracy_force_inline uint8_t MakeMessageMetadata(MessageSourceType source, MessageSeverity severity)
 {
-    static_assert( (MessageMetadata)MessageSourceType::COUNT < ( 1 << 4 ), "We use 4 bits for the messages source." );
-    static_assert( (MessageMetadata)MessageSeverity::COUNT < ( 1 << 4 ), "We use 4 bits for the messages severity." );
-    return ( (MessageMetadata)severity ) << 4 | (MessageMetadata)source;
+    static_assert( (uint8_t)MessageSourceType::COUNT < ( 1 << 4 ), "We use 4 bits for the messages source." );
+    static_assert( (uint8_t)MessageSeverity::COUNT < ( 1 << 4 ), "We use 4 bits for the messages severity." );
+    return ( (uint8_t)severity ) << 4 | (uint8_t)source;
 }
 
-tracy_force_inline MessageSourceType MessageSourceFromMetadata(MessageMetadata metadata)
+tracy_force_inline MessageSourceType MessageSourceFromMetadata(uint8_t metadata)
 {
-    assert( ( metadata & 0x0F ) < (MessageMetadata)MessageSourceType::COUNT );
+    assert( ( metadata & 0x0F ) < (uint8_t)MessageSourceType::COUNT );
     return (MessageSourceType)( metadata & 0x0F );
 }
 
-tracy_force_inline MessageSeverity MessageSeverityFromMetadata(MessageMetadata metadata)
+tracy_force_inline MessageSeverity MessageSeverityFromMetadata(uint8_t metadata)
 {
-    assert( ( ( metadata & 0xF0 ) >> 4 ) < (MessageMetadata)MessageSeverity::COUNT );
+    assert( ( ( metadata & 0xF0 ) >> 4 ) < (uint8_t)MessageSeverity::COUNT );
     return (MessageSeverity)( ( metadata & 0xF0 ) >> 4 );
 }
 
@@ -400,12 +398,12 @@ struct QueueMessageColor : public QueueMessage
 
 struct QueueMessageMetadata : public QueueMessage
 {
-    MessageMetadata metadata;
+    uint8_t metadata;
 };
 
 struct QueueMessageColorMetadata : public QueueMessageColor
 {
-    MessageMetadata metadata;
+    uint8_t metadata;
 };
 
 struct QueueMessageLiteral : public QueueMessage
--- a/server/TracyWorker.cpp
+++ b/server/TracyWorker.cpp
@@ -5655,7 +5655,7 @@ void Worker::ProcessMessageLiteral( const QueueMessageLiteral& ev )
     msg->thread = CompressThread( td->id );
     msg->color = 0xFFFFFFFF;
     msg->callstack.SetVal( 0 );
-    const MessageMetadata metadata = ev.textAndMetadata.GetTag();
+    const uint8_t metadata = ev.textAndMetadata.GetTag();
     msg->source = MessageSourceFromMetadata( metadata );
     msg->severity = MessageSeverityFromMetadata( metadata );
     if( m_data.lastTime < time ) m_data.lastTime = time;
@@ -5689,7 +5689,7 @@ void Worker::ProcessMessageLiteralColor( const QueueMessageColorLiteral& ev )
     msg->thread = CompressThread( td->id );
     msg->color = 0xFF000000 | ( ev.b << 16 ) | ( ev.g << 8 ) | ev.r;
     msg->callstack.SetVal( 0 );
-    const MessageMetadata metadata = ev.textAndMetadata.GetTag();
+    const uint8_t metadata = ev.textAndMetadata.GetTag();
     msg->source = MessageSourceFromMetadata( metadata );
     msg->severity = MessageSeverityFromMetadata( metadata );
     if( m_data.lastTime < time ) m_data.lastTime = time;
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -2512,7 +2512,7 @@ Profiler::DequeueStatus Profiler::Dequeue( moodycamel::ConsumerToken& token )
                         SendSingleString( (const char*)ptr, size );
                         tracy_free_fast( (void*)ptr );
                         
-                        const MessageMetadata metadata = (MessageMetadata)taggedPtr.GetTag();
+                        const uint8_t metadata = (uint8_t)taggedPtr.GetTag();
                         QueueItem itemWithMetadata;
                         MemWrite( &itemWithMetadata.hdr, item->hdr );
                         MemWrite( &itemWithMetadata.messageMetadata, item->message );
@@ -2530,7 +2530,7 @@ Profiler::DequeueStatus Profiler::Dequeue( moodycamel::ConsumerToken& token )
                         SendSingleString( (const char*)ptr, size );
                         tracy_free_fast( (void*)ptr );
 
-                        const MessageMetadata metadata = (MessageMetadata)taggedPtr.GetTag();
+                        const uint8_t metadata = (uint8_t)taggedPtr.GetTag();
                         QueueItem itemWithMetadata;
                         MemWrite( &itemWithMetadata.hdr, item->hdr );
                         MemWrite( &itemWithMetadata.messageColorMetadata, item->messageColor );


slomp:
> Should we break the API&ABI to add this to the current TracyMessage* defines?

No, `TracyMessage(msg)` as is should just become the "common denominator", e.g. `TracyLog(User, Info, msg)`, as you pointed (good enough in my opinion).

> Should we allow user-defined levels / channels?

I think it's a bit premature and can lead to over-engineering.
Since source and severity are enums backed by an ordinal number, we should be able to extend that in the future.

> I'd like feedback on the filters

I don't have a major preference, but A is a bit more compact and feels less busy to the eyes.

What does the "Reset" button do with the filters? Will it "check-all", besides clearing the text filter?
Will the severity levels also be parsed by the textual filter (fatal, error, warning, etc)?

Lectem:
> What does the "Reset" button do with the filters? Will it "check-all", besides clearing the text filter? 

Currently it checks all + clear text (reset all filters). It used to be linked to the text field and was walled Clear. 

> Will the severity levels also be parsed by the textual filter (fatal, error, warning, etc)?

No, but I suppose we could? One thing that does bother me a little is that you cant easily toggle off all filters but one, with this suggestion this would be possible by typing the severity name. Not sure if it's the best UX though.

Lectem:
> In my opinion, we should introduce just one new "overarching" macro to the public API:
> 
> ```
> #define TracyLogString(level, string, length, color, callstack_depth) ...
> ```

I considered that too to limit the number of potential macros, but I think we need a way to at least differentiate between literal and non-literal to avoid unnecessary copies? or we could use some size hint for litterals such that we could check for it (for example `if(length==-1) { /*litteral*/}`, but that would be introducing a new semantic, I don't like the idea very much since everywhere else we have either `string, size` or just plain `string` params).

> The exploding number of macro combinations is actually a problem.

I'd very much like to avoid it too. If we were to break compat, I would simply expose color+source+severity as a single param "metadata". I'm not sure I would even expose all the depth stack variants (I tend to think that if someone want to customize depth of callstack captures, they can just have their own macros).

So unless we break the current API (which we most likely do not want), or limit the number of macros to not expose every variant (but then we expose ourselves to questions such as "why variant XYZ does not exist), I think the best it to go the "on definition that covers all use-cases, users will create their own wrapper" such as suggested above is the best way to go.

So to sum up I'm thinking of either:

```
#define TracyLogString(sevlevel, txt, length, color, callstack_depth) /*impl*/
#define TracyLogStringL(sevlevel, txt, length, color, callstack_depth) /*impl*/
```

or 

```
// Litterals can use length == -1
#define TracyLogString(sevlevel, txt, length, color, callstack_depth) /*impl*/
```

The only other solution I have in mind would be to use variadics

```
#define TracyLogString(sevlevel, ...) tracy::Profiler::Message( tracy::MessageSourceType::User, sevlevel, __VA_ARGS__ )
```

and let the overload resolution of `tracy::Profiler::Message` take over.

wolfpld:
> * I could use `vsprintf` to bring direct value to the users.

There are `*F` macros already that do formatting.

> * The main issue I have with this is that `*printf` takes a global lock

What what now?!

> * and it would probably be better to do formatting on the server side

I don't think complete and reliable parameter transfer would be something worth doing. Seems very problematic and tiresome to do correctly.

> message culling (runtime) on the client side
> - This only requires checking the severity against a global
> - This means adding a new API or using `TracyParameterSetup` ?

There are parameters already there, if anybody wants to make this controllable. I don't think culling should be done on the client – it's both hostile to non-interactive captures, and destructive to the data you may want to later see. I'd stick to filtering on the server.

slomp:
I suppose we could go about it more incrementally.

The important/transformative feature is the source and the severity idiom, everything else can be experimented upon later.

Let's not worry about the "F" variants and `vsprintf` shenanigans for now, and start by providing the base functionality.
Same goes for culling: for now, we just log what the user wants to log, and they can do their own culling logic in the client to decide whether or not to emit the log message. (I actually like the "active" boolean flag that exist for zones, and one can always add/abstract this boolean in their own macros.)

Lectem:
I'll defer the formatting part to a future PR if needs be, for reference supporting formatting is as simple as adding the following overload
```cpp
    template<typename ...Args>
    static tracy_force_inline void LogString(MessageSourceType source, MessageSeverity severity, uint32_t color, int32_t callstack_depth, const char* fmt, Args&& ...args)
    {
        auto size = snprintf( nullptr, 0, fmt, args... );
        if( size < 0 ) return;

        char* ptr = (char*)tracy_malloc( size_t( size ) + 1);
        snprintf(ptr, size_t( size ) + 1, fmt, args... );

        LogString( source, severity, color, callstack_depth, size_t( size ), ptr );

        tracy_free_fast( ptr );
    }
```

I think this is now ready for a proper review.

manual/tracy.tex
```diff
@@ -1491,6 +1492,17 @@ \subsection{Message log}
 
 If you want to include color coding of the messages (for example to make critical messages easily visible), you can use \texttt{TracyMessageC(text, size, color)} or \texttt{TracyMessageLC(text, color)} macros.
 
+Messages can also have different severity levels: \texttt{Trace}, \texttt{Debug}, \texttt{Info}, \texttt{Warning}, \texttt{Error} or \texttt{Fatal}. 
+The previous macros will log messages with the severity "Info". To log a message with a different severity, you may use the \texttt{TracyLogString} macro that regroups all the functionalities from the previous macros. We recommend writing your own macros, wrapping the different severities for easier use. You may provide a color of 0 if you do not want to set a color for this message.
```
slomp: The previous macros -> These `TracyMessage` macros

public/common/TracyTaggedUserlandAddress.hpp
```diff
@@ -0,0 +1,52 @@
+#ifndef __TRACYTAGGEDPTR_HPP__
```
slomp: This tagged pointer abstraction in the context of the logging feature seems a bit out of place. What is the rationale for it for this branch?

public/client/TracyDebug.hpp
```diff
@@ -1,9 +1,25 @@
...
+		char buffer[4096];																									   \
+		snprintf( buffer, sizeof(buffer), __VA_ARGS__ );																	   \
+		tracy::Profiler::LogString( tracy::MessageSourceType::Tracy, severity, 0, TRACY_CALLSTACK, strlen( buffer ), buffer ); \
+	} while( 0 )
+
 #ifdef TRACY_VERBOSE
 #  include <stdio.h>
-#  define TracyDebug(...) fprintf( stderr, __VA_ARGS__ );
+#  define TracyDebug(...) do { fprintf( stderr, __VA_ARGS__ ); fputc( '\n' ); } while( 0 )
+#elif !defined(TRACY_NO_INTERNAL_MESSAGE)
```
slomp: Thinking ahead here, but if we decide to remove `TRACY_NO_INTERNAL_MESSAGE` we could have `TracyInternalMessage` defer the formatted message to a special TracyInternalLog function where we can have more actions (like always printing to stderr if it is an internal error, capturing callstacks, etc, to make our life debugging Tracy easier).

slomp:
Also, a suggestion (maybe for a follow-up PR): it'd be nice if we could add a column (or a fixed-width glyph) to the left of each message in the "Messages" UI listing with the icon of the severity (and with another icon to indicate "internal" message).

public/common/TracyQueue.hpp
```diff
@@ -343,6 +344,47 @@ struct QueuePlotDataDouble : public QueuePlotDataBase
...
+enum class MessageSourceType : MessageMetadata
+{
+    User,
+    Tracy,
+    COUNT
+};
+
+enum class MessageSeverity : MessageMetadata
+{
+    Trace,   // Broadly track variable states and events in the software program.
```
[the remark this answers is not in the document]
Lectem: I think that if we want to allow the user to extend this, the new values shouldn't be assumed as being ordered or we'll need some kind of advertising of the ordering. I can see people willing to add severities above, under, in the middle...
In general I think this should be enough for 99% use cases anyway?

Note that most libraries I've seen (and those reviewed in the paper linked in desc) order severities this way: highest values are the most critical ones.

Lectem:
> Also, a suggestion (maybe for a follow-up PR): it'd be nice if we could add a column (or a fixed-width glyph) to the left of each message in the "Messages" UI listing with the icon of the severity (and with another icon to indicate "internal" message).

Yes I think the UI can be substantially improved: Adding the severity in the table, collapsing identical messages with a counter, ...

public/common/TracyQueue.hpp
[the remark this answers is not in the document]
slomp: Sounds good! (My experience has been the opposite, with error severity growing down).

Lectem:
Just updated and removed `tracy::Profiler::Message` and `___tracy_emit_message*` since those are implementation details, and users relying on those are liable to API/ABI breaks. This helps clean the code a bit (even though we still have as many `QueueType`s as before internally)

Note that I voluntarily did not expose a `TracyCLogString` macro yet, since it would require adding 2 new ones (can't rely on overloads in C).

public/common/TracyStackFrames.cpp
```diff
@@ -16,7 +16,7 @@ const char* s_tracyStackFrames_[] = {
     "tracy::Profiler::MemFreeCallstack(void const*, int)",
     "tracy::ScopedZone::{ctor}",
     "tracy::ScopedZone::ScopedZone(tracy::SourceLocationData const*, int, bool)",
-    "tracy::Profiler::Message",
+    "tracy::Profiler::LogString",
```
wolfpld: Old function must also be here, as it is used for filtering in the UI and you can have traces from before this change.

public/common/TracyQueue.hpp
```diff
@@ -343,6 +344,47 @@ struct QueuePlotDataDouble : public QueuePlotDataBase
...
+    return ( (MessageMetadata)severity ) << 4 | (MessageMetadata)source;
+}
+
+inline MessageSourceType MessageSourceFromMetadata(MessageMetadata metadata)
+{
+    assert( ( metadata & 0x0F ) < (MessageMetadata)MessageSourceType::COUNT );
+    return (MessageSourceType)( metadata & 0x0F );
+}
+
+inline MessageSeverity MessageSeverityFromMetadata(MessageMetadata metadata)
```
wolfpld: Force inlining.

manual/tracy.tex
```diff
@@ -646,7 +646,8 @@ \subsubsection{Problematic platforms}
 
 \subsubsection{Troubleshooting}
 
-Setting the \texttt{TRACY\_VERBOSE} variable will make the client display advanced information about the detected features. By matching those debug prints to the source code, you might be able to uncover why some of the features are missing on your platform.
+By default, Tracy's diagnostics will be sent as Message logs (section~\ref{messagelog}) to the server.
+Setting the \texttt{TRACY\_NO\_INTERNAL\_MESSAGE} define will disable this feature, but setting the \texttt{TRACY\_VERBOSE} will make the client print advanced information about the detected features to the standard output. By matching those debug prints to the source code, you might be able to uncover why some of the features are missing on your platform.
```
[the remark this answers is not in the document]
slomp: All right, let's keep it.
We can always reanalyze things once the feature is out there and reconsider.

public/common/TracyTaggedUserlandAddress.hpp
```diff
@@ -0,0 +1,52 @@
...
+	TaggedUserlandAddress(TaggedUserlandAddress&&) = default;
+	
+	explicit TaggedUserlandAddress( uint64_t address, uint8_t tag = 0 )
+	{
+		SetAddress( address );
+		SetTag( tag );
+	}
+	// We want a trivial copy
+	TaggedUserlandAddress& operator=(const TaggedUserlandAddress&) = default;
+	TaggedUserlandAddress& operator=(TaggedUserlandAddress&&) = default;
```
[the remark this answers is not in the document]
Lectem: Since we have a non-default constructor (`explicit TaggedUserlandAddress( uint64_t address, uint8_t tag = 0 )`), we need to list the others explicitly, including assignment operators.

server/TracyWorker.cpp
```diff
@@ -972,13 +974,26 @@ Worker::Worker( FileRead& f, EventType::Type eventMask, bool bgTasks, bool allow
             auto msgdata = m_slab.Alloc<MessageData>();
             msgdata->time = ReadTimeOffset( f, refTime );
             f.Read3( msgdata->ref, msgdata->color, msgdata->callstack );
+            if( fileVer >= FileVersion( 0, 12, 5 ) )
```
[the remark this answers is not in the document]
Lectem: Correct, I was waiting to know what the current version was before bumping, and then forgot to do it. Bumped to 0.13.2 and using that for (de)serialization.

public/common/TracyTaggedUserlandAddress.hpp
```diff
@@ -0,0 +1,46 @@
...
+	static constexpr uint64_t kUserspaceVABits = 56;
+	static constexpr uint64_t kKernelOnlyBits = 64 - kUserspaceVABits;
+	static constexpr uint64_t kKernelOnlyBitMask = ( ( uint64_t(1) << kKernelOnlyBits ) - uint64_t(1) ) << kUserspaceVABits;
+public:
+	
+	TaggedUserlandAddress() = default;
+	explicit TaggedUserlandAddress( uint64_t address, uint8_t tag = 0 )
+	{
+		SetAddress( address );
+		SetTag( tag );
```
wolfpld: When constructing, we should just calculate the value of `m_storage` directly, using given inputs. Both these functions want to preserve the other part of the variable, which in case of the first call is garbage, and for the second call can be computed more easy.

public/client/TracyDebug.hpp
```diff
@@ -1,9 +1,25 @@
...
+        TRACY_VERBOSE_EARLY_OUT_COND;																						   \
+		char buffer[4096];																									   \
+		snprintf( buffer, sizeof(buffer), __VA_ARGS__ );																	   \
+		tracy::Profiler::LogString( tracy::MessageSourceType::Tracy, severity, 0, TRACY_CALLSTACK, strlen( buffer ), buffer ); \
+	} while( 0 )
+
 #ifdef TRACY_VERBOSE
 #  include <stdio.h>
-#  define TracyDebug(...) fprintf( stderr, __VA_ARGS__ );
+#  define TracyDebug(...) do { fprintf( stderr, __VA_ARGS__ ); fputc( '\n' ); } while( 0 )
```
wolfpld: Is there `stderr` missing in `fputc` call? The man page states the signature for the function is `int fputc(int c, FILE *stream);`.

manual/tracy.tex
wolfpld: It's not standard output, but standard error output.

public/client/TracyDebug.hpp
Lectem: indeed an error on my part

server/TracyWorker.cpp
```diff
@@ -972,13 +974,26 @@ Worker::Worker( FileRead& f, EventType::Type eventMask, bool bgTasks, bool allow
             auto msgdata = m_slab.Alloc<MessageData>();
             msgdata->time = ReadTimeOffset( f, refTime );
             f.Read3( msgdata->ref, msgdata->color, msgdata->callstack );
+            if( fileVer >= FileVersion( 0, 13, 12 ) )
```
wolfpld: Wrong version.

server/TracyWorker.cpp
```diff
@@ -972,13 +974,26 @@ Worker::Worker( FileRead& f, EventType::Type eventMask, bool bgTasks, bool allow
...
+                msgdata->severity = MessageSeverity::Info;
+            }
             m_data.messages[i] = msgdata;
             msgMap.emplace( ptr, msgdata );
         }
     }
     else
     {
         f.Skip( sz * ( sizeof( uint64_t ) + sizeof( MessageData::time ) + sizeof( MessageData::ref ) + sizeof( MessageData::color ) + sizeof( MessageData::callstack ) ) );
+        if (fileVer >= FileVersion(0, 12, 5))
```
wolfpld: And here.

PR #1187, opened 2025-11-02
Add an Unicode font

### LINKED ISSUE #586: How to support Chinese character?
Unicode character garbled.I tried to change another font, but not working!
image

stdmnpkg:
close #586


wolfpld:
1. There is no information where this font comes from.
2. The "unicode font" option description is misleading. Every font is an unicode font.
3. The font file is embedded 5 times where it could be embedded only once.

nell-dot (approved):

nell-dot (approved):

PR #1208, opened 2025-11-26
Fix build for AppleClang 15

foxtran:
This patch fixes compilation issues with AppleClang 15.0.0. Resulting build works nice.

However, older than Mac OS 13.3 targets definitely require much more workarounds, so the Tracy server can not be configured for such cases.


foxtran:
Mac Book Pro 2017

PR #1213, opened 2025-12-02
Refactoring ETW kernel session

slomp:
Switching from the shared (singleton) "NT Kernel Logger" over to private Kernel logging session (Windows 8+).
VSync events have also been unified under the same private session and event record callback.

I'll be adding CPU frequency/throttling events in the future.

### DIFF
--- a/public/client/TracySysTrace.cpp
+++ b/public/client/TracySysTrace.cpp
@@ -227,20 +227,23 @@ bool SysTraceStart( int64_t& samplingPeriod )
 
 #ifndef TRACY_NO_VSYNC_CAPTURE
     session_vsync = etw::StartUserSession( "TracyVsync" );
-    if ( etw::EnableVSyncMonitoring( session_vsync ) != ERROR_SUCCESS )
-        etw::StopSession( session_vsync );
-    else
+    if ( session_vsync.handle != 0 )
     {
-        consumer_vsync = etw::SetupEventConsumer( session_vsync, EventRecordCallback );
-        if ( consumer_vsync != INVALID_PROCESSTRACE_HANDLE )
+        if ( etw::EnableVSyncMonitoring( session_vsync ) != ERROR_SUCCESS )
+            etw::StopSession( session_vsync );
+        else
         {
-            s_threadVsync = (Thread*)tracy_malloc( sizeof( Thread ) );
-            new(s_threadVsync) Thread( [] (void*) {
-                ThreadExitHandler threadExitHandler;
-                SetThreadPriority( GetCurrentThread(), THREAD_PRIORITY_TIME_CRITICAL );
-                SetThreadName( "Tracy Vsync (ETW)" );
-                etw::EventConsumerLoop( consumer_vsync );
-            }, nullptr );
+            consumer_vsync = etw::SetupEventConsumer( session_vsync, EventRecordCallback );
+            if ( consumer_vsync != INVALID_PROCESSTRACE_HANDLE )
+            {
+                s_threadVsync = (Thread*)tracy_malloc( sizeof( Thread ) );
+                new(s_threadVsync) Thread( [] (void*) {
+                    ThreadExitHandler threadExitHandler;
+                    SetThreadPriority( GetCurrentThread(), THREAD_PRIORITY_TIME_CRITICAL );
+                    SetThreadName( "Tracy Vsync (ETW)" );
+                    etw::EventConsumerLoop( consumer_vsync );
+                }, nullptr );
+            }
         }
     }
 #endif


public/client/windows/TracyETW.cpp
```diff
@@ -0,0 +1,248 @@
...
+    desc.Size = sizeof(fe);
+    desc.Type = EVENT_FILTER_TYPE_EVENT_ID;
+    ENABLE_TRACE_PARAMETERS EnableParamaters = {};
+    EnableParamaters.Version = ENABLE_TRACE_PARAMETERS_VERSION_2;
+    EnableParamaters.EnableProperty = EVENT_ENABLE_PROPERTY_IGNORE_KEYWORD_0;
+    EnableParamaters.SourceId = DxgKrnlGuid;    // or NullGuid? Does it even matter?
+    EnableParamaters.EnableFilterDesc = &desc;
+    EnableParamaters.FilterDescCount = 1;
+
+    ULONG status = EnableProvider(session, DxgKrnlGuid,
```
Lectem: Note for future self: we might need `EVENT_CONTROL_CODE_CAPTURE_STATE` for other events, see https://github.com/google/UIforETW/issues/124#issuecomment-509838160 and https://github.com/GameTechDev/PresentMon/blob/d4385a861a2f4ed2c6db08c875bddf4d6958d797/PresentData/PresentMonTraceSession.cpp#L736

public/client/TracySysTrace.cpp
```diff
@@ -226,273 +212,69 @@ void WINAPI EventRecordCallback( PEVENT_RECORD record )
...
-    {
-        CloseTrace( s_traceHandleVsync2 );
-        CloseTrace( s_traceHandleVsync );
-        s_threadVsync->~Thread();
-        tracy_free( s_threadVsync );
-    }
-
-    CloseTrace( s_traceHandle2 );
-    CloseTrace( s_traceHandle );
+    etw::StopSession(session);
```
Lectem: Note that we now have different behavior than before, as we used to exit `ProcessTrace` before calling `ControlTrace( /*...*/, EVENT_TRACE_CONTROL_STOP)` due to `CloseTrace` being called on the main thread. This is no longer be the case, so the trace will stop before exiting `ProcessTrace`. 

While this is a change I wanted to make in the future (https://github.com/wolfpld/tracy/commit/118da3d94327c74d63de175fbc9f08981cc5d952#diff-6f6f56960a613197987758bbeda41494067a5068f0a5d30b8c7ed101ba394f0bR513), this actually means tracy may take (sometimes a lot) longer to let the application close since we'll wait for the processing of rundown events that will be generated (they were not before because we closed the trace first). 
Since this is not something documented with this PR, it might be better to keep `OpenTrace`/`CloseTrace` where they were until we confirm the impact?

public/client/windows/TracyETW.cpp
```diff
@@ -0,0 +1,248 @@
...
+    session.handle = 0;
+    return ERROR_SUCCESS;
+}
+
+static ULONG EnableProvider(
+    Session& session,
+    const GUID& ProviderId,
+    ULONG ControlCode = EVENT_CONTROL_CODE_ENABLE_PROVIDER,
+    UCHAR Level = TRACE_LEVEL_INFORMATION,
+    ULONGLONG MatchAnyKeyword = 0,
```
Lectem: Is `0` a good default, or should it be 0xFFFFFFFF ? Do we want also want `TRACE_LEVEL_INFORMATION`? I tend not to use default params to avoid such questions (personal preference)

public/client/windows/TracyETW.cpp
```diff
@@ -0,0 +1,284 @@
...
+    BOOL _iswow64;
+    IsWow64Process( GetCurrentProcess(), &_iswow64 );
+    const bool isOs64Bit = _iswow64;
+#endif
+    if( !isOs64Bit )
+        return 0 /* ERROR_SUCCESS */;   // TODO: return error instead?
+    CLASSIC_EVENT_ID stackId[1] = {};
+    stackId[0].EventGuid = EventGuid;
+    stackId[0].Type = Opcode;
+    ULONG status = TraceSetInformation( session.handle, TraceStackTracingInfo, &stackId, sizeof( stackId ) );
```
Lectem: Is there any downside of doing multiple `TraceSetInformation` vs one call with an array of eventids as before?
I suppose the only difference might be performance / failure atomicity?

slomp: If there is, it's not documented or noticeable in practice from what I can tell.

slomp:
I moved the events to `TracyETW.cpp` and added some comments and some alignment/size checks:
https://github.com/wolfpld/tracy/pull/1213/commits/7b5de9e2b9ae74ff1efd3d0453a4aee837dcade8

We could add some runtime checks during system trace initialization to query the events (and respective fields) we care about via the TDH API, but that's going to be a bit annoying to do.

slomp:
(don't merge it yet; there seems to be a quirk with VSync I need to investigate)

slomp:
Actually, never mind...
I thought there was something funny going on with VSync event setup, but it turns out VSync events are just funky.
Sometimes, out of nowhere, I stop receiving them, and that's the case with the current `master` branch as well.
After a while, I start receiving them again.

public/client/windows/TracyETW.cpp
```diff
@@ -0,0 +1,384 @@
...
+static ULONG EnableContextSwitchMonitoring( Session& session )
+{
+    ULONGLONG MatchAnyKeyword = 0;
+    MatchAnyKeyword |= SYSTEM_SCHEDULER_KW_CONTEXT_SWITCH;  // CSwitch events
+    MatchAnyKeyword |= SYSTEM_SCHEDULER_KW_DISPATCHER;      // ReadyThread events
+    ULONG status = EnableProvider( session, SystemSchedulerProviderGuid,
+                                   EVENT_CONTROL_CODE_ENABLE_PROVIDER, TRACE_LEVEL_INFORMATION, MatchAnyKeyword );
+    if( status != ERROR_SUCCESS )
+        return status;
+    status = EnableStackWalk( session, ThreadGuid, 36 );  // ThreadGuid Opcode 36: CSwitch event
```
Lectem: Could use `CSwitch::OpCode`

public/client/windows/TracyETW.cpp
```diff
@@ -0,0 +1,284 @@
...
+#ifdef TRACY_TIMER_QPC
+    props.Wnode.ClientContext = 1;  // 1: QueryPerformanceCounter
+#else
+    props.Wnode.ClientContext = 3;  // 3: CPU Ticks (e.g., rdtsc)
+#endif
+    props.Wnode.Flags = WNODE_FLAG_TRACED_GUID;
+    props.LogFileMode |= EVENT_TRACE_SYSTEM_LOGGER_MODE;
+    props.LogFileMode |= EVENT_TRACE_REAL_TIME_MODE;
+
+    // TODO: should we really be tweaking the buffering parameters?
```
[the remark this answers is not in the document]
slomp: The `BufferSize` is specified in KB, so 1024KB.
The BufferComplete never reported lost events so far (even on a 32c/64t machine).

slomp:
> It looks like what we need is [`EnumerateTraceGuidsEx()`](https://learn.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-enumeratetraceguidsex) with the `TraceGuidQueryInfo` argument. We'll get back a [`TRACE_PROVIDER_INSTANCE_INFO`](https://learn.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-trace_provider_instance_info) struct with an `EnableCount` field, and an array (of that size) of [`TRACE_ENABLE_INFO`](https://learn.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-trace_enable_info) structs with information about each enabled instance, including the `MatchAny|AllKeyword` filters.

Here's what I got from my system:
```
DxgKrnl GUID: 802EC45A-1E99-4B83-9920-87C98277BA9D
EnableCount: 3
    LoggerId: 10 | Name: EventLog-System  | Level: 255 | Any: 0x0C00000000000000 | All: 0x0000000000000000
    LoggerId: 7  | Name: DiagLog          | Level: 255 | Any: 0x0000000002030000 | All: 0x0000000000000000
    LoggerId: 44 | Name: FrameViewService | Level: 4   | Any: 0x0000000008000001 | All: 0x0000000000000001
```
The first two are from Windows itself, but they don't collect VSync events (judging by the Match keyword flags).
The last one, FrameViewService, is from NVIDIA, and is monitoring VSync events.

Experimentally, it looks like only two sessions at most will end up receiving VSync events.

Lectem:
Interesting, so if you use both PresentMon and Tracy at the same time for example, things will be dropped?

slomp:
I added some code to detect (and log) whether a provider has been enabled by another session elsewhere.

I think this concludes all the review changes and quirks I encountered.

Issue #1114, opened 2025-07-28
Crash at exit due to race condition in `StopSystemTracing`


Lectem:
I havent had time to fully investigate yet, but I discovered the issue while using tracy's ToyPathTracer example:

<img width="744" height="629" alt="Image" src="https://github.com/user-attachments/assets/d921df1f-786a-4f44-b86d-fbc519cf06bb" />

Both the `Main Thread` and `Profiler` try to stop and free system tracing resources at the same time.

slomp:
I think we can close this now.

PR #1215, opened 2025-12-03
Fixed race condition around `s_sysTraceThread`

slomp:
There's a subtle race condition when threads read `s_sysTraceThread`.
Since both the main thread (executing `~Profile()`) and the `Profiler::Worker()` thread call `StopSystemTracing()`, the condition:
```
if (s_sysTraceThread) {
    ...
}
```
within `StopSystemTracing()` is _unsynchronized_ because the declaration of `s_sysTraceThread` is not atomic:
```
static Thread* s_sysTraceThread = nullptr;
```

Fixes issue https://github.com/wolfpld/tracy/issues/1114

### DIFF
--- a/public/client/TracyProfiler.cpp
+++ b/public/client/TracyProfiler.cpp
@@ -2148,9 +2148,15 @@ void Profiler::Worker()
     while( s_symbolThreadGone.load() == false ) { YieldThread(); }
 #endif
 
-    // Client is exiting.
 #ifdef TRACY_HAS_SYSTEM_TRACING
-    // Stop filling queues with new data.
+    // On a typical shutdown scenario, the (global) Profiler object is destroyed by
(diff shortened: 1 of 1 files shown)


wolfpld:
Coding style needs to be fixed.

6yry6e:
When one request manual shutdown via `RequestShutdown` worker would break from the loop here https://github.com/wolfpld/tracy/blob/52bfac424346f00288a5460ccb06c4459204d9f0/public/client/TracyProfiler.cpp#L2085
Willing to send the rest of the data.
Profiler won't be destoyed at that moment, as it's a static variable and will live, while app is looping in 
`while (!tracy::GetProfiler().HasShutdonwFinished())`
Then we need to stop system tracing in a worker thread manually to stop filling queue with new data
But if the app doesn't wait or is terminated by any reason, the crash may occur, in a race, that's right :(.

PR #1227, opened 2025-12-16
fixbug build cuda demo error

WangFengtu1996:
1. fix bug build error can not find head file `#include <tracy/Tracy.hpp>`
2. fix bug when not enable TRACY_ENABLE macro, build error  can not find  `tracy::CUDACtx` type


slomp (changes requested):
(I tested the wrong branch; there's an issue with your include directive)

PR #1258, opened 2026-01-15
fix a nullptr crash on android devices

liungkejin:


wolfpld:
Great, the Android implementation is against what the man page shows as valid code.

```c++
           s = getpwnam_r(argv[1], &pwd, buf, bufsize, &result);
           if (result == NULL) {
               if (s == 0)
                   printf("Not found\n");
               else {
                   errno = s;
                   perror("getpwnam_r");
               }
               exit(EXIT_FAILURE);
           }

           printf("Name: %s; UID: %jd\n", pwd.pw_gecos,
                  (intmax_t) pwd.pw_uid);
```

It's also cute how the gecos field is not set anywhere.

```c++
    /*
     * pw_passwd is non-POSIX and unused (always NULL) in bionic.
     * pw_gecos is non-POSIX and missing in bionic.
     */
    dst->pw_passwd = NULL;
```

Anyways. The dereference is intentional and should not be removed.

liungkejin:
Or perhaps it should be changed to

```c
if( passwd && passwd->pw_gecos && *passwd->pw_gecos )

```

wolfpld:
Yes, that's what I hinted at.

liungkejin:
OK, ^_^

PR #1273, opened 2026-02-08
Better support for WSL2/VMs

Lectem:
This does 2 things to make it easier to use tracy on WSL2/VMs:
- correctly align Linux systrace code to use the same clock as tracy
- add `TRACY_DISALLOW_HW_TIMER` so that we can prevent hardware time usage. On VMs, the instruction is available but systrace may not use it. Ideally we would be able to detect it, but this option at least allows us to force fallback to CLOCK_MONOTONIC_RAW.

### DIFF
--- a/public/client/TracySysTrace.cpp
+++ b/public/client/TracySysTrace.cpp
@@ -885,7 +885,6 @@ bool SysTraceStart( int64_t& samplingPeriod )
         pe.disabled = 1;
         pe.inherit = 1;
         pe.config = switchId;
-        
         if( useMonotonicClockRaw )
         {
             pe.use_clockid = 1;


public/client/TracyProfiler.hpp
```diff
@@ -28,7 +28,7 @@
 #  include <mach/mach_time.h>
 #endif
 
-#if ( (defined _WIN32 && !(defined _M_ARM64 || defined _M_ARM)) || ( defined __i386 || defined _M_IX86 || defined __x86_64__ || defined _M_X64 ) || ( defined TARGET_OS_IOS && TARGET_OS_IOS == 1 ) || ( defined __APPLE__  && defined __MACH__ && TARGET_CPU_ARM64 ) )
+#if !defined TRACY_DISALLOW_HW_TIMER && ( (defined _WIN32 && !(defined _M_ARM64 || defined _M_ARM)) || ( defined __i386 || defined _M_IX86 || defined __x86_64__ || defined _M_X64 ) || ( defined TARGET_OS_IOS && TARGET_OS_IOS == 1 ) || ( defined __APPLE__  && defined __MACH__ && TARGET_CPU_ARM64 ) )
```
wolfpld: I think we need to somehow split this line to make it clear what it does.

public/client/TracySysTrace.cpp
```diff
@@ -666,11 +675,12 @@ bool SysTraceStart( int64_t& samplingPeriod )
 #endif
     pe.disabled = 1;
     pe.freq = 1;
-    pe.inherit = 1;
-#if !defined TRACY_HW_TIMER || !( defined __i386 || defined _M_IX86 || defined __x86_64__ || defined _M_X64 )
-    pe.use_clockid = 1;
-    pe.clockid = CLOCK_MONOTONIC_RAW;
-#endif
+    pe.inherit = 1;
```
wolfpld: Whitespace.

Issue #1278, opened 2026-02-11
Zooming-in with mouse wheel hangs the UI sometimes


slomp:
There seems to be an issue with the timeline zoom logic on MacOS when using a mouse wheel.

Specifically, in the following snippet, `t0` and `t1` can cross sides, such that `t0 > t1`:
https://github.com/wolfpld/tracy/blob/ea8eca6a45869513587f68144ded6b6960ce53d5/profiler/src/profiler/TracyView_Timeline.cpp#L152-L156

This leads to a debug assertion later on:
https://github.com/wolfpld/tracy/blob/ea8eca6a45869513587f68144ded6b6960ce53d5/profiler/src/profiler/TracyView_Timeline.cpp#L268C5-L268C41
```cpp
assert( m_vd.zvStart < m_vd.zvEnd );
```

But in `NDEBUG` builds of the profiler, the code marches on, and from there it ends up on some start-to-end iteration which will take a _long_ time to "wrap around".

@aaltman_adobe

PR #1281, opened 2026-02-11
Fix UI hangs due to mouse-wheel zooming

slomp:
Fixes https://github.com/wolfpld/tracy/issues/1278


wolfpld:
Is this solution correct, though? Maybe there should be at least 1 unit of difference, so that something can be displayed at least? Also consider the code above, where `p1` and `p2` are calculated from `zoomSpan`, which would be 0 from now on.

PR #1284, opened 2026-02-17
Ensure executable path is inspected when looking for PDB files

slomp:
`SymInitialize()` and `SymLoadModuleEx()` appear to operate on a different set of rules when looking for matching PDB files.

---

While `SymLoadModuleEx` is able to resolve PDB files for modules (load-time and run-time) and will look at the location where the module is located, it appears to have no effect for the host executable image itself.

For the host executable, only `SymInitialize` seems to actually run logic to find an appropriate PDB file for the executable...
According to `SymInitialize` docs:
> Note that the directory that contains the executable file for the process is not automatically part of the search path.
https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-syminitialize

This is somewhat contradictory to this article:
>the symbol handler searches the directory that contains the module for which symbols are being searched.
https://learn.microsoft.com/en-us/windows/win32/debug/symbol-paths

If we put some bits and pieces of both articles together, the intuition for the search order/logic would go like this:
1. Current Working Directory (only for the executable image itself)
2. `_NT_SYMBOL_PATH`
3. `_NT_ALT_SYMBOL_PATH`
4. **The location of the module** (only for modules)

I've tried different versions of `DbgHelp.dll` and `SymSrv.dll` to no avail.

---

In addition, if an executable does not specify a full path to a pdb file in its "Debug Directories" PE section (use `dumpbin /headers <executable>` to check it), `SymInitialize` will also fail to find the pdb file, even if it is sitting at the current working directory...

---

Given all of the above, the easiest workaround is for Tracy to append the path of the executable to `_NT_SYMBOL_PATH`. I thought about passing it via the `UserSearchPath` argument to `SymInitialize`, but I'm a bit concerned that it may break the search logic for modules. The documentation is not very clear...
https://learn.microsoft.com/en-us/windows/win32/debug/initializing-the-symbol-handler

### DIFF
--- a/public/client/TracyCallstack.cpp
+++ b/public/client/TracyCallstack.cpp
@@ -411,11 +411,21 @@ void DbgHelpInit()
     // append executable path to the _NT_SYMBOL_PATH environment variable
     char buffer [32767];  // max env var length on Windows (including null-terminator)
     DWORD length = GetEnvironmentVariableA( "_NT_SYMBOL_PATH", buffer, sizeof( buffer ) );
-    buffer[length] = ';';
-    buffer[++length] = '\0';
-    length += GetModuleFileNameA( NULL, &buffer[length], sizeof( buffer ) - length );
-    while( length > 0 && buffer[--length] != '\\ ')
-        buffer[length] = '\0';
+    if( length > sizeof( buffer ) ) {
+        SymError( "GetEnvironmentVariableA", GetLastError() );
+    } else if( length + 1 >= sizeof( buffer ) ) {
+        SymError( "_TracyAppendEnvironmentVariable", ERROR_INSUFFICIENT_BUFFER );
+    } else {
+        buffer[length] = ';';
+        buffer[++length] = '\0';
+        length += GetModuleFileNameA( NULL, &buffer[length], sizeof( buffer ) - length );
+        if( length >= sizeof( buffer ) && GetLastError() == ERROR_INSUFFICIENT_BUFFER ) {
+            SymError( "GetModuleFileNameA", GetLastError() );
+        } else {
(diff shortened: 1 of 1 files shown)


slomp:
Note for the future:
We could add a symbol handler callback to get notifications about symbol loading and lookups.
https://learn.microsoft.com/en-us/windows/win32/debug/initializing-the-symbol-handler
(I experimented with it a bit, but it did not gave me much valuable insights)

public/client/TracyCallstack.cpp
```diff
@@ -392,8 +408,23 @@ void DbgHelpInit()
     DBGHELP_LOCK;
 #endif
 
-    SymInitialize( GetCurrentProcess(), nullptr, true );
-    SymSetOptions( SYMOPT_LOAD_LINES );
+    // append executable path to the _NT_SYMBOL_PATH environment variable
+    char buffer [32767];  // max env var length on Windows (including null-terminator)
+    DWORD length = GetEnvironmentVariableA( "_NT_SYMBOL_PATH", buffer, sizeof( buffer ) );
```
slomp: Even though a 32K env var is unlikely to happen in practice, I should do some proper length check to avoid potential attacks...

dedmen:
Note somewhat unrelated.
This fixed the case where the buffer for GetModuleFileNameA is insufficient, but that problem also exists at https://github.com/wolfpld/tracy/blob/8c2a970222529417871c15791c0d5acb22faf946/public/client/TracyProfiler.cpp#L438 and is not fixed there.
Sorry for the sloppy comment here, I don't have time to deal with it properly now

PR #1286, opened 2026-02-18
Have Touchpad gestures either scroll or zoom the timeline, but not both together

slomp:
This makes using Tracy with a touchpad/trackpad much more bearable.
Before this change, a touchpad zoom gesture would coming with a horizontal scroll gesture, making navigation disorienting and frustrating.
For mouse usage, this has virtually no effect (unless the user is spinning the horizontal wheel and the regular wheel simultaneously, but even then, I wager it's best to prioritize one or the other).

### DIFF
--- a/profiler/src/profiler/TracyView_Timeline.cpp
+++ b/profiler/src/profiler/TracyView_Timeline.cpp
@@ -87,7 +87,7 @@ void View::HandleTimelineMouse( int64_t timespan, const ImVec2& wpos, float w )
         }
     }
 
-    const bool wheel_scroll = abs(io.MouseWheelH) > abs(io.MouseWheel);
+    const bool wheel_scroll = abs( io.MouseWheelH ) > abs( io.MouseWheel );
     if( IsMouseDragging( 1 ) || wheel_scroll )
     {
         m_viewMode = ViewMode::Paused;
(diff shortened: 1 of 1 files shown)


wolfpld:
What I would like instead:

The touchpad scroll action should initiate a "mode lock-in" mechanism with two operating modes:

1. Zoom mode - vertical scroll gesture zooms the timeline
2. Pan mode - scroll gesture pans the view both horizontally and vertically

The initial motion direction determines which mode is locked:
- Greater initial vertical motion → zoom mode
- Greater initial horizontal motion → pan mode

Once locked, the mode persists for the entire gesture duration, even during pauses where no scroll events arrive for multiple frames. The gesture ends only when the user lifts their finger from the touchpad.

Wayland provides the necessary signals:
- wl_pointer.axis_source indicates FINGER for touchpad gestures vs WHEEL for mouse scroll
- wl_pointer.axis_stop signals when the finger is lifted from the touchpad
- wl_pointer.frame groups related events together

ImGui does not provide a path for gesture lifecycle events:
- ImGui's event system is frame-based with no concept of gesture start/end
- ImGuiInputEventMouseWheel only carries (WheelX, WheelY, MouseSource)
- No existing event type exists for scroll gesture lifecycle
- The MouseSource field can indicate touchpad origin, but provides no gesture boundary signaling

The clean architectural solution requires an ImGui contribution to add gesture lifecycle events. Without that, any implementation will require either a parallel signaling mechanism or unreliable heuristics.

PR #1316, opened 2026-03-23
GPU: D3D12: remove NewFrame() API + drop unresolved queries when under pressure

### LINKED ISSUE #1294: [DirectX 12] Tracy GPU profiling deadlocks if the device is removed
Our program supports DirectX 12 device recovery - that means, if the DirectX 12 device is removed (for example due to an invalid API call, a device hang or the user physically removed his GPU), the program will gracefully teardown the old DirectX 12 structures and create new ones.

This works, until we start using tracy: The destructor of the DirectX 12 contxt (tracy::D3D12QueueCtx), which gets invoked during the graceful teardown of our DirectX 12 context, contains this line:
`
 while (m_payloadFence->GetCompletedValue() != m_activePayload)
     /* busy-wait ... */;
`

When a DirectX 12 device is removed, all fences set their completed value to UINT64_MAX. Therefore, this code will loop forever, causing a deadlock. 
Changing the comparison operator from not-equal (`!=`) to less-than (`<`) fixes the issue (we implemented this fix locally for us, I'm just reporting it here so it can be fixed officially).
To test this, you can use the function `ID3D12Device5::RemoveDevice()`.

slomp:
This PR removes the need for calling `NewFrame()` periodically.
(it also addresses and fixes https://github.com/wolfpld/tracy/issues/1294)

Some applications may run "headless" or may use the GPU for compute-only purposes, and do not have the concept of a frame boundary. Timestamp queries are now produced and collected in circular fashion.

In addition, the code is now resilient to situations where timestamp queries are reserved/dispensed, but the enclosing command list is never submitted to the GPU for execution. Such cases will flag warning messages and will "stall" progress, forcing "unresolvable" timestamps queries to be dropped. This also tricky covers corner cases where timestamp query ids are dispensed in one order, but the command lists are sent to the GPU queue in opposing order, or when an operation takes unusually long to complete with respect to everything else.

As such, queue signals are no longer required, nor useful: even with Signal/Wait, there are no guarantees that once the queue sets the signal value, all queries up to that value have indeed completed, since queries could be sent/executed out of the order. in which they were generated, or may not have been submitted to the queue yet (or ever).

----

The only scenario where ambiguity can still happen is as follows:

1. a command buffer (CB1) is recorded, but not yet executed
2. time goes by, and several command buffers are subsequently recorded and executed
3. Collect() ends up giving up on CB1's "reserved" queries
4. CB1 is finally sent to the queue and executed
5. but around the same instant, another command buffer is recorded, claiming query ids that clash with CB1
6. the GPU will resolve the CB1 timestamps, and write to the timestamp heap
7. Collect() is not going to be able to disambiguate the timestamp

While technically possible, it is quite a contrived case. There does not seem to be a proper solution for the case above without some amount of costly query tracking along with some CPU-GPU synchronization on top of it. The current implementation is also prone to the same "bad-timing" problem anyway.

### DIFF
--- a/public/tracy/TracyD3D12.hpp
+++ b/public/tracy/TracyD3D12.hpp
@@ -69,8 +69,6 @@ namespace tracy
     {
         friend class D3D12ZoneScope;
 
-        static constexpr UINT64 InvalidTimestamp = 0;   // arbitrary choice (afaik, it's technically a valid timestamp value)
-
         uint8_t m_contextId = 255;  // TODO: apparently, 255 means "invalid id"; is this documented somewhere?
 
         std::mutex m_collectionMutex;
@@ -90,8 +88,16 @@ namespace tracy
 
         uint32_t m_queryLimit = 0;
 
-        using AgeClock = std::chrono::high_resolution_clock;
-        std::vector<AgeClock::time_point> m_queryRequestTime;
+        struct CollectWindow
+        {
+            using AgeClock = std::chrono::high_resolution_clock;
+            static constexpr uint64_t capacity = 4 * 1024;
+            uint64_t rangeBegin = 0;
+            uint64_t rangeEnd   = 0;
+            AgeClock::time_point ageStart   = AgeClock::time_point::max();
+            UINT64 latestKnownGpuTimestamp = 0;
+            UINT64 shadowBuffer[capacity] = {};
+        } m_window;
 
         UINT64 m_prevCalibrationTicksCPU = 0;
 
@@ -200,8 +206,6 @@ namespace tracy
                 TracyD3D12Panic("Failed to create query readback buffer.", return);
             }
 
-            m_queryRequestTime.resize(m_queryLimit);
-
             {
                 D3D12_RANGE zeroRange{ 0, m_queryLimit * sizeof(UINT64) };
                 void* buffer = nullptr;
@@ -212,8 +216,7 @@ namespace tracy
                 UINT64* timestampBuffer = static_cast<UINT64*>(buffer);
                 for (uint64_t i = 0; i < m_queryLimit; ++i)
                 {
-                    timestampBuffer[i] = InvalidTimestamp;
-                    m_queryRequestTime[i] = AgeClock::time_point::max();
+                    timestampBuffer[i] = 0;
                 }
                 // TODO: any advantages to making this Map() persist?
 #if TRACY_D3D12_PERSISTENT_TIMESTAMP_BUFFER
@@ -246,6 +249,10 @@ namespace tracy
             }
             m_lastEmittedGpuTimestamp = gpuTimestamp;
 
+            // initialize CollectWindow
+            UpdateLatestKnownGpuTimestamp(gpuTimestamp);
+            AdvanceCollectWindow();
+
             // Save the device cpu timestamp, not the profiler's timestamp.
             m_prevCalibrationTicksCPU = cpuTimestamp;
 
@@ -277,6 +284,8 @@ namespace tracy
             // (need to create an ID3D12Fence and associate an Event object to it)
 
             // wait for all pending queries to be collected (busy-wait...)
+            // TODO: need to relax the while condition, since now m_previousCheckpoint
+            // advances in batches of the window size, but not the query counter...
             while (m_previousCheckpoint.load() != m_queryCounter.load())
             {
                 Collect();
@@ -308,6 +317,7 @@ namespace tracy
 #ifdef TRACY_ON_DEMAND
             if (!GetProfiler().IsConnected())
             {
+                // TODO: reset the collect window here
                 m_previousCheckpoint = m_queryCounter.load();
                 return;
             }
@@ -318,28 +328,23 @@ namespace tracy
             // Only one thread is allowed to collect timestamps at any given time
             // but there's no need to block contending threads
             if (!m_collectionMutex.try_lock())
-            {
                 return;
-            }
             std::unique_lock lock (m_collectionMutex, std::adopt_lock);
 
-            // Establish a range of queries to collect:
-            // the range starts at the "previous checkpoint" (known to have been resolved already),
-            // and can extend up to the last query id emitted so far.
-            uint64_t begin = m_previousCheckpoint.load();
-            uint64_t latestCheckpoint = m_queryCounter.load();
+            auto now = CollectWindow::AgeClock::now();
 
-            uint32_t count = RingCount(begin, latestCheckpoint);
-            if (count == 0)
+            uint64_t windowSize = m_window.rangeEnd - m_window.rangeBegin;
+            if (windowSize < m_window.capacity)
             {
-                return; // no pending timestamp queries
+                uint64_t latestIssued = m_queryCounter.load();
+                uint64_t trail = latestIssued - m_window.rangeBegin;
+                windowSize = std::min(trail, m_window.capacity);
+                m_window.rangeEnd = m_window.rangeBegin + windowSize;
+                m_window.ageStart = now;
             }
-            
-            // paranoid check...
-            if (count >= RingSize())
-            {
-                TracyD3D12Panic("Collect: FULL! too many pending timestamp queries.", return);
-            }
+
+            if (windowSize == 0)
+                return;
 
 #if TRACY_D3D12_PERSISTENT_TIMESTAMP_BUFFER
             UINT64* timestampBuffer = m_timestampBuffer;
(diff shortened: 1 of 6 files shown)


public/tracy/TracyD3D12.hpp
```diff
@@ -479,14 +676,20 @@ namespace tracy
 }
 
 #undef TracyD3D12Panic
+#undef TracyD3D12Log
+#undef TracyD3D12Assert
+#undef TracyD3D12Break
+#undef TracyD3D12Debug
+#undef TRACY_D3D12_PERSISTENT_TIMESTAMP_BUFFER
+#undef TRACY_D3D12_DEBUG
```
Lectem, line 684: `TRACY_D3D12_DEBUG_LEVEL` ?

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+        }
 
-            if (FAILED(m_readbackBuffer->Map(0, &mapRange, &readbackBufferMapping)))
+        bool Wait(uint64_t queryTicket, uint64_t timeout_ms)
+        {
+            ZoneScopedC(Color::Red4);
+            TracyD3D12Debug( ZoneValue(m_contextId) );
+            TracyD3D12Debug( ZoneValue(queryTicket) );
+            TracyD3D12Debug( ZoneValue(RingIndex(queryTicket)) );
+            auto Now = GetTickCount64;
```
Lectem, line 372: why introduce this `Now`?
Also, elapsed will be `int` instead of `uint64_t`

public/tracy/TracyD3D12.hpp
```diff
@@ -173,10 +200,10 @@ namespace tracy
                 TracyD3D12Panic("Failed to create query readback buffer.", return);
             }
 
-            if (FAILED(device->CreateFence(0, D3D12_FENCE_FLAG_NONE, IID_PPV_ARGS(&m_payloadFence))))
-            {
-                TracyD3D12Panic("Failed to create payload fence.", return);
-            }
+            UINT64* timestampBuffer = MapTimestampBuffer();
```
Lectem, line 203: should handle `MapTimestampBuffer` failure?

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+            TracyD3D12Debug( ZoneValue(m_contextId) );
+
+            uint64_t earliestTicket = m_previousCheckpoint;
+            uint64_t endTicket = m_queryCounter;
+            TracyD3D12Debug( ZoneValue(earliestTicket) );
+            TracyD3D12Debug( ZoneValue(endTicket) );
+            if (Distance(earliestTicket, endTicket) <= 0)
                 return;
+
+            UINT64* timestampBuffer = MapTimestampBuffer();
```
Lectem, line 325: should handle `MapTimestampBuffer` failure?

public/tracy/TracyD3D12.hpp
```diff
@@ -200,53 +227,59 @@ namespace tracy
...
+            // TODO: could use queue->Signal() to inject a progress point in the queue
+            // and the immediately wait for the signal, in order to avoid busy-waiting
+            // (need to create an ID3D12Fence and associate an Event object to it)
+            // NOTE: even with Signal(), there are no guarantees the queries were sent
+            // to the GPU for execution, so the Signal() does not give us much "signal"
+ 
+            // collect all pending queries up to the latest known query
+            uint64_t endTicket = m_queryCounter;
+            uint64_t lastIssuedTicket = endTicket - 2;
+            Drain(lastIssuedTicket, 200);
```
Lectem, line 268: Why `200`ms ?

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+            // may end up being collected as if it belonged to the the new query.
+            const uint64_t ticket = m_queryCounter.fetch_add(2, std::memory_order_relaxed);
+            if (Distance(m_previousCheckpoint, ticket) >= RingCapacity())
             {
                 ZoneScopedC(Color::Red4);
-                ZoneValue(int64_t(m_contextId));
-                TracyD3D12Panic("Submitted too many GPU queries!");
-                // TODO: get rid of NewFrame() and make collection "circular"
-                // TODO: decide what to do when "full" (collect, or return an error-id?)
+                TracyD3D12Log(Warning, "Too many pending GPU queries: stalling!");
```
Lectem, line 491: That's dropping since `gracePeriod_ms` is 0 no?

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+        {
+            // difference accounting for unsigned wrap-around
+            return static_cast<int64_t>(end - begin);
+        }
 
-            // Recalibrate to account for drift.
-            RecalibrateClocks();
+        uint64_t RetireTicket(uint64_t ticket)
+        {
+            TracyD3D12Assert( m_previousCheckpoint == ticket );
```
Lectem, line 468: this atomic load is `seq_cst`, is that wanted?

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+            ticket += 2;
+            m_previousCheckpoint.store(ticket);
```
Lectem, line 470: I think the atomicity of `m_previousCheckpoint` needs to be explicit. do you really want `seq_cst` here and in `NextQueryId` ? I'm guessing what you really want is acquire/release semantics?

From my understanding any thread can see the value update, but only the one collecting (guarded by a mutex) can write?

In any case, the pattern "update local `ticket` then `store` then `return ticket`" usually raises flags in my head as if you use an atomic here you would in general do `fetch_add(2)` and keep the return value+increment (which may be `!= ticket`). In this case this is fine because return val is unused and we are always writing under a mutex but I find it confusing.

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+            // dropping a query due to timeout, but said query is indeed submitted to the
+            // GPU queue for execution later on. This "late" query will eventually become
+            // resolved by the GPU asynchronously writting to the corresponding query slot.
+            // The newly produced query below could have matching ids/slot with said query.
+            // From there, there's no bullet-proof way for Collect() to distinguish between
+            // the timestamp written to the query slot belonging to the old/late query, or
+            // to the query that has just been generated. The value of the "late" query may
+            // may end up being collected as if it belonged to the the new query.
+            const uint64_t ticket = m_queryCounter.fetch_add(2, std::memory_order_relaxed);
+            if (Distance(m_previousCheckpoint, ticket) >= RingCapacity())
```
Lectem, line 488: I'd like the `m_previousCheckpoint` to have explicit memory ordering here.

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+        }
 
+    private:
+        void Collect(std::unique_lock<std::mutex>& lock, uint64_t targetTicket, bool urgent)
+        {
+            ZoneScopedC(Color::Red4);
+            TracyD3D12Assert( lock.owns_lock() );
+            TracyD3D12Debug( ZoneValue(m_contextId) );
+
+            uint64_t earliestTicket = m_previousCheckpoint;
```
Lectem, line 318: If the only writes done to `m_previousCheckpoint` are done behind the collect lock, you can make this a `relaxed` `load`

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+        {
+            ZoneScopedC(Color::Red4);
+            TracyD3D12Debug( ZoneValue(m_contextId) );
+            TracyD3D12Debug( ZoneValue(queryTicket) );
+            TracyD3D12Debug( ZoneValue(RingIndex(queryTicket)) );
+            auto Now = GetTickCount64;
+            auto ini = Now();
+            auto elapsed = 0;
+            // TODO: could use condition variable to avoid spurious lock + collect iterations
+            while ((Distance(m_previousCheckpoint, queryTicket) >= 0) && (elapsed < timeout_ms))
```
Lectem, line 376: explicit memory ordering for the load please

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
             {
-                TracyD3D12Panic("Failed to map readback buffer.", return);
+                std::unique_lock lock (m_collectionMutex);
+                Collect(lock, queryTicket, false);
+                elapsed = Now() - ini;
             }
+            return Distance(m_previousCheckpoint, queryTicket) < 0;
```
Lectem, line 382: Do you want the latest value here, or the latest one you saw from the while loop?
Otherwise explicit memory ordering for the load please.

public/tracy/TracyD3D12.hpp
```diff
@@ -265,81 +298,241 @@ namespace tracy
...
+        }
 
-            // Recalibrate to account for drift.
-            RecalibrateClocks();
+        uint64_t RetireTicket(uint64_t ticket)
+        {
+            TracyD3D12Assert( m_previousCheckpoint == ticket );
+            ticket += 2;
+            m_previousCheckpoint.store(ticket);
+            return ticket;
```
Lectem, line 471: If not needed, might as well not return it.

Lectem (changes requested):

public/tracy/TracyD3D12.hpp
slomp: (I was experimenting with timers, so it was easier to just change it in one place; I'll take it out now.)

public/tracy/TracyD3D12.hpp
slomp: Yeah, `MapTimestampBuffer` will do some logging internally, but we should do something to avoid crashes.

public/tracy/TracyD3D12.hpp
slomp: No particular reason, just felt like a good enough cut-off waiting point.
In well-behaved software, `Drain()` should be able to complete quickly, way before the timeout.

public/tracy/TracyD3D12.hpp
slomp: Yeah, this one is checked constantly, so it makes sense to relax.

public/tracy/TracyD3D12.hpp
slomp: I'll also use a more relaxed memory order for the store here, as it does affect the NextQueryId() (which is called frequently from multiple threads potentially).

public/tracy/TracyD3D12.hpp
slomp: Latest value, because the loop could have been exited due to the timeout condition being reached.

slomp:
@Lectem I addressed most of your remarks.

For the atomic memory ordering, I relaxed the two "hottest" paths: `NextQueryId()` (which gets called concurrently by multiple-threads), and `RetireTicket()` (which only gets called by whichever thread managed to acquire the lock to run `Collect()`, but the atomic store is called repeatedly therein).

I could relax a bit the other paths (`Drain()` and `Wait()`) but they are only ever called during "panic" situations, which should never or very rarely happen in a well-behaved program. If they do happen, I'd say the client has much bigger fish to fry other than worrying about the impact of strong atomics in cache synchronization. And if we were to relax, we'd need acq_rel semantics, which is not too far apart in terms of overhead from the clean/default sec_cst ordering semantics.

public/tracy/TracyD3D12.hpp
slomp: Yes, the caller thread will "stall" until enough tickets are collected to proceed.
Drain() will then instruct Collect() to start dropping timestamps immediately to ensure progress (this is a "panic" situation).
That said, DropTimestamp() will attempt to resolve it first, and drop it if it fails.

Lectem: Mhhh, where does it stall exactly? `NextQueryId` just calls `Drain` which does not stall no? `Wait(,0)` will just return `false` immediatly no?
`while ((Distance(m_previousCheckpoint, queryTicket) >= 0) && (elapsed < timeout_ms))` will fail as `(elapsed < timeout_ms)` => `0<0`. And `Distance(m_previousCheckpoint, queryTicket) < 0` should be false since we are in the `Distance(m_previousCheckpoint, queryTicket) >= RingCapacity()` no?

public/tracy/TracyD3D12.hpp
```diff
@@ -265,86 +307,241 @@ namespace tracy
...
+            // dropping a query due to timeout, but said query is indeed submitted to the
+            // GPU queue for execution later on. This "late" query will eventually become
+            // resolved by the GPU asynchronously writting to the corresponding query slot.
+            // The newly produced query below could have matching ids/slot with said query.
+            // From there, there's no bullet-proof way for Collect() to distinguish between
+            // the timestamp written to the query slot belonging to the old/late query, or
+            // to the query that has just been generated. The value of the "late" query may
+            // may end up being collected as if it belonged to the the new query.
+            const uint64_t ticket = m_queryCounter.fetch_add(2, std::memory_order_relaxed);
+            const uint64_t checkpoint = m_previousCheckpoint.load(std::memory_order_relaxed);
```
Lectem, line 506: I think relaxed is wrong, especially if you used a `store(release)` ?
If you rely on what was done before the store, you need `acquire` here?

public/tracy/TracyD3D12.hpp
Lectem: Please use `m_previousCheckpoint.load(acquire)` if you need to see latest value and rely on the fact timestamp is now free

public/tracy/TracyD3D12.hpp
Lectem: Please use `m_previousCheckpoint.load(acquire)` if you need to see latest value and rely on the fact timestamp is now free

public/tracy/TracyD3D12.hpp
Lectem: explicit load, be it seq_cst or relaxed I'd rather see the fact it's an atomic op

public/tracy/TracyD3D12.hpp
```diff
@@ -33,45 +33,68 @@ using TracyD3D12Ctx = void*;
...
-        uint32_t m_queryIdStart = 0;
-        uint32_t m_queryCount = 0;
-    };
-
     // Command queue context.
     class D3D12QueueCtx
     {
         friend class D3D12ZoneScope;
 
+        uint8_t m_contextId = 255;  // 255 represents "invalid id"
```
Lectem, line 79: Could we use a constant here and in `CreateD3D12Context` ?

Lectem (changes requested):

public/tracy/TracyD3D12.hpp
slomp: When `NextQueryId` detects the ticket slot it claimed is still pending, it will request a drain of all tickets up to to the corresponding "old ticket" that maps to the same slot:
```
            uint64_t oldTicket = ticket - RingCapacity();
            Drain(oldTicket, 0);
```
`Drain()` will first attempt to `Wait()` for the pending slots (i.e., see if they can be resolved _immediately_, that is, without a timeout grace period).
This `Wait()` is more of a last polite attempt to resolve it.
If the wait fails (which is very likely), `Drain()` will forcibly collect timestamps up until the requested query id:
```
            // can't wait anymore: urgent collect request
            std::unique_lock lock (m_collectionMutex);
            Collect(lock, queryTicket, true);
```

public/tracy/TracyD3D12.hpp
slomp: Technically, yes, `acquire` would make things more "accurate".
In practice, using `relaxed` only leads to false-positives, which are fine (you end up reading a slightly older version of `m_previousCheckpoint`).

I opted for `relaxed` here (and accepting false positives) since `NextQueryId` is by far the hottest atomic contention path (it is called with high-frequency, and can be called concurrently).

In addition, unless the client is really running into pathological cases where queries are being generated but not submitted, the whole stall/drain logic would never trigger. I expect the vast majority of the clients to not have pathological cases, but if they do, the pathological case in the client itself is much more of a problem than a false-positive drain.

Lectem: OK I think this is indeed correct since it'll just drain if buffer is full!

public/tracy/TracyD3D12.hpp
Lectem: Oh right, we don't wait for this ticket but `ticket-RingCapacity() `, that's the part I missed. OK!

public/tracy/TracyD3D12.hpp
slomp: I opened a ticket, as this needs to be fixed across all GPU back-ends.
https://github.com/wolfpld/tracy/issues/1357

public/tracy/TracyD3D12.hpp
```diff
@@ -33,45 +33,68 @@ using TracyD3D12Ctx = void*;
...
+#else
+#define TracyD3D12Debug(...)
+#define TracyD3D12Break()
+#define TracyD3D12Assert(predicate, ...) assert(predicate);
+#endif
+
+#define TracyD3D12Log(severity, msg) tracy::Profiler::LogString( tracy::MessageSourceType::Tracy, tracy::MessageSeverity::severity, tracy::Color::Red4, 0, msg );
+#define TracyD3D12Panic(msg, ...) do { TracyD3D12Log(Error, msg); TracyD3D12Assert(false && "TracyD3D12: " msg); __VA_ARGS__; } while(false);
+
+#define TracyD3D12Relax() _mm_pause() // TracyYield()
```
Lectem: If unused, remove

Lectem (approved):
Other than the `TracyD3D12Relax` being unused, it seems like a good improvement overall.
Couldn't test it sadly (app I'm working on makes it too difficult to integrate the current scope based code.

Note that I found different ways of formatting code in this PR, but the code already did not follow the usual Tracy formatting rules.
For example
```c++
ZoneScopedC(Color::Red4);
TracyD3D12Assert( lock.owns_lock() );
```

As mentioned on discord, we might not need the <cstring> include, and could use `TracyFastVector` instead of `std::vector`, but `TracyVulkan.hpp` already set a precedent, so I suppose that's ok.

Ok for me, we'll just need to watch out for users that may encounter stalls due to this change.

slomp:
- switched to `FastVector`, removed `<vector>`
- removed `TracyD3D12Relax`, guarded `TracyD3D12Break`
- `<cstring>` is needed for `strlen` and `memcpy`
- will leave consistent formatting for a future PR dedicated to that

PR #1330, opened 2026-04-06
TracyCUDA: show GPU zones for CUDA Graph-launched kernels

bmilanich:
## Problem

When kernels are launched via CUDA Graphs (`cudaGraphLaunch` / `cuGraphLaunch`), Tracy shows 0 GPU zones. CUPTI delivers `CONCURRENT_KERNEL` and `MEMCPY` activity records but no per-kernel API callback fires, so `matchActivityToAPICall()` always fails and every zone is silently dropped.

## Root cause

Each activity record has a `correlationId` that Tracy matches against API callbacks stored in `cudaCallSiteInfo`. For CUDA Graph launches there are no individual-kernel callbacks, so the map has no entries for those correlation IDs.

## Key CUPTI finding (discovered during implementation)

**All kernels in one `cuGraphLaunch` call share the same `correlationId` as the launch itself.** This was confirmed by instrumenting both the API callback and activity buffer simultaneously:

```
cudaGraphLaunch ENTER corr=9
KERNEL graphId=2  corr=9   ← same as launch
KERNEL graphId=2  corr=9   ← same as launch
```

This means the fix is straightforward once `cuGraphLaunch` is tracked in the API callback machinery.

**Note:** `CUPTI_ACTIVITY_KIND_GRAPH_TRACE` was investigated but must **not** be enabled — it suppresses individual `CONCURRENT_KERNEL` records for graph-launched kernels, replacing them with graph-level summaries.

### Core graph correlation

- Add `cudaGraphLaunch_v10000` / `cuGraphLaunch` (both runtime and driver API variants) to the existing callback tracker maps so their CPU call site is captured in `cudaCallSiteInfo`
- Introduce `matchGraphActivityToAPICall()` helper (analogous to `matchActivityToAPICall()`):
  - For the **first** kernel/memcpy/memset from a graph launch, `matchActivityToAPICall` succeeds normally (consuming the launch entry) and the result is cached in `cudaGraphCurrentLaunch[graphId]`
  - **Subsequent** operations from the same launch (same `correlationId`, same `graphId`) find the cached `APICallInfo` via `graphId`
  - Each new launch of the same graph overwrites the cache entry via `insert_or_assign`, which acquires the write lock only once
  - If neither path succeeds, the caller emits a diagnostic `matchError()` in the profiler timeline (same behaviour as for non-graph activity)
- `using GraphID = uint32_t` typedef used throughout for graphId-typed variables

### Cache retirement (prevents unbounded growth)

- On `cudaGraphExecDestroy` (ENTER callback, while handle is still valid): call `cuptiGetGraphExecId` to translate exec handle → graphId and add to a pending-retirement set (`graphExecPendingRetire`)
- Erasure is deferred to `OnBufferCompleted` because `cudaGraphExecDestroy` doesn't wait for GPU completion — CUPTI may still have undelivered activity records in its buffers
- A graphId is erased only when a full buffer arrives containing no records bearing that graphId, indicating all in-flight records have been delivered
- `getGraphIdFromRecord()` helper extracts the graphId field from `CONCURRENT_KERNEL` / `MEMCPY` / `MEMSET` activity records for per-buffer tracking
- `std::atomic<bool> graphRetirePending` dirty flag lets `OnBufferCompleted` skip the retirement mutex on the hot path (no pending retirements = no lock acquired)

### ConcurrentHashMap fixes

- `ConcurrentHashMap`: added `insert_or_assign` wrapping `std::unordered_map::insert_or_assign` under a single write lock
- `ConcurrentHashMap::fetch()`: fixed a pre-existing missing read lock — `mapping.find()` was called without holding any lock, a data race now exercised on every graph activity record

### Memory tracking cleanup

- Remove `cudaMalloc`/`cudaFree`/`cuMemAlloc_v2`/`cuMemFree_v2` (6 CBIDs total) from `cbidRuntimeTrackers` / `cbidDriverTrackers` — the MEMORY2 handler never calls `matchActivityToAPICall` and never emits GPU zones, so these entries leaked indefinitely
- MEMORY2 handler: removed the early return on failed `matchActivityToAPICall` — `CUpti_ActivityMemory3` has no `graphId` field, so graph-launched alloc nodes (`cudaGraphAddMemAllocNode`) and pre-profiling allocations can never be correlated to an API call. The handler only needs address, size, and timestamp from the activity record

### Other

- `CUPTI_ACTIVITY_KIND_SYNCHRONIZATION` handler: confirmed correct as-is. In-graph event record/wait nodes produce no `SYNCHRONIZATION` activity records (empirically verified on H100)
- `examples/CUDAGraphRepro/Makefile`: added `-arch=native` so NVCC auto-detects the target GPU architecture

## Results

Tested on H100, CUDA 13.1, driver 580.105.08.

**Single graph (10 launches × 3 nodes):**

| Tracy version | GPU zones |
|---|---|
| Unpatched | 0 |
| This PR | 30 ✅ (20 `vector_add` + 10 `CUDA::memcpy`) |

**Multiple graphs, interleaved launches** — two distinct graphs launched alternately (A, B, A, B, …) to stress graphId cache switching:

- Graph A: `kernel(add)` + `memcpy` + `kernel(add)` — 3 nodes, 5 launches = 15 zones
- Graph B: `kernel(scale)` + `kernel(add)` + `kernel(scale)` — 3 nodes, 5 launches = 15 zones

| Zone type | Expected | Got |
|---|---|---|
| `vector_add` | 15 | 15 ✅ |
| `vector_scale` | 10 | 10 ✅ |
| `CUDA::memcpy` (graph) | 5 | 5 ✅ |
| Computation result `c[0]` | 6.0 | 6.0 ✅ |

**Regular (non-graph) CUDA operations** — verified that memory CBID tracker removal doesn't regress normal profiling:

| Metric | Expected | Got |
|---|---|---|
| GPU zones | 31 | 31 ✅ |
| Memory allocs | 36 | 36 ✅ |
| Memory frees | 36 | 36 ✅ |

GPU zones are correctly correlated to the `cuGraphLaunch` CPU call site: selecting a GPU zone highlights the corresponding CPU time range in the timeline.

Debug build of `tracy-capture` (asserts enabled) confirmed no assertions fire at `server/TracyWorker.cpp:5988`.

## Note on GPU zone source locations

All CUDA Graph GPU zones point to `TracyCUDA.hpp` rather than the user's call site. This is the same behaviour as regular CUDA zones — CUPTI delivers activity records asynchronously in a background thread that has no knowledge of the original call stack.

## Files changed

- `public/tracy/TracyCUDA.hpp` — core fix
- `examples/CUDAGraphRepro/repro.cu` — updated to use TracyCUDA headers and verify zones
- `examples/CUDAGraphRepro/Makefile` — release and debug variants; `-arch=native` for GPU arch detection

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/public/tracy/TracyCUDA.hpp
+++ b/public/tracy/TracyCUDA.hpp
@@ -1110,6 +1110,11 @@ namespace tracy
             {
                 ZoneNamedN(kernel, "tracy::CUDACtx::DoProcessDeviceEvent[malloc/free]", instrument);
                 CUpti_ActivityMemory3* memory3 = (CUpti_ActivityMemory3*)record;
+                // NOTE: CUpti_ActivityMemory3 has no graphId field, so matchGraphActivityToAPICall
+                // cannot be used here. Graph-launched memory alloc nodes (cudaGraphAddMemAllocNode)
+                // share the launch's correlationId and CUPTI emits multiple MEMORY2 records per node.
+                // The first record consumes the cudaCallSiteInfo entry; subsequent ones will fire a
+                // spurious matchError and skip memory tracking. This is a known limitation.
                 APICallInfo apiCall;
                 if (!matchActivityToAPICall(memory3->correlationId, apiCall)) {
                     return matchError(memory3->correlationId, "MEMORY");
--- a/public/tracy/TracyCUDA.hpp
+++ b/public/tracy/TracyCUDA.hpp
@@ -360,6 +360,11 @@ struct ConcurrentHashMap {
         auto lock = acquire_write_lock();
         return mapping.erase(key);
     }
+    auto insert_or_assign(TKey key, TValue value) {
+        ZoneNamed(insert_or_assign, instrument);
+        auto lock = acquire_write_lock();
+        return mapping.insert_or_assign(std::move(key), std::move(value));
+    }
 };
 
 #if TRACY_CUDA_ENABLE_CUDA_CALL_STATS
@@ -917,9 +922,7 @@ namespace tracy
                     return false;
                 }
             } else if (graphId != 0) {
-                auto& cache = PersistentState::Get().cudaGraphCurrentLaunch;
-                cache.erase(graphId);
-                cache.emplace(graphId, apiCallInfo);
+                PersistentState::Get().cudaGraphCurrentLaunch.insert_or_assign(graphId, apiCallInfo);
             }
             return true;
         }
--- a/public/tracy/TracyCUDA.hpp
+++ b/public/tracy/TracyCUDA.hpp
@@ -819,9 +819,10 @@ namespace tracy
                         { CUPTI_RUNTIME_TRACE_CBID_cudaLaunchKernel_ptsz_v7000,     GET_STREAM_FUNC(cudaLaunchKernel_ptsz_v7000_params, stream) },
                         { CUPTI_RUNTIME_TRACE_CBID_cudaLaunchKernelExC_v11060,      GET_STREAM_FUNC(cudaLaunchKernelExC_v11060_params, config->stream) },
                         { CUPTI_RUNTIME_TRACE_CBID_cudaLaunchKernelExC_ptsz_v11060, GET_STREAM_FUNC(cudaLaunchKernelExC_ptsz_v11060_params, config->stream) },
-                        // Runtime: Memory
-                        { CUPTI_RUNTIME_TRACE_CBID_cudaMalloc_v3020, NON_STREAM_FUNC() },
-                        { CUPTI_RUNTIME_TRACE_CBID_cudaFree_v3020,   NON_STREAM_FUNC() },
+                        // Runtime: Memory — NOT tracked here. The MEMORY2 handler
+                        // only needs address/size/timestamp from the activity record
+                        // and never calls EmitGpuZone, so there is no consumer for
+                        // the cudaCallSiteInfo entry. Tracking them would leak entries.
                         // Runtime: Memcpy
                         { CUPTI_RUNTIME_TRACE_CBID_cudaMemcpy_v3020,      NON_STREAM_FUNC() },
                         { CUPTI_RUNTIME_TRACE_CBID_cudaMemcpyAsync_v3020, GET_STREAM_FUNC(cudaMemcpyAsync_v3020_params, stream) },
@@ -856,11 +857,7 @@ namespace tracy
                         { CUPTI_DRIVER_TRACE_CBID_cuLaunchKernel_ptsz,   GET_STREAM_FUNC(cuLaunchKernel_ptsz_params, hStream)} ,
                         { CUPTI_DRIVER_TRACE_CBID_cuLaunchKernelEx,      GET_STREAM_FUNC(cuLaunchKernelEx_params, config->hStream) },
                         { CUPTI_DRIVER_TRACE_CBID_cuLaunchKernelEx_ptsz, GET_STREAM_FUNC(cuLaunchKernelEx_params, config->hStream) },
-                        // Driver: Memory
-                        { CUPTI_DRIVER_TRACE_CBID_cuMemAlloc_v2,         NON_STREAM_FUNC() },
-                        { CUPTI_DRIVER_TRACE_CBID_cuMemAllocManaged,     NON_STREAM_FUNC() },
-                        { CUPTI_DRIVER_TRACE_CBID_cuMemAllocPitch_v2,    NON_STREAM_FUNC() },
-                        { CUPTI_DRIVER_TRACE_CBID_cuMemFree_v2,          NON_STREAM_FUNC() },
+                        // Driver: Memory — NOT tracked (see Runtime: Memory comment above).
                         // Driver: Memcpy - Synchronous
                         { CUPTI_DRIVER_TRACE_CBID_cuMemcpy,              NON_STREAM_FUNC() },
                         { CUPTI_DRIVER_TRACE_CBID_cuMemcpyHtoD_v2,       NON_STREAM_FUNC() },
--- a/public/tracy/TracyCUDA.hpp
+++ b/public/tracy/TracyCUDA.hpp
@@ -786,6 +786,10 @@ namespace tracy
                         { CUPTI_RUNTIME_TRACE_CBID_cudaEventQuery_v3020,        NON_STREAM_FUNC() },
                         { CUPTI_RUNTIME_TRACE_CBID_cudaStreamWaitEvent_v3020,   NON_STREAM_FUNC() },
                         { CUPTI_RUNTIME_TRACE_CBID_cudaDeviceSynchronize_v3020, NON_STREAM_FUNC() },
+                        // Graph launch: tracked so we can correlate GPU activities back to
+                        // the cudaGraphLaunch call site via CUPTI_ACTIVITY_KIND_GRAPH_TRACE
+                        { CUPTI_RUNTIME_TRACE_CBID_cudaGraphLaunch_v10000,      GET_STREAM_FUNC(cudaGraphLaunch_v10000_params, stream) },
+                        { CUPTI_RUNTIME_TRACE_CBID_cudaGraphLaunch_ptsz_v10000, GET_STREAM_FUNC(cudaGraphLaunch_v10000_params, stream) },
                     };
                     #undef NON_STREAM_FUNC
                     #undef GET_STREAM_FUNC
--- a/public/tracy/TracyCUDA.hpp
+++ b/public/tracy/TracyCUDA.hpp
@@ -1110,13 +1110,13 @@ namespace tracy
             {
                 ZoneNamedN(kernel, "tracy::CUDACtx::DoProcessDeviceEvent[malloc/free]", instrument);
                 CUpti_ActivityMemory3* memory3 = (CUpti_ActivityMemory3*)record;
-                // Memory alloc/free records may not have a matching API call entry —
-                // e.g. allocations made before profiling started, or nodes inside a
-                // CUDA graph launch (CUpti_ActivityMemory3 has no graphId field).
-                // The handler only needs address, size, and timestamp from the activity
-                // record, so we proceed regardless of whether correlation succeeds.
-                APICallInfo apiCall;
-                matchActivityToAPICall(memory3->correlationId, apiCall);
+                // Do NOT call matchActivityToAPICall here. The handler only needs
+                // address, size, and timestamp — apiCall is never used. More importantly,
+                // consuming the cudaCallSiteInfo entry would break kernel/memcpy zone
+                // correlation for graphs that mix alloc nodes with other node types: all
+                // activities in one graph launch share the same correlationId, so consuming
+                // it here would prevent matchGraphActivityToAPICall from finding it for the
+                // subsequent kernel or memcpy records from the same launch.
                 static constexpr const char* graph_name = "CUDA Memory Allocation";
                 if (memory3->memoryOperationType == CUPTI_ACTIVITY_MEMORY_OPERATION_TYPE_ALLOCATION){
                     auto& memAllocAddress = PersistentState::Get().memAllocAddress;
--- a/public/tracy/TracyCUDA.hpp
+++ b/public/tracy/TracyCUDA.hpp
@@ -660,15 +660,44 @@ namespace tracy
             FlushActivityAsync();
         }
 
+        // Returns the graphId field from activity record kinds that carry one,
+        // or 0 for records that are not graph-launched or don't have the field.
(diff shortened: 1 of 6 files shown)


bmilanich:
Here is how it looks in the profiler now:
tracy with CUDA Graph zones

wolfpld:
Please review if these apply:

1. **`SYNCHRONIZATION` and `MEMORY2` handlers not updated** — These still call `matchActivityToAPICall` directly. If graph-launched sync events have a non-zero `graphId`, they'd be silently dropped the same way kernels were. Should these also use `matchGraphActivityToAPICall`, or is there a reason they don't need it? (Worth a comment if intentional.)
2. **Stale cache entries in `cudaGraphCurrentLaunch`** — No cleanup when a graph is destroyed. Since CUPTI reuses `graphId` values, a stale entry could theoretically be consumed by a later unrelated graph launch on the same `graphId`. The `insert_or_assign` overwrite handles same-graph relaunches, but a destroyed-then-recreated graph with a recycled `graphId` could inherit stale `APICallInfo`. Consider adding a note or defensive check.

bmilanich:
> Please review if these apply:
> 
> 1. **`SYNCHRONIZATION` and `MEMORY2` handlers not updated** — These still call `matchActivityToAPICall` directly. If graph-launched sync events have a non-zero `graphId`, they'd be silently dropped the same way kernels were. Should these also use `matchGraphActivityToAPICall`, or is there a reason they don't need it? (Worth a comment if intentional.)


So I had a separate patch for memory tracking with graphs, I can add it here just as well: there is no `graphId` in the `CUpti_ActivityMemory3` so the solution here is proceed with memory tracking regardless of matching with `matchActivityToAPICall()` (I have updated the PR with it).
Graphs do not produce SYNCHRONIZATION activity in CUPTI, their synchronization stays inside the GPU so no need to update `SYNCHRONIZATION`.

> 2. **Stale cache entries in `cudaGraphCurrentLaunch`** — No cleanup when a graph is destroyed. Since CUPTI reuses `graphId` values, a stale entry could theoretically be consumed by a later unrelated graph launch on the same `graphId`. The `insert_or_assign` overwrite handles same-graph relaunches, but a destroyed-then-recreated graph with a recycled `graphId` could inherit stale `APICallInfo`. Consider adding a note or defensive check.

It does not seem to be an issue as the new graph with the same `graphId` will still have a different `correlationId` and will succeed in the `matchGraphActivityToAPICall()`, replacing the stale cache entry.

slomp:
@bmilanich Thank you for the PR!
I will try to carve some time next week to review/test it properly, but from a first glance, I like the solution.

Quick hypothetical question:
What happens if one calls `cuGraphLaunch()` while there's another `cuGraphLaunch()` on the same `graphId` still in-flight? Presumably, CUDA should give the second call a different `correlationId` (as it would for other API calls), but I could not find definitive documentation on that.

bmilanich:
> @bmilanich Thank you for the PR! I will try to carve some time next week to review/test it properly, but from a first glance, I like the solution.
> 
> Quick hypothetical question: What happens if one calls `cuGraphLaunch()` while there's another `cuGraphLaunch()` on the same `graphId` still in-flight? Presumably, CUDA should give the second call a different `correlationId` (as it would for other API calls), but I could not find definitive documentation on that.

This is a good point: if the same graphId and different corr id records will arrive interleaved it will mess up the cache. I need to make a good repro for this and figure how to handle that.

bmilanich:
So it seems to be impossible to produce interleaved records from the same graph handle launches. This is actually documented here: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__GRAPH.html#group__CUDART__GRAPH_1g1920584881db959c8c74130d79019b73

```
Description
Executes graphExec in stream. Only one instance of graphExec may be executing at a time. Each launch is ordered behind both any previous work in stream and any previous launches of graphExec. To execute a graph concurrently, it must be instantiated multiple times into multiple executable graphs.
```

slomp:
> So it seems to be impossible to produce interleaved records from the same graph handle launches. This is actually documented here: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__GRAPH.html#group__CUDART__GRAPH_1g1920584881db959c8c74130d79019b73
> 
> ```
> Description
> Executes graphExec in stream. Only one instance of graphExec may be executing at a time. Each launch is ordered behind both any previous work in stream and any previous launches of graphExec. To execute a graph concurrently, it must be instantiated multiple times into multiple executable graphs.
> ```

Great, thanks for digging!

public/tracy/TracyCUDA.hpp
```diff
@@ -895,6 +909,26 @@ namespace tracy
...
+        static bool matchGraphActivityToAPICall(uint32_t correlationId, uint32_t graphId,
+                                                APICallInfo& apiCallInfo, const char* kind) {
+            auto& graphLaunchCache = PersistentState::Get().cudaGraphCurrentLaunch;
+            if (!matchActivityToAPICall(correlationId, apiCallInfo)) {
+                if (graphId == 0 || !graphLaunchCache.fetch(graphId, apiCallInfo)) {
+                    matchError(correlationId, kind);
+                    return false;
+                }
+            } else if (graphId != 0) {
+                graphLaunchCache.insert_or_assign(graphId, apiCallInfo);
```
slomp, line 998: Is `insert_and_assign` really necessary (as opposed to just using operator[])?

public/tracy/TracyCUDA.hpp
```diff
@@ -337,6 +337,7 @@ struct ConcurrentHashMap {
     }
     auto fetch(TKey key, TValue& value) {
         ZoneNamed(fetch, instrument);
+        auto lock = acquire_read_lock();
```
slomp, line 340: oof! good catch!

public/tracy/TracyCUDA.hpp
```diff
@@ -895,6 +909,26 @@ namespace tracy
...
+        // the first activity consumes the cudaCallSiteInfo entry and caches it by
+        // graphId; subsequent activities from the same launch find it via graphId.
+        // Returns false (and emits a Tracy error message) only if no match is found
+        // via either path.
+        static bool matchGraphActivityToAPICall(uint32_t correlationId, uint32_t graphId,
+                                                APICallInfo& apiCallInfo, const char* kind) {
+            auto& graphLaunchCache = PersistentState::Get().cudaGraphCurrentLaunch;
+            if (!matchActivityToAPICall(correlationId, apiCallInfo)) {
+                if (graphId == 0 || !graphLaunchCache.fetch(graphId, apiCallInfo)) {
+                    matchError(correlationId, kind);
```
slomp: For debugging purposes, it's best to keep the call to `matchError` at the caller site, after it has checked the result value of the callee (the caller context can give more context)

public/tracy/TracyCUDA.hpp
```diff
@@ -997,8 +1031,8 @@ namespace tracy
                 ZoneNamedN(kernel, "tracy::CUDACtx::DoProcessDeviceEvent[kernel]", instrument);
                 CUpti_ActivityKernel9* kernel9 = (CUpti_ActivityKernel9*) record;
                 APICallInfo apiCall;
-                if (!matchActivityToAPICall(kernel9->correlationId, apiCall)) {
-                    return matchError(kernel9->correlationId, "KERNEL");
+                if (!matchGraphActivityToAPICall(kernel9->correlationId, kernel9->graphId, apiCall, "KERNEL")) {
+                    return;
```
slomp: keep matchError here, remove it from matchGraphActivityToAPICall

public/tracy/TracyCUDA.hpp
```diff
@@ -1076,10 +1110,13 @@ namespace tracy
             {
                 ZoneNamedN(kernel, "tracy::CUDACtx::DoProcessDeviceEvent[malloc/free]", instrument);
                 CUpti_ActivityMemory3* memory3 = (CUpti_ActivityMemory3*)record;
-                APICallInfo apiCall;
-                if (!matchActivityToAPICall(memory3->correlationId, apiCall)) {
-                    return matchError(memory3->correlationId, "MEMORY");
-                }
+                // Do NOT call matchActivityToAPICall here. The handler only needs
```
slomp: From the comments here, it's unclear what happens to the (correlationId, API call) entry if it is never removed from the map via `matchActivityToAPICall`.

public/tracy/TracyCUDA.hpp
```diff
@@ -1264,6 +1305,7 @@ namespace tracy
             // NOTE(marcos): these objects do not need to persist, but their relative
             // footprint is trivial enough that we don't care if we let them leak
             ConcurrentHashMap<CorrelationID, APICallInfo> cudaCallSiteInfo;
+            ConcurrentHashMap<uint32_t, APICallInfo> cudaGraphCurrentLaunch;
```
slomp: Add a typedef for graph ID:
```
 using GraphID = uint32_t;
```

slomp (changes requested):
Doing a first pass of the code here.
I'll give it a run on some workloads next week.

public/tracy/TracyCUDA.hpp
bmilanich: You are right, even though it's never used, there is still an allocated cudaCallSiteInfo entry. I thought it's going to be freed by the RELEASE handling but it really is leaking. I think not allocating CPU activity entries for memory allocations would be the solution here. Also, there is a similar issue with the graph calls cache: it's never freed, so I need to handle graphs retiring and clean it.

public/tracy/TracyCUDA.hpp
bmilanich: operator[] grabs the read lock and then returns the plain reference, which, if assigned, will go without any locking. insert_or_assign() grabs the write lock and, if necessary, updates under the lock so it's safe.

slomp: Sorry, I meant to say `emplace`, not `operator[]`.

If I am following the logic correctly, whatever is placed in `graphLaunchCache` stays there until it gets overwritten, so `emplace` would just keep remembering the very first insertion, whereas here you want to enforce that it gets rewritten, since it could be a different launch of the same graph.

(I wonder if there's a CPUTI activity record that tells us when a graph launch has started and when it has ended.)

slomp: It looks like there's this:
https://docs.nvidia.com/cupti/api/structCUpti__ActivityGraphTrace2.html

It will be posted in the activity record queue after all graph nodes have executed, so you could use that to "cleanup" the entry from the map, I suppose.

public/tracy/TracyCUDA.hpp
slomp: For the graph calls cache, [see my comment here](https://github.com/wolfpld/tracy/pull/1330#discussion_r3082378127).

For memory, yeah, agree, we can skip them.
(The cuMalloc/cuFree calls will still show up in the trace, but the activity matching is less important for this case.)

public/tracy/TracyCUDA.hpp
bmilanich: > Sorry, I meant to say `emplace`, not `operator[]`.
> 
> If I am following the logic correctly, whatever is placed in `graphLaunchCache` stays there until it gets overwritten, so `emplace` would just keep remembering the very first insertion, whereas here you want to enforce that it gets rewritten, since it could be a different launch of the same graph.
> 

Yes, but the graphLaunchCache is indexed by the graphId, which remains the same between different cuGraphLaunch() executions, so the CPU correlation would point to the very first launch of this graph.

> (I wonder if there's a CPUTI activity record that tells us when a graph launch has started and when it has ended.)

There is and I explored it but it's only tracing the whole graph execution and suppresses individual kernel activities as stated in NV's doc:
```
This activity record represents execution for a graph without giving visibility about the execution of its nodes. This is intended to reduce overheads in tracing each node. The activity kind is CUPTI_ACTIVITY_KIND_GRAPH_TRACE
```
Thus the comments at https://github.com/wolfpld/tracy/pull/1330/changes#diff-f80a13989fdff31b7a1416dbe2081b9aa8d022005f2b1a2b8dc859be3c9f33e7R1244

slomp: Hmmm, yeah, what a bizarre choice (or overlook) from NVIDIA...

slomp:
@wolfpld Are you OK with a new `example` being added with this PR?
Other than that, LGTM

slomp (approved):

wolfpld:
> Are you OK with a new `example` being added with this PR?

Well, these are not really examples, but rather repro cases.

I don't really like the specific wording around them, or that these are AI generated.

On the other hand, I don't really see a viable way to provide these repro cases in some other way that would be convenient. And the actual programs do show where the problems are.

slomp:
@wolfpld I'll add to my bucket list a task to consolidate this CUDA `example` with other CUDA code I use here for testing, and release it as a `test` instead.

PR #1336, opened 2026-04-16
TracyRocprof: fix on-demand profiling crash and missing context name

bmilanich:
## Problem

The rocprofiler GPU backend crashes when a profiler connects to an application built with `TRACY_ON_DEMAND`:

    Assertion `ctx' failed in ProcessGpuZoneBeginImplCommon

Even if the crash is worked around, the GPU context appears unnamed and kernel names are missing.

## Root cause

`gpu_context_allocate()` writes `GpuNewContext` and `GpuContextName` queue items but never calls `DeferItem()` for either. Under on-demand mode, a late-connecting client never receives these messages, so it has no GPU context when GPU zone events start arriving.

Separately, `tool_callback_tracing_callback()` gates all callbacks on `data->init`, which is only set after the calibration thread allocates the GPU context. Kernel symbol registrations (`CODE_OBJECT_DEVICE_KERNEL_SYMBOL_REGISTER`) happen at HIP init time, well before `data->init` is set, so they are silently dropped. This was a regression introduced in 86de3970 ("Add calibration thread") — the earlier `delay_init()` approach in 98047ffb had the guard placed after the code_object block, so symbols were always recorded.

## Fix

- Add `DeferItem()` calls for both `GpuNewContext` and `GpuContextName` under `#ifdef TRACY_ON_DEMAND`, replicating the pattern already used by the CUDA backend (`TracyCUDA.hpp` `SubmitQueueItem`).
- Move the `data->init` guard after the code object registration block, restoring the pre-86de3970 behavior so kernel symbols are always recorded.

## Repro case

`tests/rocprof/repro/on_demand/` contains a minimal HIP program and a `check_gpu_ctx_name` tool. See the README in that directory for details.

## Test results

Tested on AMD MI300X (gfx950), ROCm 7.1.1, both release and debug builds:

| Build | Unpatched | Patched |
|---|---|---|
| Release (`-O2`) | `tracy-capture` segfaults | Capture succeeds, ~50 GPU zones |
| Debug (`-g -O0`) | `Assertion 'ctx' failed` in `ProcessGpuZoneBeginImplCommon` | No assertions, ~50 GPU zones |
| Context name | N/A (crash) | `rocprofv3` |
| Kernel names | N/A (crash) | Resolved (`vectorAdd`) |


wolfpld:
There are conflicts to be resolved here. Otherwise ready for merge.

PR #1339, opened 2026-04-17
Add replay controls to frame and zone statistics

JunKnows:
Add replay support to Frame statistics and Find zone so users can watch histogram and summary statistics build up frame by frame instead of only seeing the final aggregated result.

- add replay, pause/continue, exit, and speed controls
- respect Limit to view / Limit range when choosing replay bounds
- keep histogram axes fixed to the full replay range during playback
- avoid initial empty-state flashing when replay starts
- align the replay toolbar behavior across both statistics panels
20260418003658_rec_

### DIFF
--- a/profiler/src/profiler/TracyImGui.hpp
+++ b/profiler/src/profiler/TracyImGui.hpp
@@ -5,6 +5,7 @@
 #  pragma warning( disable: 4244 )  // conversion from don't care to whatever, possible loss of data
 #endif
 
+#include <array>
 #include <math.h>
 #include <stdint.h>
 #include <vector>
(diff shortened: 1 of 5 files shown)


slomp:
What is the rationale behind this feature?
(You can get similar replay visualization by limiting the view, and manually dragging the view limits to the left/right).

JunKnows:
> What is the rationale behind this feature? (You can get similar replay visualization by limiting the view, and manually dragging the view limits to the left/right).

Yes, for frame statistics you can approximate something similar by limiting the view and manually dragging it, but that is awkward in practice: you need to zoom to the right scale first, drag at a consistent speed, and repeat the same manual process every time. Replay gives a one-click, controlled progression instead of a mouse-driven one.

For Find zone there is no comparable drag-based workflow, so the feature is more valuable there.

It also makes the buildup easier to read because the histogram axes stay fixed and the playback speed is explicit.

I’ve  opened an issue to explain this. This is my issue：https://github.com/wolfpld/tracy/issues/1340

PR #1342, opened 2026-04-20
Fix #1337 : On Linux respect max sample rate

### LINKED ISSUE #1337: Sampling frequency should be clamped to system's limit
For some reason on my machine, `/proc/sys/kernel/perf_event_max_sample_rate` drops down to 9000 after a while, and this causes perf_event_open() to return EINVAL since Tracy will request 10k sampling rate by default.

This can be easily workaround by setting `TRACY_SAMPLING_HZ`, but it'd be nice for the default range be automatically clamped, since the sampler appears to fail silently unless TRACY_VERBOSE is also enabled.

Lectem:
On Linux, Tracy was ignoring the system's maximum perf sampling rate, which could cause dropped events — particularly on low-performance or embedded systems.

This PR reads `/proc/sys/kernel/perf_event_max_sample_rate` at startup and caps the requested frequency to the system maximum if needed. A debug message is emitted when the rate is reduced.

The refactor of `GetSamplingPeriod` → `SamplingFrequencyToPeriodNs` is included to ensure the reported `samplingPeriod` always reflects the actual (possibly capped) frequency used.

Note: `/proc/sys/kernel/perf_event_max_sample_rate`'s value is non-zero since https://github.com/torvalds/linux/commit/723478c8a471403c53cf144999701f6e0c4bbd11 but I could had a null-check if needed.

### DIFF
--- a/public/client/TracySysTrace.cpp
+++ b/public/client/TracySysTrace.cpp
@@ -34,9 +34,9 @@ static int GetSamplingFrequency()
 #endif
 }
 
-static int GetSamplingPeriod()
+static int SamplingFrequencyToPeriodNs(  int samplingHz )
 {
-    return 1000000000 / GetSamplingFrequency();
+    return 1000000000 / samplingHz;
 }
 
 }
@@ -174,11 +174,6 @@ void WINAPI EventRecordCallback( PEVENT_RECORD record )
     }
 }
 
-static int GetSamplingInterval()
-{
-    return GetSamplingPeriod() / 100;
-}
-
 static etw::Session session_kernel = {};
 static etw::Session session_vsync = {};
 static PROCESSTRACE_HANDLE consumer_kernel = INVALID_PROCESSTRACE_HANDLE;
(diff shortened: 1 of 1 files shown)


public/client/TracySysTrace.cpp
```diff
@@ -34,9 +34,9 @@ static int GetSamplingFrequency()
 #endif
 }
 
-static int GetSamplingPeriod()
+static int SamplingFrequencyToPeriodNs(  int samplingHz )
```
wolfpld: Space.

public/client/TracySysTrace.cpp
```diff
@@ -207,10 +202,10 @@ bool SysTraceStart( int64_t& samplingPeriod )
 
 
 #ifndef TRACY_NO_SAMPLING
-    int microseconds = GetSamplingInterval() / 10;
+    samplingPeriod = SamplingFrequencyToPeriodNs( GetSamplingFrequency() );
+    const int microseconds = samplingPeriod / 1000;
     if( etw::EnableCPUProfiling( session_kernel, microseconds ) != ERROR_SUCCESS )
```
wolfpld, line 207: etw?

public/client/TracySysTrace.cpp
```diff
@@ -659,7 +654,22 @@ bool SysTraceStart( int64_t& samplingPeriod )
     const bool noVsync = noVsyncEnv && noVsyncEnv[0] == '1';
 #endif
 
-    samplingPeriod = GetSamplingPeriod();
+    int samplingFrequency = GetSamplingFrequency();
+    if( samplingFrequency > 0 )
+    {
+        FILE* f = fopen( "/proc/sys/kernel/perf_event_max_sample_rate", "r" );
```
wolfpld: Use `ReadFile` + `atoi`, like rest of the code.

public/client/TracySysTrace.cpp
Lectem: Yeah it was also using `GetSamplingPeriod()` that I replaced, and `GetSamplingInterval` was adding an indirection for nothing (dividing by 100, then divided by 10 at call site).

Alternative (not touching `GetSamplingInterval`) would have been to introduce yet another function or write the `1000000000 / samplingHz` inline in the linux code.

Your call

PR #1343, opened 2026-04-20
Better tracefs detection

Lectem:
On older kernels and some systems (e.g. Android, embedded), `tracefs` is not mounted directly but is accessible under `debugfs` at `<debugfs>/tracing`. Users may also have been instructed to use ` sudo mount -t debugfs debugfs /sys/kernel/debug`. Tracy would fail to find it and silently skip system tracing.

- Switch tracefs detection from `mnt_fsname` to `mnt_type`, consistent with `libtracefs`. A `tracefs` mount with a custom source name (e.g. `mount -t tracefs nodev /path`) would previously be missed.
- When no direct `tracefs` mount is found, fall back to `<debugfs>/tracing` if debugfs is present. Continues iterating `/proc/mounts` after a debugfs hit so a later tracefs entry still takes priority.
- Also logs the reason when `SysTraceStart` fails to help users diagnose missing systrace information.

### DIFF
--- a/public/client/TracySysTrace.cpp
+++ b/public/client/TracySysTrace.cpp
@@ -563,7 +563,7 @@ static char* GetTraceFsPath()
     char* ret = nullptr;
     while( auto ent = getmntent( f ) )
     {
-        if( strcmp( ent->mnt_fsname, "tracefs" ) == 0 )
+        if( strcmp( ent->mnt_type, "tracefs" ) == 0 )
         {
             auto len = strlen( ent->mnt_dir );
             ret = (char*)tracy_malloc( len + 1 );
--- a/public/client/TracySysTrace.cpp
+++ b/public/client/TracySysTrace.cpp
@@ -572,13 +572,12 @@ static char* GetTraceFsPath()
             ret[len] = '\0';
             break;
         }
-        else if( strcmp( ent->mnt_type, "debugfs" ) == 0 )
+        else if( !ret && strcmp( ent->mnt_type, "debugfs" ) == 0 )
         {
             const char* tracingDirName = "tracing";
             const size_t tracingDirNameLen = strlen( tracingDirName );
             auto debugFsPathLen = strlen( ent->mnt_dir );
-            // Typically there should be at most one debugfs, entry, but still possible to have multiple? hence realloc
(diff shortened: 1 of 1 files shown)


public/client/TracySysTrace.cpp
```diff
@@ -583,15 +597,19 @@ bool SysTraceStart( int64_t& samplingPeriod )
...
     const auto paranoidLevelStr = ReadFile( "/proc/sys/kernel/perf_event_paranoid" );
-    if( !paranoidLevelStr ) return false;
-#ifdef TRACY_VERBOSE
-    int paranoidLevel = 2;
-    paranoidLevel = atoi( paranoidLevelStr );
+    if( !paranoidLevelStr ) {
+        TracyDebug( "Failed to read perf_event_paranoid, cannot setup system tracing." );
+        return false;
+    }
+
```
wolfpld, line 604: Whitespace.

public/client/TracySysTrace.cpp
```diff
@@ -583,15 +597,19 @@ bool SysTraceStart( int64_t& samplingPeriod )
 #endif
 
     const auto paranoidLevelStr = ReadFile( "/proc/sys/kernel/perf_event_paranoid" );
-    if( !paranoidLevelStr ) return false;
-#ifdef TRACY_VERBOSE
-    int paranoidLevel = 2;
-    paranoidLevel = atoi( paranoidLevelStr );
+    if( !paranoidLevelStr ) {
```
wolfpld, line 600: `{` on separate line.

public/client/TracySysTrace.cpp
```diff
@@ -563,14 +563,28 @@ static char* GetTraceFsPath()
...
         {
             auto len = strlen( ent->mnt_dir );
-            ret = (char*)tracy_malloc( len + 1 );
+            // ret may be != nullptr if we already saw a debugfs entry
+            ret = (char*)tracy_realloc( ret, len + 1 );
             memcpy( ret, ent->mnt_dir, len );
             ret[len] = '\0';
             break;
         }
+        else if( strcmp( ent->mnt_type, "debugfs" ) == 0 )
```
wolfpld, line 575: Only enter if `!ret`.

PR #1346, opened 2026-04-22
Refactoring debug symbol resolving process on Windows

slomp:
Present:
- adding wrappers around the DbgHelp API to intercept/report errors
- tracking the "symbol database type" of each module to cull unnecessary work
- scoped lock for `DBGHELP_LOCK/UNLOCK`
- adding `TRACY_SYMBOL_PATH` to `_NT_SYMBOL_PATH` for `SymSrv.dll`
- decoupling "outer" frame resolve logic from inlined frame resolve logic

Planned for a future PR:
- resolve symbols via DIA (`SymGetDiaSession()`) whenever possible
- request "bulk" code symbol resolving from `cs_disasm()` loop


public/client/windows/TracyDbgHelp.cpp
```diff
@@ -0,0 +1,333 @@
...
+static DWORD64 TracySymLoadModuleEx( HANDLE hProcess, HANDLE hFile, PCSTR ImageName, PCSTR ModuleName, DWORD64 BaseOfDll, DWORD DllSize, PMODLOAD_DATA Data, DWORD Flags ) {
+    DBGHELP_SCOPED_LOCK;
+    DWORD64 BaseAddress = SymLoadModuleEx( hProcess, hFile, ImageName, ModuleName, BaseOfDll, DllSize, Data, Flags );
+#if DBGHELP_DEBUG_LEVEL >= 1
+    if( BaseAddress == 0 ) {
+        DWORD code = GetLastError();
+        if( code != ERROR_SUCCESS )
+        {
+            char msg [512] = {};
+            const char* modName = ModuleName ? ModuleName : (ImageName ? ImageName : "[NULL]");
```
Lectem, line 114: I would use ImageName first for logging. Even though ModuleName is the "Friendly name", it is sometimes necessary to differentiate two dlls named the same way by their path (yes, it's possible!).

public/client/windows/TracyDbgHelp.cpp
```diff
@@ -0,0 +1,333 @@
...
+    return status;
+}
+
+static DWORD TracySymAddrIncludeInlineTrace( HANDLE hProcess, DWORD64 Address )
+{
+#ifdef TRACY_NO_CALLSTACK_INLINES
+    return 0;
+#endif
+    using SymAddrIncludeInlineTraceProc = decltype(TracySymAddrIncludeInlineTrace)(__stdcall*);
+    static auto _SymAddrIncludeInlineTrace = (SymAddrIncludeInlineTraceProc)GetProcAddress(GetModuleHandleA("dbghelp.dll"), "SymAddrIncludeInlineTrace");
```
Lectem, line 138: I'm not sure I like having deferred loading of dll procs each in their own function. I'd rather keep this in the init function  if only so that we can log all loading failures at the same place, and know the list of things we depend on in one place.

public/client/windows/TracyDbgHelp.cpp
```diff
@@ -0,0 +1,333 @@
...
+    info.SizeOfStruct = sizeof(IMAGEHLP_MODULE64);
+    if( TracySymGetModuleInfo64( GetCurrentProcess(), baseOfDll, &info ) == FALSE ) return SymNone;
+    return info.SymType;
+}
+
+struct DbgHelpSymbol
+{
+    static constexpr size_t MaxNameSize = 8*1024;
+    SYMBOL_INFO info;
+    char buf [MaxNameSize];    // must follow SYMBOL_INFO
```
Lectem, line 262: Should we pack to avoid padding here?

public/client/TracyCallstack.cpp
```diff
@@ -122,6 +93,18 @@ extern "C" const char* ___tracy_demangle( const char* mangled )
...
 {
 
+struct ImageEntry
+{
+    uint64_t m_startAddress = 0;
+    uint64_t m_endAddress = 0;
+    char* m_name = nullptr;
+    char* m_path = nullptr;
+
+#if TRACY_HAS_CALLSTACK == 1
```
Lectem, line 103: Note: I'll probably turn this into a BuildInfo (buildid/pdb info, etc) member later

public/client/TracyCallstack.cpp
```diff
@@ -898,19 +782,22 @@ struct ModuleNameAndBaseAddress
 {
     const char* name;
     uint64_t baseAddr;
+#if TRACY_HAS_CALLSTACK == 1
```
Lectem, line 785: Not sure about this one (I'm on mobile right now), but isn't this Windows only already?

public/client/TracyCallstack.cpp
```diff
@@ -933,187 +820,149 @@ ModuleNameAndBaseAddress GetModuleNameAndPrepareSymbols( uint64_t addr )
...
-    cb_data[0].file = CopyStringFast( "[unknown]" );
-    cb_data[0].line = 0;
+static int ResolveInlines( uint64_t ptr, const ModuleNameAndBaseAddress& moduleInfo )
+{
+    // only SymPdb and SymDia are capable of carring inline information
+    if ( moduleInfo.symType != SymPdb && moduleInfo.symType != SymDia ) return 0;
 
-    return { cb_data, 1, moduleNameAndBaseAddress.name };
+    DbgHelpInline inln;
+    int inlineNum = DbgHelpTraceInline( inln, ptr );
```
Lectem, line 880: It bothers me a bit that the lock is now inside the wrappers, it means we're going to lock/unlock it for each call where before we would just lock unlock once.

public/client/TracyCallstack.hpp
```diff
@@ -7,19 +7,6 @@
 #include "../common/TracyForceInline.hpp"
 #include "TracyCallstack.h"
 
-namespace tracy
-{
-
-struct ImageEntry
```
Lectem: Note: I'll probably move it back here when adding build info, cause I'll need it to be part of the interface if we want offline resolution.

Lectem (changes requested):

PR #1347, opened 2026-04-24
Add MCP server for AI-assisted trace analysis

alandtse:
Introduce an optional Model Context Protocol (MCP) server that lets AI assistants analyze Tracy captures and live sessions through Tracy's own server engine. The server runs as a Python sidecar and talks to the existing C++ analysis code through new pybind11 bindings.

- python/bindings/ServerModule.cpp: TracyServerBindings module exposing Worker, file I/O, zones, GPU zones, frame data, plots, messages, locks, source locations, and summary statistics (zone/GPU child stats, frame timing, etc.).
- python/CMakeLists.txt: builds and installs TracyServerBindings alongside TracyClientBindings.
- extra/mcp/tracy_mcp.py: FastMCP SSE singleton with dynamic port discovery, PID-file based singleton detection, session-isolated worker instances, synchronous and background eval, task polling, and a shutdown tool to release the .pyd lock during development.
- extra/mcp/start_mcp.sh, .gitignore: launcher with local override hook; ignores generated port/pid files.
- manual/tracy.md: documents building, running, and integrating the server with an AI assistant.


alandtse:
Please note, I drafted this with AI but I have tested it extensively. Happy to make further edits.

alandtse:
As a follow-up PR, I also enable Tracy Assist to access the MCP infrastructure internally so it can query it too.

wolfpld:
> As a follow-up PR, I also enable Tracy Assist to access the MCP infrastructure internally so it can query it too.

The way I see things, Tracy Assist already has (some) tools it can use, and these should be exposed as MCP. The natural evolution would be to add more capabilities, e.g. access to function statistics and so on.

In contrast, your solution is basically an "outside view" on the data. Can you comment why you chose this approach?

alandtse:
> > As a follow-up PR, I also enable Tracy Assist to access the MCP infrastructure internally so it can query it too.
> 
> The way I see things, Tracy Assist already has (some) tools it can use, and these should be exposed as MCP. The natural evolution would be to add more capabilities, e.g. access to function statistics and so on.
> 
> In contrast, your solution is basically an "outside view" on the data. Can you comment why you chose this approach?

I'll answer your specific ask about the follow-on PR then go to your general question.

First, I didn't realize Tracy Assist had any actual tools or functionality. I may have a had a skewed view as I was using the tracy 1.13.1 release and Tracy Assist did not appear to have any visibility of the active trace. I tested with Ollama and each LLM said it didn't see anything available so I assumed nothing was activated. I had not tried dev. Once I determined Tracy Assist had limited functionality for me, I went ahead and created this PR to expose the MCP for my external Claude code workflow. As an afterthought, I realized Tracy Assist should have some tools, and instead of writing a new infrastructure for Tracy Assist figured it'd be simpler to expose the MCP directly to it.

If you're asking me more generally about the approach for the MCP server to leverage the existing cpp bindings instead of exposing more tools, it's based on my experience with https://github.com/EdenLabs/agentic-renderdoc#why-this-design which uses a different approach. It tries to use fewer semantically grounded tools vs a large suite of tools. It tries to take advantage of the fact LLMs can already code, so just exposes a coding interface with some hints on usage.

I had previously done a tools approach working with https://github.com/symgraph/GhidrAssistMCP (and other forks) and it felt like I was reinventing the wheel by exposing more and more tools. The ecosystem is littered with forks with differing sets of tools and given the complexity of Ghidra, it just seemed like a huge scaling problem. And I also noticed that the LLM only really used a couple of the exposed tools so many of the added tools seemed wasteful. So it was mainly about the scalability and maintenance burden and wanting to avoid the pattern of 1) create a cpp API and 2) then create a MCP tool to access that api.

alandtse:
I noted I was behind and just rebased onto current upstream and pushed a couple of follow-up commits to address some issues I noticed with live connect and the bootstrap problem.

Rather than re-implementing Tracy Assist's tools through MCP, I'm now sharing your system.prompt.md as an MCP resource plus a small bindings-layer guide. I understand you may want the curated tools approach. I can adjust this server if needed to follow that architecture, but currently leaving in the `eval` approach in case that seems interesting to you. 

If you'd prefer the curated direction, happy to follow up by factoring the TracyLlmTools.cpp handlers into a shared C++ component that both Tracy Assist and the MCP can call — that way new tools you add show up on both surfaces from a single implementation, and eval can stay as an escape hatch for one-off queries particularly for frontier llms.

For Tracy Assist, I don't have to add the follow-up PR since it does make sense to have the built in tool be more curated generally.

wolfpld:
> I may have a had a skewed view as I was using the tracy 1.13.1 release and Tracy Assist did not appear to have any visibility of the active trace. I tested with Ollama and each LLM said it didn't see anything available so I assumed nothing was activated.

Yes, the current functionality is rather limited. User manual lists what's available – basically user manual reference and source code exploration. These work very well, with the attachments (call stacks, etc.) manually added by the user as a starting point.

The main problem in adding new tools is in how the large amounts of data can be provided to the LLM in a concise, context-friendly format. A good intertwined design is needed here, with the end goal of being able to just query "how to make this program faster", and getting an actionable (and correct) solution.

> If you're asking me more generally about the approach for the MCP server to leverage the existing cpp bindings instead of exposing more tools

The current implementation in Tracy focuses on doing things *with* a trace. The functionality you're exposing allows loading or listing traces on disk, presumably also starting or stopping a capture – that's a vastly different scope. So I guess this is a question of what is your use case.

I can see how this could be useful, with an agent trying different approaches to a problem and profiling each one, but I'd like to keep things separated, both at conceptual and review stages.

> It tries to take advantage of the fact LLMs can already code, so just exposes a coding interface with some hints on usage.

The problem here is that this requires a Python interpreter available, which is not something Tracy (the GUI application) integrates, or even considered doing so. It may well be that exposing a programming surface to the LLM is a fine solution here, but that means that the same thing should be somehow available to the user, and that requires layers of additional design.

> If you'd prefer the curated direction, happy to follow up by factoring the TracyLlmTools.cpp handlers into a shared C++ component that both Tracy Assist and the MCP can call — that way new tools you add show up on both surfaces from a single implementation, and eval can stay as an escape hatch for one-off queries particularly for frontier llms.

I think this PR can serve as a stop-gap solution for anyone wanting the functionality right now.

alandtse:
> The main problem in adding new tools is in how the large amounts of data can be provided to the LLM in a concise, context-friendly format. A good intertwined design is needed here, with the end goal of being able to just query "how to make this program faster", and getting an actionable (and correct) solution.
> 

That's a good point. In reality, I'm deferring to the LLM to query the data and parse it. I don't really know how much context it's getting from tracy but it is responding with reasonable data about performance tied to our use of tracy zones.

> The current implementation in Tracy focuses on doing things _with_ a trace. The functionality you're exposing allows loading or listing traces on disk, presumably also starting or stopping a capture – that's a vastly different scope. So I guess this is a question of what is your use case.

I'm mainly using it while performance optimizing. We are using Claude Code for example, to propose changes to our [existing code](https://github.com/doodlum/skyrim-community-shaders). As part of validation, we have it also review the tracy capture from before and after the change. I haven't actually gotten live capture set up, but it's not far fetched if that's available through the `eval` tool. My normal use case is to tell it that the pre-optimization tracy is named x and the post- is named y, "please analyze the run to see if the planned optimizations had the desired effect. Etc."
 
> I can see how this could be useful, with an agent trying different approaches to a problem and profiling each one, but I'd like to keep things separated, both at conceptual and review stages.

> The problem here is that this requires a Python interpreter available, which is not something Tracy (the GUI application) integrates, or even considered doing so. It may well be that exposing a programming surface to the LLM is a fine solution here, but that means that the same thing should be somehow available to the user, and that requires layers of additional design.

I think we could drop exposing access to the MCP server from Tracy Assist. I think as long as the MCP access can access whatever you build for Tracy Assist, it solves the maintainability problem I was concerned about. So basically, if you build it for Tracy Assist and expose it as a cpp binding, the MCP identifies it and allows access via eval.

wolfpld:
> expose it as a cpp binding

I suppose that's a python thing? I don't really know how these python bindings work.

I looked at the scope of the changes you made, and this all seems pretty well separated from anything else. I think this can be merged, with the mental note that a different solution will probably be implemented in future.

One thing needs changing though. The markdown manual is automatically generated from the latex one. You need to update the latex manual.

alandtse:
> > expose it as a cpp binding
> 
> I suppose that's a python thing? I don't really know how these python bindings work.
> 
> I looked at the scope of the changes you made, and this all seems pretty well separated from anything else. I think this can be merged, with the mental note that a different solution will probably be implemented in future.
> 
> One thing needs changing though. The markdown manual is automatically generated from the latex one. You need to update the latex manual.

Should be good to go. I've fixed both issues and rebased.

Manual: Added \subsection{MCP Server} to tracy.tex so the LaTeX is now the source of truth. The hand-written section in tracy.md has been removed.

On the bindings question: TracyServerBindings is a compiled C++ pybind11 extension, not Python code. The idea was: if you add new analysis methods to the C++ Worker class, pybind11 exposes them automatically, and the MCP sidecar can call ctx.new_method() via eval with no extra work. Python is only the calling layer, not the implementation.

wolfpld:
> with the end goal of being able to just query "how to make this program faster", and getting an actionable (and correct) solution.

fwiw, with 61875e1b this is now working.

PR #1351, opened 2026-04-30
Allow scrolling flamegraph threads

rmarker:
Previously, when there were many threads in the flamegraph, it could be hard to access the last threads when expanded. They would go off the bounds of the area.
Now, the threads are within a child component with a capped height. A scrollbar will appear when there are many threads, so that the bottom threads can still be reached.

TracyFlamegraphThreads


wolfpld:
Fixed in a more comprehensive way on master.

wolfpld:
obraz

rmarker:
Nice work.
Thanks.

PR #1370, opened 2026-05-22
Add `TRACY_PLATFORM_HEADER` hook for unsupported platforms

Lectem:
Introduce a small extension point so private or otherwise unsupported platforms can plug in their own implementations of the kernel/libc primitives Tracy depends on, without having to patch the existing `#if`/`#elif` chains.

## How it works

Projects supply a "platform header" via `-DTRACY_PLATFORM_HEADER="\"my_platform.h\""` at build time. Tracy pulls it into any translation unit that needs the hooks via a conditional `#include TRACY_PLATFORM_HEADER`. The platform header is the single place where the per-category `TRACY_HAS_CUSTOM_*` macros are toggled and where the corresponding `tracy::Platform*` functions are declared.

`TRACY_PLATFORM_HEADER` is propagated to all consumers of `TracyClient` (`PUBLIC` in CMake, `declare_dependency` in Meson), so hook macros are consistently visible across every TU in a build — which lets fields they supersede (e.g. the POSIX pipe descriptors used by the default `SafeCopyProlog`) be conditionally compiled out without ODR risk.

## Hooks

All in `tracy::` namespace.

- `TRACY_HAS_CUSTOM_THREAD_ID` → `PlatformGetThreadId`
- `TRACY_HAS_CUSTOM_USER_INFO` → `PlatformGetHostname`, `PlatformGetUserLogin`, `PlatformGetUserFullName`
- `TRACY_HAS_CUSTOM_SAFE_COPY` → `PlatformSafeMemcpy`
- `TRACY_HAS_CUSTOM_ALLOCATOR` → `PlatformMalloc`, `PlatformFree`, `PlatformRealloc`, `PlatformAllocatorInit`, `tracy::PlatformAllocatorThreadInit`, `PlatformAllocatorFinalize`, `PlatformAllocatorThreadFinalize`

Each hook is wired as the first arm of its respective `#if`/`#elif` chain, so existing supported platforms are unaffected and the change is purely additive. Only these categories are pluggable today; other platform-specific subsystems (call stack collection, context switch capture, crash handling, system tracing, etc.) still have to be turned off at build time via the matching `TRACY_NO_*` macro on platforms that cannot support them.

The allocator hook reuses Tracy's existing per-thread init plumbing, so user code only has to supply once-per-process and once-per-thread setup functions; they can be empty if your allocator needs no setup. Public `TRACY_API InitRpmalloc()` entry point has been renamed to `InitAllocator()` since it now dispatches to either rpmalloc or the custom allocator hooks. (This is an ABI break for anyone calling the symbol from outside Tracy, which we probably don't support?)

## Templates and documentation

`examples/CustomPlatform/CustomPlatform.h` is a template for the user-provided platform header; `examples/CustomPlatform/CustomPlatform.cpp` is a template for the impl TU. Together they show how to enable any subset of the categories and what each function is expected to do (including the footguns).

A new "Porting to unsupported platforms" subsection in `manual/tracy.tex` was added to document the mechanism.

## Note

The existing non-`TRACY_HAS_CUSTOM_SAFE_COPY` defaults are unsafe in general — the non-Windows path takes the `pipe(2)` trick on every target unconditionally, and MinGW falls into a plain `memcpy`. Tightening that chain to whitelist known-good platforms (Linux is the well-tested one) and `return false` everywhere else could be a better default.
  ### Why I went with this

  1. **Build-system agnostic** — one `-D` works with CMake, Meson, raw `make`, or hand-rolled scripts.
  2. **Compile-time and zero-overhead** — no runtime dispatch, no weak-symbol fallback.
  3. **ODR-safe** — the macro is a build-system constant, identical in every TU, which is what lets me conditionally drop members like `m_pipe` from the `Profiler` class.
  4. **Composable** — per-category `TRACY_HAS_CUSTOM_*` lets users opt in to only the hooks they need, and we don't need to expose every single one of them to the build system. Still allows grouping such as `TRACY_HAS_CUSTOM_SYSTRACE` in the future.
  5. **Familiar** — the pattern is well-established (autoconf-style config headers, Boost's `user-config.hpp`, etc.).

  ### Tradeoffs

  - The macro is set once per build (not per-TU). This is the property that makes layout-gating ODR-safe.
  - The user owns both a header (declarations + toggles) and an impl TU. This keeps Tracy itself unaware of the user's platform.


CMakeLists.txt
```diff
@@ -124,9 +124,20 @@ macro(tracy_set_option_value var help default)
...
 
+# Like tracy_set_option_value but embeds the value as a C string literal
+# (wrapped in quotes), for options whose value is a path or other string used
+# verbatim in C/C++ code — e.g. in #include MACRO indirections.
+macro(tracy_set_option_value_as_string var help default)
+    set_option_value(${var} "${help}" "${default}")
+    if(${var})
+        target_compile_definitions(TracyClient PUBLIC "${var}=\"${${var}}\"")
+    endif()
+endmacro()
```
wolfpld: options.cmake

wolfpld:
This should be 3 commits.

1. Rename.
2. CMake macro.
3. The rest.

CMakeLists.txt
Lectem: Ok, I'll move `tracy_set_option` and `tracy_set_option_value` there too and rename previous `set_option` and `set_option_value` to the `tracy_*`  prefixed names then.

PR #1374, opened 2026-05-26
Allow zooming and panning the flamegraph view.

### LINKED ISSUE #1153: In the flame graph view, switching between instrumentation and samples does not update the scroll bar height
I have a case where my instrumentation flame graph has less max depth than my samples. When I open the flame graph, there's a scroll bar on the right that lets me see all of the instrumentation. When I toggle to "samples", the scroll bar height remains the same, so the bottom of the sample flame graph is cut off with no way to see it.

Image

rmarker:
It could be challenging to examine fine details within the flamegraph.
I would expand the window as large as my monitor would support, but be unable to see areas of interest in sufficient detail.

The flamegraph has been enhanced so that it allows zooming with the mouse wheel, and then panning with the right mouse button.
This provides a familiar experience to the timeline view.

This also resolves #1153 at the same time, as it required rebuilding the view when changing between instrumentation and sampling.

### DIFF
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -4463,7 +4463,7 @@ The default sorting order of the zones on a flame graph \emph{approximates} the 
 
 You can use an alternative sorting method by enabling the \emph{Sort by time} option. This will place the most time-consuming zones first (to the left) on the graph.
 
-You can navigate the flame graph using the mouse. Use the mouse wheel to zoom in and out around the mouse pointer. Pressing the \keys{\ctrl} key makes zooming more precise, while pressing the \keys{\shift} key makes it faster. Dragging with the \RMB{}~right mouse button pans the graph horizontally and vertically.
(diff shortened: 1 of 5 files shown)


wolfpld:
1. I don't understand what the ruler is showing me here:

obraz

2. There should be a way to reset the view to show everything.

3. The instant zoom-in / zoom-out on mouse wheel is quite jarring. It should reuse the timeline animation.

4. There's a max zoom level in sampling view, while there are still details to be uncovered:

obraz

5. The feature is not compatible with live capture where new items are constantly being added, but there is no interaction lock in the logic. This results in a single-frame artifacts appearing when trying to interact with the graph.

rmarker:
I've made several improvements to deal with the issues.

1. Yeah, I thought that keeping the ruler in absolute coordinates, would help keep the user orientated. Though, zooming in it loses clarity. I reworked it to instead be fixed to start at 0 and just show the delta across the view. I added a new horizontal bar to help with orientation instead.

2. I added a button to reset the view.

3. I updated the zooming to use a similar animation to the timeline view.

4. I increased how far it will zoom in. It should be far enough to see the finer details.

5. I think I managed to remove the artifacts when live data is being captured. I don't know how useful it is to zoom in during this time, but I thought it might be less confusing than nothing happening.

wolfpld:
It feels very good now.

Issue #872, opened 2024-08-28
VkCtx::Calibrate can be stuck in an infinite loop, freezing our render thread


VincentHypx:
After roughly 3 minutes of running the profiler on a laptop using the vulkan backend, the calibration values returned by m_vkGetCalibratedTimestampsEXT are never below the expected deviation, therefore it stays stuck in this loop.

Our project is very demanding my current guess is that the video card changes behaviour because of the workload
Either it's struggling with our render demands or it enters a different energy mode, who knows. 
Either way, it ends up delivering less accurate calibration results.

Would it be possible to add a retry approach, so attempt to pass the desired deviation threshold and if it fails after say 8 times, it choses the result that had the least amount of deviation?

Here's the while loop in question
https://github.com/wolfpld/tracy/blob/e26c34346bed7667118ff6dce6b1aa118701efa3/public/tracy/TracyVulkan.hpp#L357C1-L363C42

maxime-modulopi:
I've had success working around this issue by collecting in a separate thread, but I'm not 100% sure this is thread-safe on the Tracy side

tuccio:
I just bumped into the same issue. For me it consistently gets stuck shortly after I connect Tracy, even when the gpu workload is minimal.

In my case, it seems that the check in the `while` loop ends up being consistently 1-2 microseconds off, which leads into `Calibrate` being stuck on that for minutes.

slomp:
@tuccio This one was just merged https://github.com/wolfpld/tracy/pull/1375 and should fix this issue.

PR #1375, opened 2026-05-26
GPU: Vulkan: upper-bound to the calibration loop

slomp:
addresses https://github.com/wolfpld/tracy/issues/872

### DIFF
--- a/public/tracy/TracyVulkan.hpp
+++ b/public/tracy/TracyVulkan.hpp
@@ -164,7 +164,7 @@ public:
         else
         {
             FindCalibratedTimestampDeviation();
-            Calibrate( device, m_prevCalibration, tgpu );
+            Calibrate( m_prevCalibration, tgpu );
             tcpu = Profiler::GetTime();
         }
 
@@ -211,7 +211,7 @@ public:
         // We require a host time domain to be available to properly calibrate.
         FindCalibratedTimestampDeviation();
         int64_t tgpu;
-        Calibrate( device, m_prevCalibration, tgpu );
+        Calibrate( m_prevCalibration, tgpu );
         int64_t tcpu = Profiler::GetTime();
 
         CreateQueryPool();
@@ -263,7 +263,7 @@ public:
             m_tail = head;
             m_oldCnt = 0;
             int64_t tgpu;
-            if( m_timeDomain != VK_TIME_DOMAIN_DEVICE_EXT ) Calibrate( m_device, m_prevCalibration, tgpu );
+            if( m_timeDomain != VK_TIME_DOMAIN_DEVICE_EXT ) Calibrate( m_prevCalibration, tgpu, 10 );
             return;
         }
 #endif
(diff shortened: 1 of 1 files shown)


public/tracy/TracyVulkan.hpp
```diff
@@ -359,13 +359,16 @@ class VkCtx
...
         uint64_t ts[2];
         uint64_t deviation;
-        do
+        constexpr int limit = 10;
+        for( int i = 0; i <= limit; i++ )
         {
-            m_vkGetCalibratedTimestampsEXT( device, 2, spec, ts, &deviation );
+            if( m_vkGetCalibratedTimestampsEXT( device, 2, spec, ts, &deviation ) != VK_SUCCESS ) return;
+            if( deviation <= m_deviation ) break;
+            if( i == limit ) return;
```
wolfpld: Why is it here?

slomp: My understanding is that `FindCalibratedTimestampDeviation()` is called during context initialization to establish a baseline deviation (`m_deviation = minDeviation * 3 / 2`) that is used to accept/reject samples on subsequent recalibrations.

The reported issue seem to indicate that the `while( deviation > m_deviation )` can loop for a long time, suggesting that baseline `m_deviation` can become statistically improbable under some circumstances. Given that `Calibrate()` may be called on every `Collect()` call, it can hang the caller for a non-deterministic amount of time.

But now that you mentioned, `Calibrate()` is called immediately after `FindCalibratedTimestampDeviation()` and is expected to always succeed at that time. I'll need to refactor a few things to ensure that is the case.

slomp: (and the `if( i == limit ) return;` is due to the fact that the for-loop is from 0 to limit _inclusive_ (`<=`), and it _returns_ from the function instead of breaking out of the loop)

wolfpld: Move this check out of the loop.

slomp:
`Calibrate()` now takes `maxSamples` as argument.
During context initialization, `maxSamples` is basically "infinite" to ensure it gets something.
During timestamp collection, `maxSamples` is a smaller number (10) to ensure it won't freeze the thread as time moves on.

public/tracy/TracyVulkan.hpp
```diff
@@ -352,30 +352,47 @@ class VkCtx
...
+    tracy_force_inline int64_t VulkanTimeToPlatformTime(uint64_t tCpu)
     {
+#       if defined _WIN32
+            return tCpu * m_qpcToNs;
+#       elif defined __linux__ && defined CLOCK_MONOTONIC_RAW
+            return tCpu;
+#       else
+            assert( false );
+            return 0;
+#       endif
```
wolfpld: Wrong indentation.

public/tracy/TracyVulkan.hpp
```diff
@@ -352,30 +352,47 @@ class VkCtx
...
+    }
+    tracy_force_inline bool GetCalibratedTimestamps( int64_t& tCpu, int64_t& tGpu, uint64_t& tDeviation )
```
wolfpld, line 367: Newlines.

PR #1380, opened 2026-06-02
Use GL_QUERY_RESULT_NO_WAIT

sacereda:
This should be a little bit faster than the current implementation with GL_QUERY_RESULT_AVAILABLE and GL_QUERY_RESULT.

### DIFF
--- a/public/tracy/TracyOpenGL.hpp
+++ b/public/tracy/TracyOpenGL.hpp
@@ -155,10 +155,18 @@ public:
 
         while( m_tail != m_head )
         {
+#ifdef GL_QUERY_RESULT_NO_WAIT
             uint64_t time = ~0ull;
             glGetQueryObjectui64v( m_query[m_tail], GL_QUERY_RESULT_NO_WAIT, &time );
             if (time == ~0ull) return;
+#else
+            GLint available;
+            glGetQueryObjectiv( m_query[m_tail], GL_QUERY_RESULT_AVAILABLE, &available );
+            if( !available ) return;
 
+            uint64_t time;
+            glGetQueryObjectui64v( m_query[m_tail], GL_QUERY_RESULT, &time );
+#endif
             TracyLfqPrepare( QueueType::GpuTime );
             MemWrite( &item->gpuTime.gpuTime, (int64_t)time );
             MemWrite( &item->gpuTime.queryId, (uint16_t)m_tail );
--- a/public/tracy/TracyOpenGL.hpp
+++ b/public/tracy/TracyOpenGL.hpp
@@ -155,12 +155,9 @@ public:
 
         while( m_tail != m_head )
         {
-            GLint available;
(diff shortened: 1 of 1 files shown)


slomp:
We may need to check whether OpenGL 4.4 is available in the context:

https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGetQueryObject.xhtml
> `GL_QUERY_RESULT_NO_WAIT` is accepted for `pname` **only if** the GL version is 4.4 or greater.

public/tracy/TracyOpenGL.hpp
```diff
@@ -155,13 +155,18 @@ class GpuCtx
 
         while( m_tail != m_head )
         {
+#ifdef GL_QUERY_RESULT_NO_WAIT
```
slomp, line 158: I suppose this solution would work for the most part.
The problem is that it's quite possible to use a GL header/loader that defines this macro, but be operating in a GL context that does not recognize the underlying value of it.

I think the ideal solution would be to query the GL context for support when the context is created, store that in a boolean, and then test the boolean value here instead of testing the macro.

I asked Gemini and it recommended something like this:
```cpp
bool SupportsQueryResultNoWait() {
    // 1. Check if the core version is 4.4 or higher
    GLint major, minor;
    glGetIntegerv(GL_MAJOR_VERSION, &major);
    glGetIntegerv(GL_MINOR_VERSION, &minor);
    
    if (major > 4 || (major == 4 && minor >= 4)) {
        return true;
    }

    // 2. Fallback: Check if the extension is supported on an older context (e.g., OpenGL 4.3)
    GLint numExtensions = 0;
    glGetIntegerv(GL_NUM_EXTENSIONS, &numExtensions);

    for (GLint i = 0; i < numExtensions; ++i) {
        const char* ext = (const char*)glGetStringi(GL_EXTENSIONS, i);
        if (ext && std::string(ext) == "GL_ARB_query_buffer_object") {
            return true;
        }
    }

    return false;
}
```

slomp (commented):
I haven't used much OpenGL over the past years, so maybe it's just the "shell shock" of having to deal with OpenGL extensions in the past that is making me pedantic and overzealous here.

public/tracy/TracyOpenGL.hpp
wolfpld: What's the minspec required for this kind of extensions check?

slomp: According to the docs, `glGetStringi` et al. was introduced to core in OpenGL 3.0:
https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGetString.xhtml
There's also the old way of doing things (`glGetString`, OpenGL 2) where you get a gigantic string with all extensions on it and split/parse it manually.

slomp:
Addressed in https://github.com/wolfpld/tracy/pull/1404 (merged).

PR #1392, opened 2026-06-06
Speeding-up GitHub build-bots

slomp:
Reducing the number of parallel workers when building the profiler has reduced build times substantially.
Notably, macos used to take 45+min to build, and now builds in about 3min (must be because the bots were running out of available RAM, and started swapping memory to disk).

The windows and linux builds were much less impacted (5 to 6min), remaining in the same ballpark as before (6 to 8min).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -29,11 +29,7 @@ jobs:
     - name: Profiler GUI
       run: |
         cmake -B profiler/build -S profiler -DCMAKE_BUILD_TYPE=Release -DGIT_REV=${{ github.sha }}
-        if [ "${ACT:-}" != "true" ] && [ "${FORGEJO_ACTIONS:-}" != "true" ]; then
-          cmake --build profiler/build
-        else
-          cmake --build profiler/build --parallel
-        fi
+        cmake --build profiler/build --parallel 2
     - name: Update utility
       run: |
         cmake -B update/build -S update -DCMAKE_BUILD_TYPE=Release -DGIT_REV=${{ github.sha }}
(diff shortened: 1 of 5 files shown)


.github/workflows/linux.yml
```diff
@@ -29,11 +29,7 @@ jobs:
     - name: Profiler GUI
       run: |
         cmake -B profiler/build -S profiler -DCMAKE_BUILD_TYPE=Release -DGIT_REV=${{ github.sha }}
-        if [ "${ACT:-}" != "true" ] && [ "${FORGEJO_ACTIONS:-}" != "true" ]; then
```
wolfpld, line 32: Why was this removed?

.github/workflows/macos.yml
```diff
@@ -27,8 +27,14 @@ jobs:
...
       run: |
+        # To disable LTO: -DNO_LTO=ON
+        # To enable Enable ClangBuildAnalyzer: -DCMAKE_CXX_FLAGS="-ftime-trace"
+        # brew install clang-build-analyzer
         cmake -B profiler/build -S profiler -DCMAKE_BUILD_TYPE=Release -DGIT_REV=${{ github.sha }}
-        cmake --build profiler/build --parallel --config Release
+        # ClangBuildAnalyzer --start profiler/build
+        cmake --build profiler/build --parallel 2 --config Release
+        # ClangBuildAnalyzer --stop profiler/build profiler_build.bin
+        # ClangBuildAnalyzer --analyze profiler_build.bin
```
wolfpld: What's the function of these comments?

.github/workflows/linux.yml
slomp: Ah, ops, I nuked them just to make sure I was running the right path remotely when adjusting the `--parallel`.

.github/workflows/macos.yml
slomp: These were the commands necessary to run clang-analyzer.
I left them there as future reference, but I'll remove if preferred.

PR #1394, opened 2026-06-06
fix: add opengl drift correction for gpu zones

### LINKED ISSUE #1393: Implement OpenGL Automatic Drift Correction
## Task
Implement automatic drift correction in OpenGL, like we've done with vulkan and d3d12. 

## Context
On my machine (nvidia 5070ti, amd 9950x3d, windows 11) opengl drift is nearly but not totally constant, so the manual drift adjustment isn't very good in longer traces. 

Here's drift logged:

Image

and of course, showing the actual drift (drift is not manually corrected here):
Image
Image

bruno-dasilva:
Fixes https://github.com/wolfpld/tracy/issues/1393

## Description
Implements opengl drift correction using the existing api (similar to vulkan and d3d12).

### Before
(no clock drift correction applied):
Screenshot 2026-06-06 170937
Later:
Screenshot 2026-06-06 170612

### After:
image

Drift at beginning of game:
image

Later:
image

## AI Disclosure
Implemented via Claude Code, tested by me.

### DIFF
--- a/manual/tracy.tex
+++ b/manual/tracy.tex
@@ -1701,6 +1701,12 @@ logo=\bcattention
 \end{itemize}
 \end{bclogo}
 
+\subparagraph{Calibrated context}
+
+By default, the OpenGL context is uncalibrated: the CPU and GPU clocks are aligned only once, when the context is created, so over long captures the two time domains may drift apart (section~\ref{options} describes correcting this drift manually). Defining \texttt{TRACY\_OPENGL\_AUTO\_CALIBRATION} before including \texttt{TracyOpenGL.hpp} enables periodic recalibration instead: roughly once per second Tracy samples the GPU and CPU clocks together and emits a calibration event, allowing the profiler to track and remove the drift automatically.
+
+This is opt-in because OpenGL exposes no atomic CPU+GPU timestamp query (unlike Vulkan's \texttt{VK\_EXT\_calibrated\_timestamps} or Direct3D~12, whose contexts are always calibrated). Recalibration therefore reads the GPU clock with \texttt{glGetInteger64v(GL\_TIMESTAMP)}, which forces a CPU/GPU synchronization (a pipeline stall) each time it runs. Enable it only when the improved long-capture alignment is worth the periodic stall.
+
 \subsubsection{Vulkan}
 
 Similarly, for Vulkan support you should include the \texttt{public/tracy/TracyVulkan.hpp} header file. Tracing Vulkan devices and queues is a bit more involved, and the Vulkan initialization macro \texttt{TracyVkContext(physdev, device, queue, cmdbuf)} returns an instance of \texttt{TracyVkCtx} object, which tracks an associated Vulkan queue. Cleanup is performed using the \texttt{TracyVkDestroy(ctx)} macro. You may create multiple Vulkan contexts. To set a custom name for the context, use the \texttt{TracyVkContextName(ctx, name, size)} macro.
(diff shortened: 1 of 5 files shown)


wolfpld:
`glGetInteger64v(GL_TIMESTAMP)` forces CPU to GPU synchronization. This is a no go for the default configuration.

bruno-dasilva:
> `glGetInteger64v(GL_TIMESTAMP)` forces CPU to GPU synchronization. This is a no go for the default configuration.

Thank you, good catch. I suppose there's two options:
1. make a configurable non-default option
2. find a way to do this without cpu<->gpu synchronization

wolfpld:
I don't think you would find any. Making this opt-in should be ok.

bruno-dasilva:
> I don't think you would find any. Making this opt-in should be ok.

We could switch this over to a glQueryCounter maybe? the drift fix wouldn't be perfect (since it would be delayed at worst to a couple frames?) but could still be better than default? I'll try it out and report back.

Otherwise, happy to put the synchronous version behind a opt-in config. You know better than I, would that be best as a compiler flag or is there a better way to handle this config?

wolfpld:
Compiler flag, like the other existing config options.

bruno-dasilva:
For anyone else reading this in the future:
I did try and get the async approach to work but it's got a fundamental problem: getting a cpu timestamp whenever the gpu actually grabs the gpu timestamp. You'd need a Collect() at _exactly_ the time the gpu has finished getting the timestamp. trying to do that within your regular frames will lead to random drifts every time you re-calibrate - worse than just letting things drift.
You _could_ run a separate thread that polls the gpu consistently but that's frankly overkill and has its own can of worms.

So yea, I'll go with synchronous but behind a compiler flag.

manual/tracy.tex
```diff
@@ -1701,6 +1701,12 @@ \subsubsection{OpenGL}
 \end{itemize}
 \end{bclogo}
 
+\subparagraph{Calibrated context}
+
+By default, the OpenGL context is uncalibrated: the CPU and GPU clocks are aligned only once, when the context is created, so over long captures the two time domains may drift apart (section~\ref{options} describes correcting this drift manually). Defining \texttt{TRACY\_OPENGL\_AUTO\_CALIBRATION} before including \texttt{TracyOpenGL.hpp} enables periodic recalibration instead: roughly once per second Tracy samples the GPU and CPU clocks together and emits a calibration event, allowing the profiler to track and remove the drift automatically.
+
+This is opt-in because OpenGL exposes no atomic CPU+GPU timestamp query (unlike Vulkan's \texttt{VK\_EXT\_calibrated\_timestamps} or Direct3D~12, whose contexts are always calibrated). Recalibration therefore reads the GPU clock with \texttt{glGetInteger64v(GL\_TIMESTAMP)}, which forces a CPU/GPU synchronization (a pipeline stall) each time it runs. Enable it only when the improved long-capture alignment is worth the periodic stall.
+
```
bruno-dasilva, line 1709: feel free to edit the position/content of this to match y'alls standards - not sure if worth mentioning vulkan here for example.

bruno-dasilva:
just to confirm the alignment was correct I added a glfinish() at the end of my frames, and you can see it line up more or less perfectly:
image

wolfpld:
Please also add a line in `meson.options`.

bruno-dasilva:
> Please also add a line in `meson.options`.

added to both `meson.options` and `meson.build` :) Thanks!

bruno-dasilva:
thank you @wolfpld !

PR #1399, opened 2026-06-10
Allow OpenGL on MacOS when possible

slomp:
While OpenGL has been deprecated on macOS, it still runs with some caveats:
- Apple Silicon: most hardware do not have driver support for timestamp commands, so TracyOpenGL.hpp won't produce meaningful timestamps.
- Intel: TracyOpenGL should work just fine.


slomp:
I decided to add a `Valid()` interface to allow for instrumentation to remain as is in case it is determined at run-time that the OpenGL implementation is incapable of the timestamp granularity required by TracyOpenGL.

slomp:
(closing this until I can get further confirmation)

PR #1400, opened 2026-06-10
adding OpenGL example (spinning triangle)

slomp:


slomp:
Should be good to go now: Windows, macOS, Linux X11 and through XWayland, all via RGFW which is small/quick to fetch via the CMakeLists.txt.

slomp:
Now through ~SDL2~ SDL3.

PR #1404, opened 2026-06-16
Possibly faster GL timestamp query result check

slomp:
Addresses https://github.com/wolfpld/tracy/pull/1380

(Co-authored by @sacereda)

### DIFF
--- a/public/tracy/TracyOpenGL.hpp
+++ b/public/tracy/TracyOpenGL.hpp
@@ -244,7 +244,7 @@ private:
         return exts && strstr( exts, feature ) != nullptr;
     }
 
-    bool GetTimestamp( uint64_t& timestamp, unsigned int queryId )
+    tracy_force_inline bool GetTimestamp( uint64_t& timestamp, unsigned int queryId )
     {
         if( m_supportsQueryBufferObject )
         {
--- a/public/tracy/TracyOpenGL.hpp
+++ b/public/tracy/TracyOpenGL.hpp
@@ -248,7 +248,7 @@ private:
     {
         if( m_supportsQueryBufferObject )
         {
-            constexpr uint64_t sentinel = ~0ull;
+            constexpr uint64_t sentinel = ~uint64_t(0);
             uint64_t time = sentinel;
             glGetQueryObjectui64v( m_query[queryId], GL_QUERY_RESULT_NO_WAIT, &time );
             if ( time == sentinel ) return false;
@@ -258,7 +258,7 @@ private:
         {
             GLint available;
             glGetQueryObjectiv( m_query[queryId], GL_QUERY_RESULT_AVAILABLE, &available );
-            if( !available ) return false;
+            if( available == GL_FALSE ) return false;
             uint64_t time;
             glGetQueryObjectui64v( m_query[queryId], GL_QUERY_RESULT, &time );
             timestamp = time;
--- a/public/tracy/TracyOpenGL.hpp
+++ b/public/tracy/TracyOpenGL.hpp
@@ -102,6 +102,7 @@ public:
         : m_context( GetGpuCtxCounter().fetch_add( 1, std::memory_order_relaxed ) )
         , m_head( 0 )
         , m_tail( 0 )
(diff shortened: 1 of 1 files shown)


slomp:
Two things to consider:
- Is `~0ull` a good default? Should we use `0` instead? Could these timestamps actually happen and cause trouble?
- Should the `GetTimestamp` helper function be `tracy_force_inline`?

wolfpld:
> Is `~0ull` a good default? Should we use `0` instead? Could these timestamps actually happen and cause trouble?

The actual (non-conforming) Intel implementation I looked at did something like mod 36 bits, so you can get 0, but not ~0.

slomp:
I suppose if we wanted to be paranoid about it, we would:
a) initialize the timestamp to the `˜0ull` sentinel
b) call `glGetQueryObjectui64v` with `GL_QUERY_RESULT_NO_WAIT`
c) if the resulting timestamp is the sentinel, fall-through the habitual `GL_QUERY_RESULT_AVAILABLE` path to be sure

slomp:
> I suppose if we wanted to be paranoid about it, we would: a) initialize the timestamp to the `˜0ull` sentinel b) call `glGetQueryObjectui64v` with `GL_QUERY_RESULT_NO_WAIT` c) if the resulting timestamp is the sentinel, fall-through the habitual `GL_QUERY_RESULT_AVAILABLE` path to be sure

On a second thought, this could sort of defeat the purpose of the `GL_QUERY_RESULT_NO_WAIT` optimization, as it would just short-circuit the case where the timestamp is ready, but would extend the case where it is not. So it's a trade-off between what to expect on average with the pending queries every time `Collect()` gets called.

wolfpld:
**`GL_QUERY_RESULT_NO_WAIT` is used unconditionally without a compile-time guard.** The symbol only exists when GL headers include `GL_ARB_query_buffer_object`. The runtime `CheckFeature` check doesn't help if the symbol is absent at compile time — the file won't compile. Projects that currently compile against `TracyOpenGL.hpp` without this extension in their headers will break.

The existing code does `!defined GL_TIMESTAMP && defined GL_TIMESTAMP_EXT` check and "renames" the alternative symbols and defines. How is this handled here for the GLES and/or non-core path?

sacereda:
I guess in that case you could guard that branch using the original ifdef:
```
if (0)
    ;
#ifdef GL_QUERY_RESULT_NO_WAIT
else if( m_supportsQueryBufferObject )
{
    ...
}
#endif
else
{
    ...
}
```
At some point I also considered a compile-time opt-in mechanism as an alternative (TRACY_FAST_COLLECT or similar), but I'm not sure whether you already have something along those lines in the codebase.

slomp:
According to the [registry](https://registry.khronos.org/OpenGL/extensions/registry.py), there's no preceding `EXT` variant of `GL_QUERY_RESULT_NO_WAIT`. It seems to first appear as an AMD extension in [AMD_query_buffer_object](https://registry.khronos.org/OpenGL/extensions/AMD/AMD_query_buffer_object.txt), and immediately promoted to core in [ARB_query_buffer_object](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_query_buffer_object.txt)).

I suppose we could do something like this:
```
#ifndef GL_QUERY_RESULT_NO_WAIT
#define GL_QUERY_RESULT_NO_WAIT 0x9194
#endif
```

I am not well-versed in GLES, but the `glGetString` and `glGetStringi` should be there to check for extensions. The "version check" I am not so sure, but even if `glGetIntegerv( GL_MAJOR_VERSION, &major );` is not applicable for GLES, `CheckFeature` would fall-through the common-denominator (`glGetString`) path.

slomp:
`glGetString` is available on GLES 2:
https://docs.gl/es2/glGetString

`glGetStringi` is available on GLES 3:
https://docs.gl/es3/glGetString

`GL_MAJOR_VERSION` (`glGetIntegerv()`) is on GLES 3:
https://docs.gl/es3/glGet

So I suppose we would also need an `#ifndef GL_MAJOR_VERSION` check to cover all basis.

PR #1407, opened 2026-06-17
feat: display threads in a consistent order

shekhirin:
This makes the flamegraph thread picker use the same thread ordering as the timeline.

Previously, the timeline maintained `m_threadOrder`, while the flame graph visible-thread picker iterated over raw worker thread data. That made thread order differ between views.

This change:
- Extracts the timeline thread-order maintenance into `View::UpdateThreadOrder()`.
- Keeps timeline behavior unchanged by calling the shared helper.
- Makes the flame graph visible-thread picker use `m_threadOrder`.

This is especially useful for having the threads sorted in all views.

image


wolfpld:
The thread list is also used in `TracyView_ContextSwitch.cpp` and `TracyView_Messages.cpp`.

shekhirin:
@wolfpld done

PR #1414, opened 2026-06-26
Fix image and symbol resolution on AerynOS and elsewhere

YaLTeR:
The fixes are likely helpful on Linux in general, I'm just currently testing on AerynOS.

| Before | After |
| --- | --- |
| Screenshot from 2026-06-26 10-35-42 | image |

Commit 1: fixes wrong elf header ordering assumption that prevented tracy-client from detecting .so files in the address space. Seems to be particularly triggered by Aeryn's clang-built .so files where the last header is a tiny-sized note.

Commits 2 and 3: multi-frame zstd support from upstream libbacktrace: https://github.com/ianlancetaylor/libbacktrace/commits/master/ Required to decompress most debuginfo on Aeryn.

Commit 4: my non-single-segment zstd patch, submitted to libbacktrace: https://github.com/ianlancetaylor/libbacktrace/pull/165 Required to decompress certain debuginfo on Aeryn.

### DIFF
--- a/public/client/TracyCallstack.cpp
+++ b/public/client/TracyCallstack.cpp
@@ -237,9 +237,18 @@ private:
         if( cache->ContainsImage( startAddress ) ) return 0;
 
         const uint32_t headerCount = info->dlpi_phnum;
-        assert( headerCount > 0);
-        const auto endAddress = static_cast<uint64_t>( info->dlpi_addr +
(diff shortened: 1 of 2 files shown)


wolfpld:
Thanks. This brings libbacktrace up-to-date with the current upstream, right?

YaLTeR:
Not sure tbh, I only took those commits because they solved the issue at hand, I haven't checked the rest of commit history against Tracy's in-tree version.

PR #1420, opened 2026-06-30
Fix "quantization" of sampling flame graph horizontal panning

YaLTeR:
Store sampling flame graph time as time instead of samples. Practically, this fixes horizontal panning being "quantized" to 100 µs steps (one period). I guess it also simplifies the code a bit and makes "time" actually mean time.

I think I updated everything and as far as I can see, things work properly.

One thing that changed is maximum zoom level (you can zoom in more), but you can also do that on the timeline so maybe it's ok? I can try to adjust it if needed.

Before:

https://github.com/user-attachments/assets/de2374b4-fbe2-4f22-a372-397cffb73cff

After:

https://github.com/user-attachments/assets/555a45ea-8af1-4ab1-b106-f7a10b07d475

### DIFF
--- a/profiler/src/profiler/TracyView.hpp
+++ b/profiler/src/profiler/TracyView.hpp
@@ -334,7 +334,7 @@ private:
     void DrawManual();
     void DrawFrameStatistics();
     void DrawFlameGraph();
-    void DrawFlameGraphHeader( int64_t vStart, int64_t vEnd, uint64_t period );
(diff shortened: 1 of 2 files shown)


wolfpld:
> One thing that changed is maximum zoom level (you can zoom in more), but you can also do that on the timeline so maybe it's ok? I can try to adjust it if needed.

It's ok.

PR #1421, opened 2026-07-06
Change musl bits/reg.h to sys/reg.h

YaLTeR:
This is the new location for `__WORDSIZE`. `bits/reg.h` is an empty header. The change was made in 2024:
https://git.musl-libc.org/cgit/musl/commit/?id=e709a6f07ade208ba513f9225222336f30c304b0

Fixes symbol resolution on musl (particularly, postmarketOS aarch64).

### DIFF
--- a/public/client/TracyElf.hpp
+++ b/public/client/TracyElf.hpp
@@ -3,11 +3,6 @@
 
 #include <stdint.h>
 
-#if !defined(__GLIBC__) && !defined(__WORDSIZE)
-// include __WORDSIZE headers for musl
-#  include <bits/reg.h>
-#endif
-
 namespace tracy
 {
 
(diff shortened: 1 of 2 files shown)


public/client/TracyElf.hpp
```diff
@@ -5,7 +5,7 @@
 
 #if !defined(__GLIBC__) && !defined(__WORDSIZE)
 // include __WORDSIZE headers for musl
```
bugaevc: `!defined(__GLIBC__)` is of course not a very great way to detect musl; on top of that `bits/` is an internal implementation detail that should not be relied upon.

And `__WORDSIZE` is private too, as indicated by the double underscore. AFAICS, it is only used to understand whether we're on 32-bit or 64-bit system as far as ELF formats are concerned? Can't we do that in some much much easier way, without depending on explicitly private APIs?

wolfpld: In practice `__WORDSIZE` is depended on in various programs, regardless of the original intent. Which is probably why it came as an answer to the question how to determine what to set for libbacktrace. I don't have a particular need to keep using it, especially as it's a source of problems, like the one this PR resolves. Do you have an alternative proposal? `sizeof(void*)`?

bugaevc: `sizeof (void *)`, or `__LP64__`

wolfpld: @YaLTeR How does this look for you?

YaLTeR: Could try and see if it breaks anything I guess

wolfpld: Looking at the kernel, `sizeof(long)` may be more appropriate?

obraz

I don't have a checkout at hand to see how what the elf struct uses for its sizes in kernel.

YaLTeR: Replaced `__WORDSIZE` with `__SIZEOF_POINTER__` documented in https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html

PR #1432, opened 2026-07-20
fix(mcp): recover from hangs and evict idle files

alandtse:
## Summary

Addresses the "server just vanishes" pattern from alandtse/tracy#2, where `Unable to connect` gives no way to tell a segfault apart from a hang and required manually killing an unresponsive process before restart:

- `faulthandler.enable()` at startup writes a thread-state traceback to `tracy_mcp.crash.log` on a genuine fatal crash (works on Windows via `SetUnhandledExceptionFilter`).
- The periodic sweep loop now also logs a heartbeat (uptime, instance/task counts, evictions) so a hung event loop is distinguishable from a dead process by the last timestamp on disk.
- `_is_our_server_running()` now backs its `os.kill(pid, 0)` check with an HTTP self-ping. A deadlocked-but-alive process passed the old PID-only check, silently blocking restart; a non-responsive server is now reported by PID and a fresh instance starts on a new port instead.

Also closes a separate gap: automatic eviction only ever covered disconnected live instances, so a file-loaded capture — potentially many GB — stayed resident forever until the `TRACY_MCP_MAX_INSTANCES` cap forced an LRU eviction to make room. Generalizes the sweep into `_evict_idle`: file-loaded instances now get their own idle-since-last-use TTL (`TRACY_MCP_FILE_IDLE_TTL_S`, default 1800s matching the disconnected-live TTL). Safe to evict on a timer since they're already durably on disk — `load_capture` brings them back. Connected live instances are untouched.

Updates `eval_guide.md`'s "Instance lifecycle" section, which previously claimed "two backstops" and omitted the new file-idle TTL.

## Test plan

- [x] `py_compile` clean
- [x] Unit-style test against `_evict_idle` covering three cases: stale file-loaded instance evicted (and `shutdown()` called), fresh file-loaded instance kept, connected live instance kept regardless of idle time — all pass
- [x] Live-tested against the real MCP server: `list_instances` correctly reports `background_done`, faulthandler/heartbeat wired up at startup

### DIFF
--- a/extra/mcp/eval_guide.md
+++ b/extra/mcp/eval_guide.md
@@ -97,11 +97,14 @@ frees it just because your conversation ends.
 - `list_instances` reports `idle_seconds`, `connected`, and `background_done`
   per instance; use it to spot stale ones before starting a new session,
   especially if you're about to load several captures for comparison.
-- The server does provide two backstops, both configurable via environment
+- The server provides three backstops, all configurable via environment
   variables at startup: it evicts the least-recently-used *evictable*
   instance (disconnected live sessions, or any loaded file capture — never
   a still-connected live one) once the instance count reaches
-  `TRACY_MCP_MAX_INSTANCES` (default 4), and it drops a disconnected live
+  `TRACY_MCP_MAX_INSTANCES` (default 4); it drops a disconnected live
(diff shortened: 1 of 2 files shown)


wolfpld:
Needs resolving conflicts.

PR #1435, opened 2026-07-27
Ordering the clients in the connection dialog to prevent jumping in the list. Ordering by time still caused jumps.

damcclos:


wolfpld:
Some notes from LLM, might be worth considering:

- The sort + std::vector allocation runs every frame even when the client set is unchanged. Given the expected small n, this is fine; flagging only for completeness.
- A client that times out (4s) and reappears is assigned a fresh clientOrder++, so it moves to the bottom of the list. This is consistent with "stable by first-seen order" and matches the PR's intent, but it's a behavior worth being aware of: brief network blips will cause that client to jump to the bottom rather than recover its original slot.

I will review it properly later.

damcclos:
I feel the sort+std::vector every frame is reasonable for now compared to either changing the clients map to some ordered version or a vector itself and all the changes that comes with that.

As for the network blips, the reason I made this change is my work causes many processes to start with tracy compiled in so the client list grows quickly and jumps around. Identifying the correct process is sometimes a challenge and other times I go to choose the process and another one plops in where I was clicking. So, a network blip causing a process to disappear and the drop to the bottom, I think, is better than the whole list shoving down after it just shifted up.

PR #1437, opened 2026-07-29
Fix potential out-of-bounds write in InsertLockEvent by adding checks for MaxLockThreads. Ensure thread index is valid before accessing lockmap.range.

damcclos:


damcclos:
I am not entirely convinced this is the best option to fix this issue. The application I am playing with has a ton of threads with a bunch of locks. I've had to disable locks before because of the 64 max lock count but this was a different signature.

server/TracyWorker.cpp
```diff
@@ -12,6 +12,7 @@
 
 #include <cctype>
 #include <chrono>
+#include <cstdio>
```
wolfpld, line 15: Why is this added?

server/TracyWorker.cpp
```diff
@@ -3726,12 +3726,24 @@ void Worker::InsertLockEvent( LockMap& lockmap, LockEvent* lev, uint64_t thread,
...
     if( it == lockmap.threadMap.end() )
     {
+        // waitList / range[] only support MaxLockThreads (64). Exceeding that
+        // used to uint8_t-truncate the index and OOB-write lockmap.range[],
+        // corrupting adjacent slab objects (non-deterministic findIdx AVs).
+        if( lockmap.threadList.size() >= MaxLockThreads )
+        {
+            return;
+        }
         assert( lockmap.threadList.size() < MaxLockThreads );
```
wolfpld, line 3736: This is redundant now.

server/TracyWorker.cpp
```diff
@@ -3726,12 +3726,24 @@ void Worker::InsertLockEvent( LockMap& lockmap, LockEvent* lev, uint64_t thread,
     auto it = lockmap.threadMap.find( thread );
     if( it == lockmap.threadMap.end() )
     {
+        // waitList / range[] only support MaxLockThreads (64). Exceeding that
+        // used to uint8_t-truncate the index and OOB-write lockmap.range[],
+        // corrupting adjacent slab objects (non-deterministic findIdx AVs).
```
wolfpld, line 3731: "uint8_t-truncate" is misleading: OOB occurs for threadList.size() in 64, 255 where the uint8_t value is stored unchanged. Truncation only kicks in past 255. Root cause is unbounded threadList growth, not truncation.

wolfpld:
> I am not entirely convinced this is the best option to fix this issue. The application I am playing with has a ton of threads with a bunch of locks. I've had to disable locks before because of the 64 max lock count but this was a different signature.

Changing the current implementation is not worth it, because it fundamentally solves the wrong problem (assumes lock events can arrive out-of-order; they can't). A redesign is needed, and it will solve most of the existing problems.

PR #1449, opened 2026-08-07
add TracyFormat.h to common includes

Nelonn:
I'm using tracy install script so this problem is very important for me


wolfpld:
30d14caa

PR #1453, opened 2026-08-13
fix(client): correct TracyPlotConfig color byte order (#1232)

### LINKED ISSUE #1232: Color byte mismatched in 'TracyPlotConfig'
`TracyPlotConfig` sends color code directly to server without swapping byte order:

```C++
#define TracyPlotConfig( name, type, step, fill, color ) \
                         tracy::Profiler::ConfigurePlot( name, type, step, fill, color )

MemWrite( &item->plotConfig.color, color );
````

I have noticed that some fixes were made, but `TracyPlotConfig` still uses the wrong protocol
[https://github.com/wolfpld/tracy/pull/514](https://github.com/wolfpld/tracy/pull/514)

rachelwelchio:
Fixes #1232

`TracyPlotConfig` now packs plot colors the same way as zone/message color events (swap R/B in the 0xRRGGBB value before writing to the queue), so the profiler displays the intended color.

### DIFF
--- a/public/client/TracyProfiler.hpp
+++ b/public/client/TracyProfiler.hpp
@@ -446,7 +446,8 @@ public:
         MemWrite( &item->plotConfig.type, (uint8_t)type );
         MemWrite( &item->plotConfig.step, (uint8_t)step );
         MemWrite( &item->plotConfig.fill, (uint8_t)fill );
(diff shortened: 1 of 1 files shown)


wolfpld:
AI spam

PR #1461, opened 2026-08-26
Added Profiler::GetPort() and TracyPort macro

ofoure:
I needed the port that Tracy is really using so my process could programmatically start the profiler viewer and auto connect by specifying the port via argument "-p".
Without that it's always connecting to the default port which could be any process on the same ip.

Maybe someone else could need that too.

TracyPort and Profiler::GetPort() both return 0 by default or before listen port is established.


wolfpld:
This has not been done with proper concurrency in mind.

ofoure:
Right, it was subject to a race condition.

wolfpld:
Also needs documentation in the manual.

PR #1466, opened 2026-09-02
Use CMAKE_OSX_DEPLOYMENT_TARGET 15.0 by default

akx:
As it is, the CI-built macOS binaries refuse to work on macOS Sequoia because it defaults to the runner machine's level.

This sets a lower-than-default default for the target.


akx:
Thanks for the merge! And thanks for Tracy altogether! ❤️

PR #1467, opened 2026-09-08
Add option to group section rows by category

psmyles:
We recently replaced our own frame-spanning markers (a long zone per game wave and named frame sets for scene loads and timelines) with sections introduced in 0.14 release, as one category per kind of activity. 

Right now sections are packed into timeline rows purely by length, so sections from different categories share rows and the same category can move between rows along the timeline. When categories mark different kinds of activity (game levels, loading screens, timelines) this makes the row layout hard to read.

The new "Group by category" option in the section options lays out each category in its own set of rows, in category identifier order, with the existing length-first packing applied within a category. This is off by default, so the current layout is unchanged unless the user enables it. It is only shown when the trace has more than one section category.

Before:
before
After:
after


manual/tracy.tex
```diff
@@ -4345,7 +4345,7 @@ \subsection{Options menu}
 \item \emph{\faSignature{} Draw CPU usage graph} -- You can disable drawing of the CPU usage graph here.
 \end{itemize}
 \item \emph{\faEyeDropper{}~Draw stack samples} -- Controls if stack samples for each thread are displayed on the timeline.
-\item \emph{\faArrowsLeftRightToLine{}~Draw sections} -- Allows disabling display of sections (see chapter~\ref{sections}). If there are multiple section categories, there's also an expandable list of categories that can be disabled or enabled.
+\item \emph{\faArrowsLeftRightToLine{}~Draw sections} -- Allows disabling display of sections (see chapter~\ref{sections}). If there are multiple section categories, there's also an expandable list of categories that can be disabled or enabled. The \emph{\faLayerGroup{}~Group by category} option lays out each category in its own set of rows, ordered by category identifier, instead of packing all sections together by length.
```
wolfpld: This is not a valid newline.

psmyles: Ah, missed that. Fixed it now, please check again.

PR #1473, opened 2026-09-16
Add "pin" toggle to anchor threads and plots (and GPU queues)

slomp:
When pinned, threads (and GPU queues) are anchored to the top of the viewport, and plots are anchored to the bottom, and will stay in-place until unpinned. The pin/unpin toggle is available in the menu when right-clicking a thread/plot name.


profiler/src/profiler/TracyTimelineController.cpp
```diff
@@ -138,29 +165,82 @@ void TimelineController::End( double pxns, const ImVec2& wpos, bool hover, bool
...
+    int topOffset = 0, normalOffset = 0, bottomOffset = 0;
     for( auto& item : m_items )
     {
+        const int off = itemYOffset( item, topOffset, normalOffset, bottomOffset );
         if( item->WantPreprocess() && item->IsVisible() )
         {
-            const auto yPos = wpos.y + yOffset;
-            const bool visible = m_firstFrame || ( yPos < yMax && yPos + item->GetHeight() >= yMin );
+            const auto yPos = wpos.y + off;
+            const bool visible = item->IsPinned() || m_firstFrame || ( yPos < yMax && yPos + item->GetHeight() >= yMin );
```
wolfpld: Should pinning imply visibility?

wolfpld:
- CPU-data track eats clicks under a pinned band at profiler/src/profiler/TracyView_CpuData.cpp:448. Its release hit-test is not gated on hover, and the rect is clipped only to the item's own clip rect, which extends under the band. Clicking a zone on the pinned track above it silently clears the CPU-timeline thread selection.
- Thread cropper accepts clicks on invisible circles at profiler/src/profiler/TracyView_ZoneTimeline.cpp:651. A depth-limited thread scrolled under a band still runs its raw mouse-distance test and can erase or change the depth limit with no visual feedback.
- Same-frame double draw when pinning at profiler/src/profiler/TracyTimelineController.cpp:223. Pin state is re-evaluated after the normal loop, so an item pinned from its own popup is drawn again by the band pass in the same frame. Produces a one-frame ghost header and a double height adjustment.
- Opaque band backgrounds occlude parent-window overlays at profiler/src/profiler/TracyTimelineController.cpp:219. Frame separators, the frame-target highlight, and outside-trace shading are drawn into the parent draw list and vanish inside the bands, giving a visible discontinuity. The WindowBg colour also does not match the actual timeline background in the default root-window build.
- Stale band height during collapse animation at profiler/src/profiler/TracyTimelineController.cpp:125. The top band height is summed before Draw, so vertical-centre scroll compensation lags by a frame while a pinned thread animates. Content under the cursor creeps instead of staying put.
- Tracks hidden under bands are fully preprocessed each frame at profiler/src/profiler/TracyTimelineController.cpp:187. The visibility test uses the raw viewport rather than the band-narrowed region, so hidden threads walk their zone trees and dispatch tasks that Draw then discards.
- Redundant pre-scan in UpdateCenterItem at profiler/src/profiler/TracyTimelineController.cpp:58. The first and last normal key lookup doubles the per-frame item visits.
- Band edges computed twice at profiler/src/profiler/TracyTimelineController.cpp:206. The mouse-in-band bools duplicate the narrowed context's yMin and yMax.

